PR #8, opened 2015-04-24
Add X10 Support

edc1591:
This PR adds support for controlling X10 modules via my [rest-mochad](https://github.com/edc1591/rest-mochad) app.

X10 is an old protocol for controlling lights and pretty much any appliance that plugs into an outlet. It's very popular among hobbyists in the home automatic area because it's pretty cheap. `rest-mochad` is a REST API that sits in front of `mochad`. Mochad is a command-line utility that interfaces with a couple different X10 controllers (which transmit commands to the actual modules that are connected to the lights).

I've tested this pretty extensively and it works pretty well. Let me know if you have any questions. Thanks!


nfarina:
Beautiful, love it! Thanks for the PR!

edc1591:
Thanks!

PR #9, opened 2015-05-12
Add Wink Shim

maddox:
This adds a [Wink](http://www.wink.com) accessory shim. 

This only controls light bulbs right now, but it supports turning them on, off, and setting their brightness.


nfarina:
This is awesome. Thanks for the PR!

PR #12, opened 2015-05-19
Platforms

maddox:
This is a rough first pass at adding support for full platforms.

Rather than adding items in `config.json` for every item on your HA network, this hopes to let homebridge find all of your items for you. The first example usage of this is Wink.

Wink support for specific accessories has been removed and replaced with a simple way to set it up as a platform.

Platforms have the same sort of pattern as accessories in that they take a json hash as an input. You can define it as anything you want for your Platform library to use what you need to fetch your items.

The most important function to implement is the `accessories()` function. Here is where you'll do your fetching of all of your accessories, turn them into the appropriate objects, and return them with the callback.

Once the accessories are returned, they are looped through and registered in the same exact manner that accessories are registered on an individual basis.

There's also no mention of this support in the README yet. I figured I'd wait till it was ready to land before writing that up.

This idea came out of the conversation on #10.

Comments welcome!


maddox:
The Wink platform example only supports lightbulbs. But there's nothing stopping another library from returning different TYPES of accessories from the `accessories()` function. Then, combined with the right definitions at the bottom of your library, you can support everything your platform does, hence bridging everything pretty much automatically.

nfarina:
This is fantastic!! I think this is perfectly done.

Would the `accessories/Wink.js` accessory still be needed? Or could you always get everything via the platform approach?

maddox:
@nfarina I debated it, and possibility turning it into a more general library and putting it under `/lib`, but seeing that this is a HomeKit bridge, I can't come up with a single reason you'd want cherry picked devices. 

This way is just way easier. Add creds and 💥 done. Why confuse with 2 different ways to add devices, right?

I figure, if it's a problem, we can just roll it back in.

nfarina:
Yes agreed with that conclusion. I only asked because accessories/Wink.js was still present in this PR :)

But yes, I think just a README update and this is good to go.

maddox:
> I only asked because accessories/Wink.js was still present in this PR :)

Nope, it was 🔥d in https://github.com/maddox/homebridge/commit/c4026e997af9ac13c6937fb3026f964b4b204453

It's just confusing cause the overall diff has it as a moved file with some changes. 

> But yes, I think just a README update and this is good to go.

Ok, i'll update the readme soon and submit it.

maddox:
Ok, the README was updated.

nfarina:
Ah gotcha, OK awesome. Looks perfect! Merging!

jnewland:
This is great. I know what I'm doing tonight!

PR #22, opened 2015-06-10
Added platform shim for Domoticz.

joepv:
I added a platform shim for controlling light/switches within Domoticz (http://www.domoticz.com) and removed a JSON markup error from config-sample.json and added a Domoticz example record.


nfarina:
Looks great, thanks for the contribution!!

Issue #5, opened 2015-03-03
Stimm for HTTPREQUEST ?


petkoe:
Hi, is it possible to do a shimm for a httprequest? 

Peter

tommyd75:
Here's what I used. 


  sType: types.LIGHTBULB_STYPE, 
    characteristics: [{
    	cType: types.NAME_CTYPE,
    	onUpdate: null,
    	perms: ["pr"],
		format: "string",
		initialValue: "Kitchen Light",
		supportEvents: false,
		supportBonjour: false,
		manfDescription: "Bla",
		designedMaxLength: 255   
    },{
        cType: types.POWER_STATE_CTYPE,
    	
               onUpdate: function(value)
{
    target = "off";
    if (value === true) {
        target = "on";
    }
    request("http://192.168.1.1/kitchen"+target);
},

nfarina:
It would be great to have this as an official "accessory"! Where you could set the target URL in the `config.json`.

tocks53:
Hello,
I also want to launch a http.

You can add an example in the config.json

thank you

nfarina:
Closed by #26!

PR #26, opened 2015-06-12
HTTP Accessory

maddox:
This is a first stab at supporting generic http requests as accessories. For now this just assumes light bulb support (like practically all the shims).

The idea behind the HTTP accessory, is that it can receive commands over HomeKit and corresponding URLs can be hit. This gives a lot of flexibility and allows almost anything to be bridged. 

This means you can create an accessory, expose it over HomeKit, and do things to it. What does it support?

• on
• off
• brightness

So what this really means is, you can create a generic HTTP accessories for ANYTHING YOU WANT,

## How Do I Set It Up?

Just add an HTTP accessory to your config. 

```json
{
  "accessory": "Http",
  "name": "Kitchen Lamp",
  "on_url": "http://127.0.0.1:4567/devices/23222/on",
  "off_url": "http://127.0.0.1:4567/devices/23222/off",
  "brightness_url": "http://127.0.0.1:4567/devices/23222/brightness/%b",
  "http_method": "POST"
}
```

* `name`: The name of the accessory. You'll use this via Siri.
* `on_url`: The URL that will be requested when requesting that the accessory is turned on.
* `off_url`: The URL that will be requested when requesting that the accessory is turned off.
* `brightness_url`: The URL that will be requested when requesting that the accessory is dimmed.
* `http_method`: The HTTP method that will be used when hitting the declared URLs.

The `brightness_url` accepts a variable in the URL. You can put `%b` anywhere in your URL and it will be swapped in for the value requested for the brightness (0-100).

## After This PR

In the future, I suppose we can support many many more accessory types in this shim, offering the ability to define it via the config json. This might allow different types of commands to be 'said' to control things.

I'd love to talk more about what we can do with that, but for now I think this allows for all kinds of cool stuff. I think this could be real cool in terms of scenes as well. I.E. `set up movie time` -> turn down lights, and turn on tv/receiver. Even if it means having to have ANOTHER intermediary server to hit and do these things, heh.

### DIFF
--- a/accessories/Http.js
+++ b/accessories/Http.js
@@ -1,0 +1,160 @@
+var types = require("../lib/HAP-NodeJS/accessories/types.js");
+var request = require("request");
+
+function HttpAccessory(log, config) {
+  this.log = log;
+
+  // url info
+  this.on_url = config["on_url"];
+  this.off_url = config["off_url"];
+  this.brightness_url = config["brightness_url"];
+  this.http_method = config["http_method"];
+
+  // device info
+  this.name = config["name"];
+}
+
+HttpAccessory.prototype = {
+
+  httpRequest: function(url, method, callback) {
+    request({
+      url: url,
+      method: method
+    },
+    function (error, response, body) {
+      callback(error, response, body)
+    })
+  },
+
+  setPowerState: function(powerOn) {
+    var url;
+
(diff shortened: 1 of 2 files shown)


nfarina:
Beautifully simple and elegant. Love it. This should close #5.

PR #35, opened 2015-06-25
Read State

maddox:
:construction::construction: **Don't merge this yet, since there's some concern for missed cases being handled, as [mentioned here](https://github.com/KhaosT/HAP-NodeJS/pull/82#issuecomment-115113385).**:construction::construction:

This adds the ability to read state from accessories. With KhaosT/HAP-NodeJS#82 landed, we can now return requests for state via callbacks. This way your device lib can fetch state asynchronously. 

As an example, the Wink accessory/platform was updated to support fetching state for `poweredOn` and `brightness`.

### DIFF
--- a/app.js
+++ b/app.js
@@ -123,6 +123,7 @@ function createHAPServer(name, services) {
         //loop through characteristics
         for (var k = 0; k < services[j].characteristics.length; k++) {
             var options = {
+                onRead: services[j].characteristics[k].onRead,
                 type: services[j].characteristics[k].cType,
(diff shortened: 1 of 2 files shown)


maddox:
whaton
lighton

nfarina:
So. Awesome.

maddox:
howbright

Pretty cool.

maddox:
This now includes the final working version of the async `onRead` stuff.

So I think it might be ready to merge :dancer:

nfarina:
:dart:

PR #36, opened 2015-06-26
HTTP Shim Read States

maddox:
This adds the ability to set URLs to read the powered state and brightness state.

This is starting to get a little weird. While the commands (`on`, `off`, `setBrightness`) were easy, because they simply hit a url and thats it...reading is more complicated. This is an abstract implementation, but at the same time, we need to somehow be specific in order to fetch values.

Since we're reading something, we have to decide how we want this to work. Somehow the value has to be read and passed back.

As it is now, all the functions do is return the WHOLE body from the `GET` request. Kinda jank.

## Value Fetching Options
So I think there's 3 options here:

1. Just return the body of the response, and make people understand they have to return a single value in their respones.
2. Specify a key that people need to use in their response (this assumes JSON) that we'll read from.
3. Add a specific read key in the `config.json` that would be used to traverse JSON to fetch a key.

In the first two, it pretty much assumes the user **has** to control the response (hitting their own server they have control over). The third option means they can sort of specify things in a little more detail. But can this be reliable? What happens if the value they need to get to is inside an array or something.

This is all really tricky.

Comments **verrrrrrry** welcome.

### DIFF
--- a/accessories/Http.js
+++ b/accessories/Http.js
@@ -42,6 +42,24 @@ HttpAccessory.prototype = {
       }
     });
   },
+
+  getBrightness: function(callback){
+    if (!this.read_brightness_url) { callback(null); }
+
+    var that = this;
+    this.log("checking brightness level for: " + this.name);
+
+    this.httpRequest(this.read_powered_url, 'GET', function(error, response, body){
+      if (error) {
+        that.log('http get brightness level failed:', error);
+        callback(null)
+      }else{
+        that.log('http get brightness level succeeded!');
+        callback(body)
+      }
+    });
+  },
+
   setPowerState: function(powerOn) {
     if (!this.on_url) { return; }
     if (!this.off_url) { return; }
@@ -173,6 +191,11 @@ HttpAccessory.prototype = {
         cType: types.BRIGHTNESS_CTYPE,
         onUpdate: function(value) {
           that.setBrightness(value);
+        },
+        onRead: function(callback) {
+          that.getBrightness(function(level){
+            callback(level);
+          });
         },
         perms: ["pw","pr","ev"],
         format: "int",
--- a/accessories/Http.js
+++ b/accessories/Http.js
@@ -27,6 +27,9 @@ HttpAccessory.prototype = {
   },
 
   setPowerState: function(powerOn) {
+    if (!this.on_url) { return; }
+    if (!this.off_url) { return; }
+
     var url;
 
     if (powerOn) {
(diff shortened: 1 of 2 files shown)


nfarina:
Oh, I didn't realize that the HTTP accessory was masquerading as a "light bulb" - that's a bit strange. Although I'm not sure what else would be appropriate besides "switch".

For reading state, I would worry that options 2 and 3 will lead us into a path of adding more and more parsing features and "language" for extracting and converting whatever existing APIs people are trying to use.

Although maybe it could be made generic enough by having you write a regex or something - if it matches, then it's on? Although that doesn't help with brightness.

This is pretty squirrely!

maddox:
> Oh, I didn't realize that the HTTP accessory was masquerading as a "light bulb" - that's a bit strange. Although I'm not sure what else would be appropriate besides "switch".

Yeah, the plan was to actually add more and more device types, so there could also be `up_url`, `down_url`, so you could control all the types with the different commands. Commands here being "what word do you want to say".

The idea of the HTTP shim was that it would allow you to quickly hack out accessory support for yourself, without having to: 

1. create your own shim
2. worse, create own lib if there wasn't one already.

It would just allow anyone to come in and drop in the URLs that need to be hit and voila.

There of course is, no authentication story, lol. Again, this was a first quick pass to get something going. And it was all fine when they were blind requests shooting off to turn things on/off.

Now, this ALSO of course lets you lean on this pattern to let you hit any sort of endpoints you want to hack your own things. So you can just use it as a hack. Sort of what #5 was asking for.

BUT, this all falls apart for reads.

So, it seems to me, that if you're going this far in depth with reads, shouldn't you just make your own shim that just uses the request object on it's own, doing the domain specific stuff? Probably so :/





> For reading state, I would worry that options 2 and 3 will lead us into a path of adding more and more parsing features and "language" for extracting and converting whatever existing APIs people are trying to use.
> 

Boy do I agree with that.

> Although maybe it could be made generic enough by having you write a regex or something - if it matches, then it's on? Although that doesn't help with brightness.
> 
> This is pretty squirrely!

Ok, so maybe I made some points up top, and this shim shouldn't even HAVE read. Maybe it should exist solely for shooting off requests. By not supporting it at all, we don't have to debate anything, lol. I'm a fan of not doing anything at all unless there's a solid logical thing to do.

As another argument for keeping this shim, the new IFTTT Maker channel that supports webhooks would still work totally fine using this simple HTTP shim. So yeah, maybe we should just keep it simple like it was before.

</ramble>

nfarina:
Good points. I think that a generic HTTP shim is more useful the simpler it is. If it gets complicated and supports lots of device types and reading/parsing then you're just spending a lot of time writing config files instead of just writing code :)

maddox:
Yeah, ok, I'm gonna just close this.

PR #37, opened 2015-06-27
Adding support for ISY.

rayzorben:
This adds support for Universal Devices ISY controller which communicates with an array of Insteon and Z-Wave devices. Initial commit is tested working for controlling Light Bulb (dimming), Switch, and Ceiling Fan devices. It uses the published REST API available at http://wiki.universal-devices.com/index.php?title=ISY_Developers:API:REST_Interface

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -21,6 +21,14 @@
             "name": "Domoticz",
             "server": "127.0.0.1",
             "port": "8005"
+        },
+        {
+            "platform": "ISY",
+            "name": "ISY",
+            "host": "192.168.1.20",
+            "port": "8000",
+            "username": "username",
+            "password": "password"
         }
     ],
 
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "request": "2.49.x",
     "node-persist": "0.0.x",
     "xmldoc": "0.1.x",
+    "xml2js": "0.4.x",
 
     "carwingsjs": "0.0.x",
     "sonos": "0.8.x",
--- a/platforms/ISY.js
+++ b/platforms/ISY.js
@@ -1,0 +1,385 @@
+var types = require("../lib/HAP-NodeJS/accessories/types.js");
+var xml2js = require('xml2js');
+var request = require('request');
+var util = require('util');
+
+var parser = new xml2js.Parser();
+
+
+var power_state_ctype = {
+    cType: types.POWER_STATE_CTYPE,
+    onUpdate: function(value) { return; },
+    perms: ["pw","pr","ev"],
+    format: "bool",
+    initialValue: 0,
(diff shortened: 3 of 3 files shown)


nfarina:
Awesome, sounds great!

maddox:
I saw a lot of issues in this code. As mentioned [here](https://github.com/KhaosT/HAP-NodeJS/issues/91), the `onRead` was apparently not being called right.

[here](https://github.com/nfarina/homebridge/pull/37/files#diff-12025cc17baf8c499491e5def13cb0c2R256):

```js
 onRead: function() {
   return this.query();
}
```

The function isn't defined right. The function has a callback, and you're supposed to return the value with the callback, not a return.

```js
 onRead: function(callback) {
   callback(this.query());
}
```

though, `query()` wasn't defined to use a callback either, which it probably should. to keep things from blocking, (or even working correctly at all right?)

From what I can tell just by breezing over this pull though, its not exactly working.

rayzorben:
I changed it as suggested to utilize the callback. However, I simply put a console.log in there

    onRead: function(callback) {
        console.log("Query");
        callback(this.query());
    }

I do not see the log of 'Query' ever showing up. I also log out in my query() code and it does not run either. I guess what I am asking is - when is this supposed to be executed? I never see it getting called.

maddox:
It gets called when homekit queries the current state of your device. 

For example, ask Siri which light are on. It will get called. 

As for the logs, I'm not sure you can log via that way, I've seen them not show up like that. Check the logging pattern used in the other Accessory shims. 

Do this, instead, just return a value instead of running the query (which isn't working anyways because it doesn't support a callback). 

So just use callback(true). In the logs you should then see it.

nfarina:
If `onRead` is not being called, then that is the first thing to investigate. From your code it does look like it's defined in the right place. This may be a silly question but did you make sure to update git submodules to get the latest version of `HAP-NodeJS`?

rayzorben:
Really appreciate your help. Still not seeing what I think I should be seeing. Looking at the code:

    console.log("Characteristics.js:valueForUpdate(): called, Siri has asked for the accessory's status");

    if (this.onRead) {
        console.log("Characteristics.js:valueForUpdate(): invoking callback");

My log is showing the first message, but it is not showing the second message (invoking callback). I have simplified it as you recommended. I have verified that the version of HAP-NodeJS is correct and what is in Characteristic.js matches what is in your commit.

rayzorben:
Ignore me - I pulled the latest HAP-NodeJS, but not the latest HomeBridge as I was in the middle of a commit and didn't pull it back down after. So I missed the update for the onRead to HomeBridge which copies it across. It is calling the callback now.

nfarina:
:+1:

PR #39, opened 2015-06-28
Logitech Harmony Platform [WIP]

madmod:
I've added basic on/off control of Harmony activities as switch type devices. (Turn on Roku, Turn on my Playstation, Turn of the TV, etc) I'm currently having issues with Siri conflicting with names and being generally unreliable. I can get it to work once in a while, but I have to say things just the right way or it conflicts. I still need to add reporting device state on init and change before this is ready. I might also add some kind of an alias config for activities because saying "Turn on the play a game" doesn't make sense. :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,14 @@
     "url": "git://github.com/nfarina/homebridge.git"
   },
   "dependencies": {
-    "request": "2.49.x",
-    "node-persist": "0.0.x",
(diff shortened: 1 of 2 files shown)


madmod:
I had to change the repo I forked from so the repo that this was in is gone now. I will close this and re-open a PR from my new repo. :confused:

PR #28, opened 2015-06-14
Philips Hue integration

stephenyeargin:
Uses the [node-hue-api](https://github.com/peter-murray/node-hue-api) package to interact with the bridge.

I've gotten it to where the devices will start appearing in Inteon+. However, when adding it, it complains that each device "does not conform" to HomeKit and removes it. I'm mostly soliciting help at this point if anyone can salvage this PR into something that works with their bridge.

- [x] Hue devices appear in HomeKit app (Insteon+)
- [x] Hue devices add to the database (without "does not conform" error)
- [x] Hue lights can be turned off/on
- [x] Hue lights can have colors changed in-app
- [x] Hue lights can have brightness changed
- [x] Hue lux lights can be turned off/on
- [x] Hue lux lights can have brightness changed
- [x] Code is cleaned up to fully use the `node-hue-api` package (instead of direct bridge calls)
- [x] Bridges can use the `detectBridges` API endpoint to further ease setup process (stretch goal)

When it merges, it fixes #21

---

Note for development: In the event Insteon+ or other application deems the device incompatible, simply delete the `./persist` directory (will nuke other things) and they will show back up on next launch.

PR #40, opened 2015-06-28
Simplify Hue for new users

nfarina:
- Sample config auto-discoveres Bridge IP
- Username left blank so that the server will prompt you
- Hue platform walks you through username creation (in as friendly a
way currently possible)

/cc @stephenyeargin, #28

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -25,8 +25,7 @@
         {
             "platform": "PhilipsHue",
             "name": "Phillips Hue",
-            "ip_address": "127.0.0.1",
-            "username": "252deadbeef0bf3f34c7ecb810e832f"
+            "username": ""
         },
         {
             "platform": "ISY",
--- a/platforms/PhilipsHue.js
+++ b/platforms/PhilipsHue.js
@@ -10,6 +10,15 @@
 //     }
 // ],
 //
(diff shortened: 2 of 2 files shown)


stephenyeargin:
:+1: I think you just cleared it off @madmod's todo list :smile:

madmod:
:+1: Looks good. I can't try it out right now, but I will test it in a few hours. I think if we could add some ability for saving to the config.json we wouldn't need to stop running if this fails.

nfarina:
Yes agreed it would be great to have either a method for updating `config.json` (which I'm not super into, it feels wrong to fiddle with "your" config) or a method for accessory storage of preference data. Or an official method of "first time setup" for accessories. Lots to think about...

maddox:
👎 on writing to the config
👍 to some sort of local storage (if more than 2 or 3 cases can be come up with)
👍 to not even thinking about it on this pull 😄

maddox:
Ps: this pull is cool. Love it.

nfarina:
Thanks but it's just icing on the cake - thanks to @stephenyeargin and @madmod for doing all the real work supporting Hue! This was the last thing in my home that wasn't hooked up to HomeKit :smile_cat:

Issue #42, opened 2015-06-29
Philips Hue rate limiting


madmod:
The Philips Hue bridge only allows a certain number of requests each second. When I change a scene it is sending an API call for each characteristic of each bulb. A possible solution to this might be to group characteristic changes for each bulb into a single call. I'm trying to figure out a good way to cache characteristic changes by bulb for a few ms before sending them to the bridge. I am working on a PR but don't have a lot of time today so it may have to wait until tomorrow.

alistairg:
I can't wait for this PR! :)

Are you also considering supporting Hue Groups?

madmod:
I will try to get around to this today or tomorrow. 

I'm not sure what would be required to represent Hue groups in HomeKit. I think the easiest solution would be to manually create zones in the insteon+ app. If there is some way push zones to HomeKit I will look into adding it.

It would also be great if Hue scenes could be synced to HomeKit scenes somehow. I think this would require some more structure around the setup process to allow for merging scenes from platforms to existing HomeKit scenes.

alistairg:
I think the challenge is that Hue sees scenes / groups as “devices” as far as HomeKit is concerned. HomeKit’s scenes / zones are just logical groupings of accessories, services, and characteristics. 

We’d likely need to create a “group device” for each of up to 16 groups in Hue and map them back to HomeKit...

madmod:
I think you're right that we would need to create a "group device" right now. I'm not sure that is something we should do because it isn't the HomeKit way of doing things. A better solution to that problem might be an iOS app to import groups/scenes into HomeKit from homebridge platforms through some kind of API.

Tom2Drake:
Is there any way to create a workaround that implement a delay between the command communication the the hue bridge? This may will "solve" the issue. When i try to turn off all my 14 Hue devices at once with siri, 3 of them didn't change the state.

nfarina:
Does the hue API allow for a single "batch" of changes at once? I can think of two solutions to this problem:

  1. New HAP-NodeJS API for handling "batch set" of characteristic changes
  2. New helper function to wait until next runloop and accumulate changes ourself

Tom2Drake:
Well, regarding the API i just found olny the follwing:

"How many commands you can send per second?

You can send commands to the lights too fast. If you stay roughly around 10 commands per second to the /lights resource as maximum you should be fine. For /groups commands you should keep to a maximum of 1 per second."

Tom2Drake:
[...]

PR #43, opened 2015-06-29
Logitech Harmony platform

madmod:
This is a copy of #39 because I had to delete my old repo which was forked from another fork. :confused:

> I've added basic on/off control of Harmony activities as switch type devices. (Turn on Roku, Turn on my Playstation, Turn of the TV, etc) I'm currently having issues with Siri conflicting with names and being generally unreliable. I can get it to work once in a while, but I have to say things just the right way or it conflicts. I still need to add reporting device state on init and change before this is ready. I might also add some kind of an alias config for activities because saying "Turn on the play a game" doesn't make sense. :smile: 

Features:
- [x] Turn on activities
- [x] Turn off all activities
- ~~Turn on devices?~~ I decided that activities are good enough for me. If someone wants this I will add it. (Might lead to too many conflicts depending on how strict Siri is.)
- ~~Optional volume control (Tell Siri the audio source is a dimmable light.)~~ I decided this was too much of a hack.
- ~~Report initial activity state~~ Maybe in a future version
- ~~Poll for activity changes (stretch goal)~~ I don't really need this right now. When we get some built in polling functionality from homebridge or HAP-NodeJS I might add it.
- [x] onRead activity state support

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Since Siri supports devices added through HomeKit, this means that with HomeBrid
  * _Siri, turn on the Dehumidifier._ ([WeMo](http://www.belkin.com/us/Products/home-automation/c/wemo-home-automation/))
  * _Siri, turn on Away Mode._ ([Xfinity Home](http://www.comcast.com/home-security.html))
  * _Siri, turn on the living room lights._ ([Wink](http://www.wink.com), [SmartThings](http://www.smartthings.com), [X10](http://github.com/edc1591/rest-mochad), [Philips Hue](http://meethue.com))
+ * _Siri, set the movie scene._ ([Logitech Harmony](http://myharmony.com/))
 
 If you would like to support any other devices, please write a shim and create a pull request and I'd be happy to add it to this official list.
 
--- a/config-sample.json
+++ b/config-sample.json
@@ -42,6 +42,10 @@
             "port": "8000",
             "username": "username",
             "password": "password"
+        },
+        {
+            "accessory": "LogitechHarmony",
+            "name": "Living Room Harmony Hub"
         }
     ],
 
--- a/package.json
+++ b/package.json
@@ -11,19 +11,21 @@
   },
   "license": "ISC",
   "dependencies": {
-    "hap-nodejs": "git+https://github.com/khaost/HAP-NodeJS#2a1bc8d99a2009317ab5da93faebea34c89f197c",
     "ad2usb": "git+https://github.com/alistairg/node-ad2usb.git#local",
-    "request": "2.49.x",
-    "node-persist": "0.0.x",
-    "xmldoc": "0.1.x",
-    "node-hue-api": "^1.0.5",
-    "xml2js": "0.4.x",
     "carwingsjs": "0.0.x",
+    "elkington": "kevinohara80/elkington",
+    "hap-nodejs": "git+https://github.com/khaost/HAP-NodeJS#2a1bc8d99a2009317ab5da93faebea34c89f197c",
+    "harmonyhubjs-client": "^1.1.4",
+    "harmonyhubjs-discover": "git+https://github.com/swissmanu/harmonyhubjs-discover.git",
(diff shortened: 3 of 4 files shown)


maddox:
There's conflicts in this pull. Could you merge master first and resolve them? Thanks!

madmod:
Will do.  This needs a bit more work before it's ready.  I think I'm having some hardware issues with my Logitech Harmony Smart Control being unreliable.  I will get back to it in the next few days after I fix #42.  If someone else has one and could try it out that would be great.  If not then consider buying one because they are awesome.  :wink:

maddox:
I've been eyeing [this one](http://www.amazon.com/Logitech-Harmony-Home-Control-Devices/dp/B00N3RFC4G/ref=sr_1_1?ie=UTF8&qid=1435600246&sr=8-1&keywords=harmony+home) for months, now that Harmony finally got their act together and understands the proper layout of a remote (let our thumbs rest on the pause button!)

Now that I see we can hit these hubs over IP, I'm SUPER excited!

This basically means we can bridge IP to IR, wowzers!

madmod:
It gets even better because it can do bluetooth HID and talk to certain APIs over wifi. I have mine controlling my PS4 and MacBook over bluetooth and my Roku over wifi which means I can tell it to go to directly to Netflix, Hulu, or whatever.  Also its great using my iPhone as a keyboard and trackpad for my MacBook over bluetooth.  I live for the simple things. :smile: (I don't think that is exposed by the node API I'm using at the moment but maybe I will add it at some point.) 

I was annoyed when I found out that I can't map buttons on my older Smart Control to my Hue lights. Kind of lame that it requires a whole new unit but I understand the need to up sell. I ended up making a custom node-red flow that polls the current activity every second for my "lights on/off" mode, then goes back to the last activity. Not great but it gets the job done.

maddox:
Yeah, the IP stuff talking to my LG tv would be awesome. Being able to get the on/off state of my TV is a big deal when it comes to automating my lights off in my family room. I want them off at 11pm, but if my TV is on, I don't. This is cool.

I suppose I could bridge directly to my LG tv, but this is even easier. It gives a single place to talk to multiple devices.

madmod:
Your point about power state is a good one. I will add it to my feature list.

madmod:
I was able to fix the issue with my Harmony hub and now this is working great. I've added the onRead functionality but I don't see any changes in my Insteon+ app and I'm not sure when it decides to poll for updates. If there is a way to push events to HomeKit with homebridge I can look into adding event notifications to [harmonyhubjs-client](https://github.com/swissmanu/harmonyhubjs-client) so that we can use the activity state for triggers as @maddox described.

madmod:
Also here is a fun tip for anyone who decides to use this.

I created an activity in the Harmony app that has a sequence to turn on my TV, adjust the volume all the way down then back up by a certain amount, and send a series of button presses to my Roku that open a news app and play the top video.

It will be pretty great when it all works with a HomeKit scene to turn my Hue lights on and play news tomorrow! :smile: 

Also a hack to get Siri to understand that you are talking about a HomeKit device is to say "custom" after the name. For example saying "Hey Siri, Turn on Roku custom" will work with HomeKit even if you have the Roku app installed.

nfarina:
Sounds awesome :) I'm tempted ... 

This ready to merge?

madmod:
Yes it is. :smile: 

I put a scene as the example in the readme because using the activity names with Siri isn't straightforward. I can't simply tell Siri to turn on the Playstation, Roku, or Television because it always conflicts with app names.

I still have a bit of trouble with it being slow and unresponsive sometimes, but I think that is a hardware or API issue, not an issue with the bridge code. I also have responsiveness issues with the official app and even the remote occasionally. Perhaps my repeated testing shows the issue more than normal use would.

It would be great if someone else who has a Harmony Smart Control could test this out and see if they have any issues. Run homebridge with `DEBUG=harmonyhubjs* npm start` to get more info on what is going on with the API and try turning a few activities on and off.

PR #45, opened 2015-06-29
Changed approach to getting accessories

### LINKED ISSUE #44: Connection Reset interacting with Hue
I'm getting this odd error interacting with the two Hue hubs in my home. Both hubs are running the latest firmware, and both happily went through the initial linking / user generation process.

Any thoughts? :|

```
[Media Room Hue] Initializing PhilipsHue platform...
[Media Room Hue] Fetching Philips Hue lights from 172.16.1.23, user <Removed>...
Loading 0 accessories...
/Users/alistair/Dropbox/Development/Node/homebridge/node_modules/node-hue-api/node_modules/q/q.js:126
                    throw e;
                          ^
Error: connect ECONNRESET
    at exports._errnoException (util.js:746:11)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1010:19)
```

alistairg:
Avoid hitting rate limiting iterating over bulbs for installations with many hue bulbs on a hub.

This changes the approach to pull a (more expensive) full list of bulb states, rather than iterating over each returned bulb and getting its state.

This will resolve #44 - tested with multiple hubs with 80+ hue bulbs.

### DIFF
--- a/platforms/PhilipsHue.js
+++ b/platforms/PhilipsHue.js
@@ -102,25 +102,28 @@ var locateBridge = function (callback) {
 };
 
 PhilipsHuePlatform.prototype = {
+
   accessories: function(callback) {
     this.log("Fetching Philips Hue lights...");
     var that = this;
     var getLights = function () {
       var api = new HueApi(that.ip_address, that.username);
-      // Connect to the API and loop through lights
-      api.lights(function(err, response) {
+
+      // Connect to the API
+      // Get a dump of all lights, so as not to hit rate limiting for installations with larger amounts of bulbs
+
+      api.fullState(function(err, response) {
         if (err) throw err;
-        response.lights.map(function(light) {
-          var foundAccessories = [];
-          // Get the state of each individual light and add to platform
-          api.lightStatus(light.id, function(err, device) {
-            if (err) throw err;
-            device.id = light.id;
-            var accessory = new PhilipsHueAccessory(that.log, device, api);
-            foundAccessories.push(accessory);
-            callback(foundAccessories);
-          });
-        });
+
+        var foundAccessories = [];
+        for (var deviceId in response.lights) {
+          var device = response.lights[deviceId];
+          device.id = deviceId;
+          var accessory = new PhilipsHueAccessory(that.log, device, api);
+          foundAccessories.push(accessory);
+        }
+        callback(foundAccessories);
+
       });
     };
 


madmod:
:shipit: Looks good! I can test it if someone wants to buy me 36 more Hue lights! :laughing:

nfarina:
Wow, 80 bulbs, you are all in!

alistairg:
Yeah, I'm a little home-automation crazy ;p.

madmod:
Did you replace the lightswitches in your house also? That is my biggest annoyance with Hue right now. The Hue Tap is not responsive enough to be a good replacement.

alistairg:
"Pre-hue" I changed everything for Insteon switches. My plan is to hardware the lighting circuits "on" then add a software response to the logical switch for a given light, as all the insteon switches now have to be left permanently on.

At the moment, all lighting scenes are triggered by a combination of motion, time of day, and security system status as you move through the house - so it doesn't bother me too much. I have a custom iOS app with wall-mounted iPads to control most scenes for the occasion the switch would be needed.

You're right though - it is infuriating. I'd hoped that Wink would solve that problem with their "smart switches", but they're frustratingly "dumb" when it comes to Hue.

nfarina:
I'm very intrigued by these: http://www.slashgear.com/philips-hue-dimmer-switch-and-brighter-bulb-incoming-05387038/

If I can work up the energy to replace / coverup my old switches, I might go the 80+ bulbs route myself :bulb:

alistairg:
Interesting! Only real downsides I can see are that they're not mains powered (I have battery powered shades that have held up for two years, so likely no biggie) and that it may not be possible to pair them to "scenes".

I'm hopeful that the Insteon Pro hub (w/Homekit) I'm waiting on will provide events fast enough that I can trap them with HomeKit, and trigger associated lighting updates. The new HomeKit stuff in iOS9 finally makes triggers more useful....

maddox:
lol, that thing is HUGE. Why is no one making wall switch replacements :/

madmod:
That looks excellent! I think a great mod for that would be a custom wall plate that magnetically holds it in place. Inductive charging would be nice also but probably not worth it as I bet it's pretty low power anyway.

alistairg:
The one I'm holding out for being awesome is http://www.umbrela.co.... ;)

maddox:
To be honest, the easiest way to this stuff (in a custom way that you def want!) is to just use an older tablet.

I set up a home [Dashing dashboard](http://dashing.io). It runs on an iPad mini that's mounted to the side of my refrigerator in our kitchen. It's great for glance-able info for our household, and a PERFECT place to put HA buttons. It really ended up being a game changer when I added buttons to control my Wink accessories. To me, this is a way better way to control stuff than finding an app inside your phone that's inside your pocket. Siri is a close second if not # 1.

This is what it looks like:

screen shot 2015-06-30 at 4 35 20 pm

Most of the widgets on here I've customized or made myself. I made the Wink buttons. I also spun up a simple full screen browser iOS app project to run it. Since there's no really great full screen browsers on iOS (why?!). 

But the beauty of this is...its just a web site. So you could run it with a shitty Android tablet. You can pick something up for like $50 on amazon that would probably run this just fine. I like the idea of spreading these around, but in smaller form factors.

It's kinda cool.

nfarina:
Wow that is _gorgeous_. Want.

maddox:
The cool thing is, I set the UIWebView to paginate across pages, so instead of scrolling loosly, it scrolls per page, like panes. It feels much more legit. This allows me to actually have pages and pages of buttons if I wanted to. You can see how I hide some other ones at the bottom here (lots of widgets as filler, heh):

screen shot 2015-06-30 at 4 39 54 pm

maddox:
@nfarina Dashing projects are kinda weird, in that you GENERATE your own, from the gem. So sharing is a little weird, in so far as forking. But I could make mine public, and you can just steal all my custom and custom-ized widgets. I'll audit everything and see if I can make it public. 

I did most everything with configs, so I think its probably fine.

madmod:
I'm considering buying a [Philips Living Colors Remote](http://www.ebay.com/sch/i.html?_from=R40&_trksid=p2047675.m570.l1313.TR0.TRC0.H1.Xphilips+living+colors+remote.TRS0&_nkw=philips+living+colors+remote&_sacat=0) from ebay because the Hue Tap is not very responsive. Also being able to control the color of my lights dynamically without an app sounds nice.

alistairg:
Very cool! I really like the content tiles idea.... My current app is a little old now, it's all objective-c and interacting with a custom Python backend I wrote for Indigo.

My plan is to rewrite this all on top of HomeKit, and retire Indigo. I'm using old / refurbished iPads mounted in LaunchPort induction charging docks that snap to a few wall mounted chargers around the house (and they snap to the fridge well for recipes while cooking!).

The Hue support in this isn't that well thought out - the way groups works with Indigo is a little hokey (admittedly I wrote that code, so I've got no-one else to blame!). 

I also want to use TouchID to control the alarm, as well as iBeacons in rooms so that I can correctly prompt for the right controls on Apple Watch.... more code I haven't yet written ;p.
img_0006
img_0002
img_0003
img_0004
img_0005

nfarina:
@alistairg Follow up question - is it correct that, in order to add your 80 Hue bulbs the way things are today, you have to add each one separately and enter the PIN code like ... 80 times in a row?

alistairg:
Yup :(

I mostly use groups on the hue hub to control them at the moment,
which gives a seamless effect when changing them. Will likely need to
visit that, if just for my own installation, to make them usable....

nfarina:
So for Hue Groups, you only need to add a single group and not all the bulbs contained?

alistairg:
That's my theory, yeah. It's the way I control things now via Indigo....

madmod:
To provide smoother group changes we could queue up changes to Hue bulbs over the course of a few ms with a resetting timeout every time a new characteristic is changed. Once all the changes are queued we could then try to match the lights to existing Hue groups, then fallback to individual bulb changes.

The same technique could possibly work for matching scene changes to Hue scenes for better performance. Changing individual characteristics of even 10 bulbs could result in noticeable lag because its unlikely that each bulb would have unique color values.

There are three ways I can think of to match scenes in HomeKit to scenes saved on the Hue bridge.

1. Make the scene names the same in Hue and HomeKit. This has the advantage of not requiring homebridge to collect the details of each scene which is complicated. (see 2) The obvious disadvantage is that you couldn't have a "Movie" scene in homekit that starts the "Night" scene in Hue. Also you would probably want to manually start the Hue scene before programming a HomeKit scene so that the values would match what was actually happening.

2. Match the brightness, hue, and saturation values for each scene in Hue and HomeKit. Because the Hue API has no way to get color values for an entire scene it would require actually running each scene and then reading each light to get the exact values. This could be ran as part of the setup process and would take about 10 seconds for each scene depending on the transition times. We wouldn't have a way to know transition times so the automated discovery could fail for slower scenes. These characteristic values would then be saved for later use and manually updated by the user or updated on startup when a new Hue scene is added. 

3. If there is a way to detect when a new HomeKit scene is added or updated we could mirror those in the Hue bridge for smooth transitions. This would have no downsides other than not respecting modifications made in Hue apps.

The first two are non-ideal and it might be worth leaving the slow transition of per-bulb/group changes to avoid the complexity in setup and in maintaining two sets of scenes.

Also consider the edge case where you set one or two characteristics on a group of lights but one or two other characteristics are different for every light. We could apply a group change for the characteristics that are the same and then individual bulb changes for the unique values. This would still not produce the same transition as a Hue scene change but may be smoother than a per bulb change.

PR #46, opened 2015-06-30
Adding support for arming Elk M1 security system.

rayzorben:
Adds support to be able to arm an Elk M1 security system which has the M1XEP module. Might be a bad idea to store the pin in the text, if there are any ideas I would appreciate it.

### DIFF
--- a/accessories/ELKM1.js
+++ b/accessories/ELKM1.js
@@ -1,0 +1,126 @@
+var types = require("../lib/HAP-NodeJS/accessories/types.js");
+var elkington = require("elkington");
+
+function ElkM1Accessory(log, config) {
+  this.log = log;
+  this.name = config["name"];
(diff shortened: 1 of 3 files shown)


rayzorben:
Note, I intentionally am not allowing disarm even though the pin is available. I do not want somebody to pickup my phone and say "Turn off Security System".

nfarina:
Looks good - I think when iOS 9 is released it would be good to redo this and the Xfinity accessory to use the new "Alarm System" type instead of abusing "switches" :) Plus I would imagine that interacting with "Alarm System" would require you to unlock your phone first similar to Door Locks.

rayzorben:
I agree, when I can get ahold of iOS9 I will rewrite it to use that (assuming HAP-NodeJS supports that.

PR #54, opened 2015-07-03
AD2USB support (Honeywell/Ademco alarms via Nutech AD2USB interface)

alistairg:
This accessory provides an iOS9-ready interface to an Ademco/Honeywell alarm system.

It supports:
- Arm (Away, Stay, Night)
- Disarm
- LCD status display
- Event support (push updates for panel changes)

It requires an AD2USB interface with the associated SER2SOCK interface, and connects via the network.

### DIFF
--- a/accessories/LiftMaster.js
+++ b/accessories/LiftMaster.js
@@ -83,19 +83,60 @@ LiftMasterAccessory.prototype = {
         // parse and interpret the response
         var json = JSON.parse(body);
         var devices = json["Devices"];
+        var foundDoors = [];
 
         // look through the array of devices for an opener
         for (var i=0; i<devices.length; i++) {
           var device = devices[i];
-          if ((device["MyQDeviceTypeName"] == "GarageDoorOpener") && ((that.requiredDeviceId == undefined) || (that.requiredDeviceId == device.MyQDeviceId))) {
-            that.deviceId = device.MyQDeviceId;
-            break;
+
+          if (device["MyQDeviceTypeName"] == "GarageDoorOpener") {
+
+            // If we haven't explicity specified a door ID, we'll loop to make sure we don't have multiple openers, which is confusing
+            if (that.requiredDeviceId == undefined) {
+              var thisDeviceId = device.MyQDeviceId;
+              var thisDoorName = "Unknown";
+              for (var j = 0; j < device.Attributes.length; j ++) {
+                var thisAttributeSet = device.Attributes[j];
+                if (thisAttributeSet.AttributeDisplayName == "desc") {
+                  thisDoorName = thisAttributeSet.Value;
+                  break;
+                }
+              }
+              foundDoors.push(thisDeviceId + " - " + thisDoorName);
+            }
+
+            // We specified a door ID, sanity check to make sure it's the one we expected
+            else if (that.requiredDeviceId == device.MyQDeviceId) {
+              that.deviceId = device.MyQDeviceId;
+              break;
+            }
+
           }
+
         }
 
+        // If we have multiple found doors, refuse to proceed
+        if (foundDoors.length > 1) {
+          that.log("WARNING: You have multiple doors on your MyQ account.");
+          that.log("WARNING: Specify the ID of the door you want to control using the 'requiredDeviceId' property in your config.json file.");
+          that.log("WARNING: You can have multiple liftmaster accessories to cover your multiple doors");
+
+          for (var j = 0; j < foundDoors.length; j++) {
+            that.log("Found Door: " + foundDoors[j]);
+          }
+
+          throw "FATAL: Please specify which specific door this Liftmaster accessory should control - you have multiples on your account";
+
+        }
+
+        // Did we get a device ID?
         if (that.deviceId) {
           that.log("Found an opener with ID " + that.deviceId +". Ready to send command...");
           that.setTargetState();
+        }
+        else
+        {
+          that.log("Error: Couldn't find a door device, or the ID you specified isn't associated with your account");
         }
       }
       else {
--- a/accessories/LiftMaster.js
+++ b/accessories/LiftMaster.js
@@ -9,6 +9,7 @@ function LiftMasterAccessory(log, config) {
   this.name = config["name"];
   this.username = config["username"];
   this.password = config["password"];
+  this.requiredDeviceId = config["requiredDeviceId"];
 }
 
 LiftMasterAccessory.prototype = {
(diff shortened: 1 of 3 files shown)


nfarina:
Looks good! How did you figure out the new characteristic/service constants?

alistairg:
Handily, the HomeKit Accessory Simulator seems to save accessories created using NSKeyedValueArchiver. It's fairly straightforward to write quick code that recreates the object (using the unpublished framework inside the application, and NSKeyedValueUnarchiver) to write accessory services in Swift - as well as iterate over the services and characteristics to find their UUIDs....

I'm working on accessories to control an alarm (this one), Somfy window coverings, and virtual "occupancy sensor" zones that'll be part of the alarm system. I merged the constants needed to do those things into KhaosT's underlying framework to make it easier for us :)

I'm currently adjusting the core "app.js" code to also load "bridge" accessory types, as that'll make more sense for the Alarm accessory type. To make sure I'm not doubling up effort - is that something you're already considering?...

nfarina:
I've played with bridged accessories but haven't really gotten it to work well yet. I love the idea - it seems like a perfect world would be to simply add a single "HomeBridge" bridge accessory to HomeKit instead of adding all your accessories one by one and typing in the PIN code over and over. But I don't fully understand the implications of this.

alistairg:
OK - I'll have a play. The advantage with the simulator is that it's possible to dynamically add bridged accessories at runtime, "child" accessories all go offline with the parent, plus there's only one PIN to enter ;)

maddox:
> I've played with bridged accessories but haven't really gotten it to work well yet. I love the idea - it seems like a perfect world would be to simply add a single "HomeBridge" bridge accessory to HomeKit instead of adding all your accessories one by one and typing in the PIN code over and over. But I don't fully understand the implications of this.

omg, yeah, this sounds amazing. This means we could keep a config somewhere, instead of having to configure HomeKit all the time. This is especially horrible when deving on this stuff. Tearing it down and building it back up is horrible.

What would be even better is if we could do this for actions/scenes too. 

It also means we could feed it these from ANOTHER project, that way we could truly have our stuff all in a single place.

nfarina:
I think this is definitely important to explore; added it to the [roadmap](https://github.com/nfarina/homebridge/wiki/Roadmap).

madmod:
In some cursory research I read that bridged accessories are going to be very limited because apple wants to incentivize hardware devs to be MiFi certified and use HomeKit. Not sure if those restrictions are technical or legal though. Probably a mix of both.

nfarina:
Ah damn. Link?

maddox:
> In some cursory research I read that bridged accessories are going to be very limited because apple wants to incentivize hardware devs to be MiFi certified and use HomeKit. Not sure if those restrictions are technical or legal though. Probably a mix of both.

This is true. http://9to5mac.com/2015/01/22/apple-details-homekit-compatibility-with-competing-home-automation-platforms/

> There are restrictions on which accessories can be bridged to HomeKit. For starters, Apple will not allow home automation accessories connecting over Wi-Fi, such as a Nest Thermostat, to be bridged. Those will have to go the HomeKit protocol route and become licensed MFi products.

> Apple will allow some Bluetooth LE accessories to be bridged, but only those that don’t offer users control of the home. For example, a non-MFi Bluetooth lightbulb product must not be bridged. Instead, like Wi-Fi products, it would have to go through the HomeKit protocol and get MFi certification. Apple’s restrictions on bridged accessories appear to be at least partially security-related. The company also notes, for instance, that any accessory regardless of how it connects cannot be bridged if it allows physical access to a home, such as a door lock.

Looks like a limitation on liability. 

This is why I was excited about `HomeBridge`. Because it means we can bridge OURSELVES, outside of these limitations. So, I guess if this bridging part still works inside the HomeKit world, then yeah, it's not in our interests to use it. :(

madmod:
[This article](http://9to5mac.com/2015/01/22/apple-details-homekit-compatibility-with-competing-home-automation-platforms/) makes it sound pretty limited. No thermostats or light bulbs. (at least not officially)

alistairg:
IMHO this is just a legal restriction imposed by Apple, to ensure they
keep control of the HomeKit standard. Otherwise someone could make an
MFi approved bridge (e.g home bridge on a raspberry pi) and they'd
lose out on the per-accessory revenue is bet they're getting.....

nfarina:
I think this is purely CYA and not a technical restriction (at least not at this time). In my testing, I can run `HAP-NodeJS` in "bridged mode" `node BridgedCore.js` and I only need to add the "Node Bridge" accessory, then I can lock and unlock the included sample Door Lock accessory at will.

maddox:
Not really. These limitations are on the licensees. IE, if you're a paid HomeKit bridge licensee (Insteon), you still can't bridge products that 'control' the physical home. IE, thermostats, smoke alarms, locks, home alarms. Things that that are in a safety context. This is about liability for Apple.

> Otherwise someone could make an
> MFi approved bridge (e.g home bridge on a raspberry pi) and they'd
> lose out on the per-accessory revenue is bet they're getting

We're doing that RIGHT now.

maddox:
> I think this is purely CYA and not a technical restriction (at least not at this time). In my testing, I can run HAP-NodeJS in "bridged mode" node BridgedCore.js and I only need to add the "Node Bridge" accessory, then I can lock and unlock the included sample Door Lock accessory at will.

Cool, so it still WORKS, and HomeKit Bridge makers just limit things manually then? Perfect for us! heh.

nfarina:
Yeah, I think it works fine (except for the untested potential issues in adding/removing accessories after adding the bridge).

madmod:
This is great because it will allow us to add/remove accessories dynamically without having to set it up on all of our HomeKit devices. Also we won't need to use a bunch of ports and resources running several servers at once. Are there any downsides to having every platform and accessory on the same bridge?

maddox:
What makes me excited is being able to have a central location of our devices defined somewhere, and allowing HomeBridge to just ADVERTISE them. Instead of having them in HomeKit AND somewhere else.

nfarina:
So here's an example of a problem here - try running `HAP-NodeJS` in bridged mode, add the bridge (which bridges a few sample accessories). Now you can do things like "Open the garage door" and it works behind the bridge.

However, now shut down HAP-NodeJS, remove/rename the `GarageDoorOpener_accessory.js` and restart the server. HomeKit doesn't "know" that the garage door is now gone. It still shows up in your HomeKit apps on your phone. And you can't Delete it yourself, because it was a bridged accessory. It's just sort of a zombie.

nfarina:
Moving this discussion to #58 :)

PR #56, opened 2015-07-04
Adding support for iOS9 "transport category"

alistairg:
- Accessories can specify a "this.transportCategory" constant from the HAP-NodeJS types.js. 
- iOS9 sees this transport type which Siri may well use to help understand the accessory type
- Defaults to iOS8 behavior ("Other" transport type) if none is specified

Updated NPM reference to HAP-NodeJS to include latest revision with these constants and underlying transport support

### DIFF
--- a/app.js
+++ b/app.js
@@ -54,7 +54,7 @@ function loadAccessories() {
         var services = accessory.getServices();
 
         // Create the HAP server for this accessory
-        createHAPServer(name, services);
+        createHAPServer(name, services, accessory.transportCategory);
     }
 }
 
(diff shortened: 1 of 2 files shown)


nfarina:
Interesting - is it something like, transportCategory helps determine the "primary" service type for a particular accessory?

alistairg:
Yeah - I think it’s something like that. It also makes browsing for accessories easier, as you know the “type” before pairing. I’m sure we’ll find out more once we see some “official” iOS9 apps….

nfarina:
Cool. Thanks for adding!

maddox:
Should we update platforms and accessories for this? Or does it just come free?

alistairg:
I didn't get around to doing that.... apologies :|

It's "simply" a case of adding a this.transportCategory = <TYPE>; to the appropriate platform. The types are all now in the HAP-NodeJS accessories/types.js file.

Issue #58, opened 2015-07-06
Explore "Bridged mode" server


nfarina:
(Discussion forked from #54)

So here's an example of a problem here - try running HAP-NodeJS in bridged mode, add the bridge (which bridges a few sample accessories). Now you can do things like "Open the garage door" and it works behind the bridge.

However, now shut down HAP-NodeJS, remove/rename the GarageDoorOpener_accessory.js and restart the server. HomeKit doesn't "know" that the garage door is now gone. It still shows up in your HomeKit apps on your phone. And you can't Delete it yourself, because it was a bridged accessory. It's just sort of a zombie.

madmod:
It looks like we need some kind of "cleanup" step on startup to un-register accessories associated with a bridge that aren't in the config.json. It [looks like](https://developer.apple.com/library/prerelease/ios/documentation/HomeKit/Reference/HMAccessory_Class/index.html#//apple_ref/occ/instp/HMAccessory/bridged) HomeKit has no way for a user to remove bridged accessories without removing the bridge. Removing the bridge each time we want to remove accessories might remove them from scenes, triggers, zones, and remove Siri names unless HomeKit uses the serial number or something to identify it rather than the bridge id. 

Is there a way to remove an accessory with HAP?

I think every instance of homebridge needs to remember what devices it has registered and remove the ones that aren't in `config.json` only if the user explicitly asks for it with a command line flag, as the file might be simply unavailable when homebridge is restarted. Once we get a web interface we could have more specific workflows to deal with new devices, changing names, etc.

nfarina:
So it turns out that _adding_ a new accessory creates an odd situation as well: a new accessory _does_ magically "appear" in the list, but everything is now sort of "mixed up". The names of the accessories are all wrong (there are two thermostats, no garage door) even though the underlying services are correct. But Siri doesn't work with the newly-added Garage Door at all.

iOS does all the `/accessories` API on HAP-NodeJS periodically so it does seem like this should work. Either it hasn't been tested much or we're doing something wrong.

madmod:
[...]

PR #59, opened 2015-07-07
Bridge support

alistairg:
Bridge support enables a platform to run all its accessories with a single PIN. Accessories added with subsequent starts should automatically show on iOS devices that are paired with the top level bridge accessory.

Changes:
1. Added "bridges" directory
2. Altered app.js to support loading bridges at startup
3. Adjusted PhilipsHue.js as a bridge platform. This may be unnecessary - the only difference is that the export is now "bridge" rather than platform
4. Added config-sample.json bridge sample

Note that I had to move some code around to get the requires and global tracking variables in the right places....

References #58


madmod:
Looks good to me. :shipit:

nfarina:
So this is great! Some concerns though - main one is that this doesn't solve the "stable accessory ID" problem referenced in [this comment](/nfarina/homebridge/issues/58#issuecomment-119024143). If accessories are added or removed behind the bridge, the user will need to remove the bridge entirely and re-add it (they'll also need to delete the `persist` directory to work around a bug with HAP-NodeJS).

In order to fix this we'll need to do some thinking about how to enforce stable accessory IDs and patch support for that into `HAP-NodeJS` if necessary.

Also I think it's not ideal to have a 3rd "bucket" for accessories in config.json - I'd actually rather standardize on _everything_ being bridged behind the scenes, and possibly eliminating standalone "accessories" altogether (see [roadmap discussion on that](https://github.com/nfarina/homebridge/wiki/Roadmap)).

For stable IDs - does each Hue bulb have some kind of hardcoded unchanging identifier we could leverage?

madmod:
The each bulb has a unique ID which is given as the `SERIAL_NUMBER_CTYPE` currently.

nfarina:
I wonder if unique serial numbers would make stable accessory IDs unnecessary. It's worth testing; my previous tests were with HAP-NodeJS's "sample accessories" which all have the same fake serial number.

madmod:
I think a bridge MAC address combined with a device ID and maybe a model number would be enough for devices that don't give a unique I'd like Hue. Assuming that the device ID is stable.

alistairg:
I'm going to kill this PR - this is too buggy in this format.... it shouldn't go into the mainline code. Apologies for the excited checkin!

nfarina:
No worries, it should be useful when we do this for real :) I'd just like to do a bit more testing to be 100% sure that stable accessory IDs are required. In my previous test all accessories shared the same default MAC identifier, so it's possible that if the MACs are unique (as they should be in real life) that the accessory IDs can change.

PR #61, opened 2015-07-07
Initial support for MiLight/LimitlessLED/EasyBulb/etc... bulbs using …

dotsam:
Initial support for MiLight/LimitlessLED/EasyBulb/etc... bulbs using the node-milight-promise node module


nfarina:
Thanks!

PR #63, opened 2015-07-09
Adding support for Telldus Live platform

johngson:
I've implemented support for the Telldus Live platform. It will only work with connected light bulbs but I'm planning to add support for other devices as well.


nfarina:
Cool!

Issue #53, opened 2015-07-02
Should be Accessory getServices be asynchronous? 


nikita-leonov:
Should getServices be asynchronous to support services discovery? As for by providing this interface we assume that list of services and characteristics is static which could be not always true.

nfarina:
Yes I believe it should be, especially to match the `accessories` call in Platforms. Perhaps there could create a new method `services` that passes along an async callback and `app.js` could prefer that before calling `getServices` synchronously.

nikita-leonov:
Here is a list of things that asynchronous on iOS — https://developer.apple.com/library/ios/documentation/HomeKit/Reference/HMAccessoryDelegate_Protocol/index.html

I think this behaviors should be implemented in a such way between bridge and devices as well.

nfarina:
That would be awesome - it's more a question for @KhaosT over at [HAP-NodeJS](https://github.com/KhaosT/HAP-NodeJS) though. I'm not sure what that library supports in terms of proactively notifying HomeKit about accessory state.

KhaosT:
@nfarina to support -[HMAccessoryDelegate accessoryDidUpdateServices:], you'll need to modify Advertiser.js, every time you want to force iOS device to fetch new services, you have to increase the "c#" number and update the TXT Record. Too bad that I can't make any changes right now, feel free to send PR to that repo :)

nfarina:
Closing this for now - `getServices` must be synchronous right now in order to serve `/accessories` efficiently.

wurmr:
Has there been any thought to revisit this issue? It seems to me that having getServices be async is the only way to allow a accessory to do discovery of unknown services on startup.

But it is entirely possible I'm missing something.

snowdd1:
@wurmr  
Hi, since API 2 is out (~ March 2016) there is the possibility to make platforms dynamic, so they can add devices right after startup, or whenever they are discovered. But the platform must take care of accessories that are gone, too. Search for dynamic in the issues, there are several of them.

PR #64, opened 2015-07-10
Add support for getting the state of the power to WeMo accessories.

hachidorii:
Hopefully I haven't done anything too terrible here. I'm not too familiar with JavaScript or Node but it's definitely working for me. The only thing that doesn't seem to work and I think it's been discussed before is that if I manually turn the power on the WeMo device (like using the WeMo app or physically turning it on) I'll have to kill whatever app I'm using and/or wait for it to repoll the state to reflect the changes.

### DIFF
--- a/accessories/WeMo.js
+++ b/accessories/WeMo.js
@@ -54,6 +54,28 @@ WeMoAccessory.prototype = {
     });
   },
 
+  getPowerState: function(callback) {
+
+    if (!this.device) {
+      this.log("No '"+this.wemoName+"' device found (yet?)");
+      return;
+    }
+
+    var that = this;
+
+    this.log("checking power state for: " + this.wemoName);
+    this.device.getBinaryState(function(err, result) {
+        if (!err) {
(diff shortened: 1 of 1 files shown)


nfarina:
Looks correctly implemented to me - I don't follow the "kill the app" part though - you mean Siri doesn't actually poll for state when you ask if something is recently turned on/off?

hachidorii:
Siri definitely polls right away and gets it correctly. Let's say I have a HomeKit app open and I go over and I manually press a light switch on. It's not going to reflect the new state inside of a HomeKit app because it's not continually polling the state. I'm not sure how to fix that.

nfarina:
Oh I see what you mean! Yeah I don't think that's going to work with the current underlying `HAP-NodeJS` lib. I don't think any of the "notification" type APIs have been built yet. There was a small discussion about this in #53.

Meanwhile if this works for querying via Siri then I'll go ahead and merge, thanks for the patch!

PR #70, opened 2015-07-18
Add support for Nest platform.

### LINKED ISSUE #30: [Feature] Support Nest
Possibly via this "unofficial" API (as getting access to the official API is quite onerous): https://github.com/johnwyles/node-nest-api

hachidorii:
Most support for Nest is here. Some things I've been unable to test, such as the response for the current_schedule_mode because I do not have AC. The responses seem to mirror the output of the target_temperature_type though and can verify the output there should be correct by the results of others taking a look at the nest api. So with any luck the responses from current_schedule_mode should be correct but I cannot verify it. Other things to note, it doesn't seem like the api provides near current updates and for whatever reason HomeKit's automatic conversion to Fahrenheit seems to be off by a smidgen (.5) when comparing the temperature to what's on the thermostat. The module I was using for nest support had to be forked in order to accept setting the thermostat to off so I forked it and the package in the json points to my fix in a repo of mine. Let me know what you think.

[Edit by @nfarina: Fixes #30]

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -3,6 +3,12 @@
 
     "platforms": [
         {
+            "platform" : "Nest",
+            "name" : "Nest",
+            "username" : "username",
+            "password" : "password"
+        },
+        {
             "platform" : "TelldusLive",
             "name" : "Telldus Live!",
             "public_key" : "telldus public key",
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "wink-js": "0.0.5",
     "elkington": "kevinohara80/elkington",
     "node-milight-promise": "0.0.2",
-    "telldus-live" : "0.2.x"
+    "telldus-live" : "0.2.x",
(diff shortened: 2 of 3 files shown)


nfarina:
This looks awesome and works great for me!! Last piece of my home that wasn't on homebridge :santa:

nfarina:
I think it's fine to point at your forked nest API until they merge your PR.

nfarina:
So it looks like HomeKit thermostats can be set to Celsius or Fahrenheit directly. Notice how in the HomeKit Accessory Simulator you can set the unit type. And in the JSON service info, there's a `TEMPERATURE_UNITS_CTYPE` setting, as well as a `unit:` property for the `TARGET_TEMPERATURE_CTYPE`. I'm trying to figure out how this is supposed to work though.

nfarina:
Moved last comment to #71.

PR #75, opened 2015-07-22
Add SSL and basic auth support to Domoticz shim

lukeredpath:
I've got a bit more testing to do of this as I'm having problems getting this working on the office network but everything starts up.

My server runs behind HTTPS and requires basic auth - I imagine this is a common configuration for anyone who has their server accessible outside their home network. This adds support for username, password and SSL. A bit of refactoring at the same time.

### DIFF
--- a/platforms/Domoticz.js
+++ b/platforms/Domoticz.js
@@ -79,7 +79,7 @@ DomoticzPlatform.prototype = {
 					if (json['result'] != undefined) {
 						var sArray=sortByKey(json['result'],"Name");
 						sArray.map(function(s) {
-							accessory = new DomoticzAccessory(that.log, that.server, that.port, false, s.idx, s.Name, s.HaveDimmer, s.MaxDimLevel, (s.SubType=="RGB")||(s.SubType=="RGBW"));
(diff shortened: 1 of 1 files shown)


nfarina:
I don't have this hardware so hard for me to verify this works :) I think the original contributor for Domoticz was @joepv - maybe he can chime in?

lukeredpath:
I can now confirm this works, for me at least. I did notice some issues getting things to appear using the Elgato Eve app on a very busy office network however I have a suspicion that this is due to discoveryd (can we blame all known network issues on discoveryd?). Got it working right now with my laptop tethered to my iPhone and it detects things easily. I will test at home with homebridge running on the same machine as my Domoticz install later.

Homebridge has basically saved me from writing something myself so thanks for that. I'm going to see if I can extend Domotciz support to things like temperature devices etc. I have Domoticz set up to talk to my Nest thermostat so I could use that. OTOH, I'd like to give the Homebridge Nest platform support a go too.

nfarina:
OK sounds good - let me know when you confirm that it works on your home network and will merge! I am always happy to blame `discoveryd`, it was certainly the cause of many issues for me :imp:

joepv:
This is on my "roadmap", but I'm very busy at the time. If others work on it before I can it would be awesome, else I work on it when I get some more spare time. :)

PR #77, opened 2015-07-23
Make Sonos a platform rather than an accessory

bezzers:
Replaces the Sonos accessory with a Platform that will pick up and set up all Sonos boxes on the network automatically. Previously the Sonos accessory code would pick up only one of the Sonos boxes in an unpredictable way.


nfarina:
I think this is the correct approach. I will try this at home today!

bezzers:
I added a missing line of code. Should actually work now!

farberm:
Newbie here. I have Sonos working as an accessory but trying to convert to a platform

I have added the platform info into the config file for sonos. I am not sure how I create a sonos.js, sonos platform handler and delete as an accessory. Is that down by placing all three sets of code into the config file of do I do that somewhere else thru terminal commands?

Sorry about the newbie simple questions

PR #79, opened 2015-07-23
Add Tesla accessory

bezzers:
Adds a new accessory shim for managing climate control on a Tesla Model S


nfarina:
Hm, looks like I'm going to have to get a Tesla in order to properly test this :car: :moneybag:

bezzers:
The sacrifices we have to make! I closed this and created a new one that was separated from the Sonos changes.

Issue #78, opened 2015-07-23
Z-Wave thermostat


flynhawaiian:
Is there anyway to get a wave thermostat working?  I'm running homebridge on my pi2 running through smart things.  I'm wondering if it would be possible to have Siri set the temperature or turn on and off the thermostat.  I'm guessing a Js needs to be written for it but I have no idea. any thoughts would be cool!

flynhawaiian:
Sorry it was supposed to be zwave

nfarina:
Hi, I'm not familiar with SmartThings but @jnewland added support for it originally so perhaps he can chime in here.

flynhawaiian:
So the support for the thermostat is in the coding itself already?  Any idea which file it may be in so I can take a look at it.  I might be able to figure it out, but any help is really appreciated.

flynhawaiian:
Okay it looks like the smartthings.js needs to be changed to add some code regarding the thermostat.  Looking at the nest.JS it seems like the basic for setting the temp would be:

But theres a function that would also be needed from the looks of the coding.

{
                cType: types.CURRENT_TEMPERATURE_CTYPE,
                onUpdate: null,
                onRead: function(callback) {
                    that.getCurrentTemperature(function(currentTemperature){
                        callback(currentTemperature);
                    });
                },
                perms: ["pr","ev"],
                format: "int",
                initialValue: 20,
                supportEvents: false,
                supportBonjour: false,
                manfDescription: "Current Temperature",
                unit: "fahrenheit"
            },
perms: ["pr","ev"],
                format: "int",
                initialValue: 20,
                supportEvents: false,
                supportBonjour: false,
                manfDescription: "Current Temperature",
                unit: "fahrenheit"
            },{
                cType: types.TARGET_TEMPERATURE_CTYPE,
                onUpdate: function(value) {
                    that.setTargetTemperature(value);
                },
                onRead: function(callback) {
                    that.getTargetTemperature(function(targetTemperature){
                        callback(targetTemperature);
                    });
                }

tommyd75:
OK, so I'm trying to get a thermostat working on here, and I had it working before with the Thermostat_accessory.js  type of files on the HAP-NodeJS but I can't seem to get it working here. I've looked at the Nest platform file and I see how it's working but I'm trying to take out all the nest related stuff. The Nest file runs but them crashes because it tries to log into the website.
I removed some of the login info but that didn't help, I keep getting errors on other things, can anyone steer me in the right direction?

[...]

PR #81, opened 2015-07-24
Z-Wave Thermostat

flynhawaiian:
Is there anyway to get a wave thermostat working? I'm running homebridge on my pi2 running through smart things. I'm wondering if it would be possible to have Siri set the temperature or turn on and off the thermostat. I don't know if a JS would have to be created or it would be a simple thing.

I have all of the coding to go into the config file.  I pulled out all of the information for the various settings and can send it if necessary.  I am wondering if I would call the main heading thermostat versus switch or hue in the config file as well.


nfarina:
I think you may have created this PR accidentally - it's from a branch I was working on that isn't intended to be merged yet. I see you filed Issue #78, let's move this topic there.

PR #87, opened 2015-07-28
Added Docker Support

amitgandhinz:
This change adds a Dockerfile to create an ubuntu image running
nodejs and homebridge.

To run:

docker build -t homebridge .
docker run homebridge

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:14.04
+MAINTAINER Amit Gandhi <amit@gandhi.co.nz>
+
+RUN apt-get -qq update
+RUN apt-get -qq upgrade
+
+# Install npm
+RUN apt-get -qqy install npm curl libavahi-compat-libdnssd-dev git
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN curl -sL https://deb.nodesource.com/setup_0.12 | sudo bash -
+RUN apt-get install -qqy nodejs
+
+RUN git clone https://github.com/nfarina/homebridge.git /usr/src/app
+RUN npm install
+
+COPY config.json /usr/src/app/
+
+CMD [ "npm", "start" ]
+
+EXPOSE 51826
+


amitgandhinz:
Note - This is still a Work In Progress.

There is still work remaining on this PR to do appropriate Port mapping as it seems like home bridge runs multiple services, each on its own Port.  I need to figure out a way to dynamically expose each of those ports on the container to the host.

nfarina:
Very cool. What kind of machine are you running HomeBridge on?

amitgandhinz:
Thanks.  I'm running OS X Yosemite.  But I like to run things through Docker.  Also this helps alleviate some of the build issues I've seen people have in getting home bridge running properly.

nfarina:
I think this will be great once #94 is merged - that will change homebridge to expose just one predictable port and all Accessories will be hosted behind that.

amitgandhinz:
awesome.

nfarina:
#94 is now merged. Could you try your Docker file with the latest Homebridge version? It should create a single Bridge accessory on port 51826.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+RUN curl -sL https://deb.nodesource.com/setup_0.12 | sudo bash -
+RUN apt-get install -qqy nodejs
+
+RUN git clone https://github.com/nfarina/homebridge.git /usr/src/app
+RUN npm install
+
+COPY config.json /usr/src/app/
```
velocityware: this like should probably be:

COPY config-sample.json /usr/src/app/config.json

shtirlic:
Hi there, why not use official https://hub.docker.com/_/node/ images from node hub? Based on debian image

amitgandhinz:
I am abandoning this change.  I finally got around to looking at docker support and found a bunch of implementations.  The one I like best is https://github.com/ViViDboarder/docker-rpi-homebridge due to its simplicity in getting it set up and not requiring a bunch of different files to get it working.

PR #88, opened 2015-07-28
Initial attempt at a Yamaha AVR control module, like Sonos.

SphtKr:
Module to control Yamaha networked receivers, like my RX-475 and RX-673.

Not working quite yet, but close. Looking for feedback here, appreciate it! I added `mdns` as a dependency directly (probably not a problem since HAP-nodes uses it), but also unsure if I'm committing any faux-pas...I'm fairly new to Node.

I can probably figure out why it's not working (looks like it does, but nothing shows up in Eve), but would appreciate any hints there too! Thanks!

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -57,6 +57,11 @@
             "platform": "Sonos",
             "name": "Sonos",
             "play_volume": 25
+        },
+        {
+            "platform": "YamahaAVR",
+            "play_volume": -35,
+            "setMainInputTo": "AirPlay" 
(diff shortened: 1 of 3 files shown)


nfarina:
Very cool!

SphtKr:
Thanks...I've got it working now (I made a few incorrect assumptions about debugger behavior and promises :-/ ).

nfarina:
Awesome, thanks for the PR!

PR #93, opened 2015-07-31
added FHEM.js platform shim for the FHEM home automation system

justme-1968:
the shim supports switches, lightbulbs (FS20, homematic, EnOcean, HUE, WifiLight,... dimmers and hue/rgb colored lights), homematic window blinds, MAX and homematic thermostats, homematic door&window sensors, PRESENCE, SONOS players (including volume) and probably many FHEM devices.

the shim is bidirectional and uses longpoll to push fhem events back to homekit. fhem events are cached so homekit querys for known values will not cause a roundtrip to fhem. modifications for integer values like dim levels and blind levels for rf devices and volume levels for SONOS are delayed for 1 second so the value can be adjusted without generating a flood of events and only a final value is transmitted.

for SONOS volume support a OUTPUTVOLUME_CTYPE hast to be added to HAP-NodeJS/accessories/types.js: exports.OUTPUTVOLUME_CTYPE = stPre + "27" + stPost;
this is non standard (?) and probably a different identifier should be used.

a small debug web server on port 8080 is included

more (in german):
http://forum.fhem.de/index.php/topic,32652.msg314166.html#msg314166

### DIFF
--- a/platforms/FHEM.js
+++ b/platforms/FHEM.js
@@ -451,6 +451,20 @@ FHEMAccessory(log, connection, s) {
     log( s.Internals.NAME + ' is NOT a thermostat. set for target temperature missing' );
   }
 
+  var event_map = s.Attributes.eventMap;
+  if( event_map ) {
+    var parts = event_map.split( ' ' );
+    for( var p = 0; p < parts.length; p++ ) {
+      var map = parts[p].split( ':' );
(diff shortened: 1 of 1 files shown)


nfarina:
Wow, epic shim! Thanks for the PR!

PR #94, opened 2015-08-03
Use WIP version of HAP-NodeJS refactor

nfarina:
* [x] Support Platforms

Wanted to throw this out there for anyone that is able to do some testing. This branch uses the WIP version of a HAP-NodeJS refactor that I've been working on here: https://github.com/KhaosT/HAP-NodeJS/pull/110#issuecomment-127223072

You can test by checking out this branch, deleting `HAP-NodeJS` from your `node_modules` in homebridge, then doing another `npm install`.

### DIFF
--- a/accessories/WeMo.js
+++ b/accessories/WeMo.js
@@ -68,7 +68,7 @@ WeMoAccessory.prototype = {
         if (!err) {
             var binaryState = parseInt(result)
             that.log("power state for " + that.wemoName + " is: " + binaryState)
-            callback(binaryState)
+            callback(binaryState > 0 ? 1 : 0);
         }
         else {
             that.log(err)
--- a/app.js
+++ b/app.js
@@ -18,7 +18,7 @@ if (!fs.existsSync(configPath)) {
 }
 
 // Initialize persistent storage
-//storage.initSync();
+storage.initSync();
 
 // Start by creating our Bridge which will host all loaded Accessories
 var bridge = new Bridge('HomeBridge', uuid.generate("HomeBridge"));
@@ -27,18 +27,19 @@ var bridge = new Bridge('HomeBridge', uuid.generate("HomeBridge"));
 var config = JSON.parse(fs.readFileSync(configPath));
 
 // keep track of async calls we're waiting for callbacks on before we can start up
+// this is hacky but this is all going away once we build proper plugin support
 var asyncCalls = 0;
+var asyncWait = false;
 
 function startup() {
-//    if (config.platforms) loadPlatforms();
+    asyncWait = true;
+    if (config.platforms) loadPlatforms();
     if (config.accessories) loadAccessories();
+    asyncWait = false;
     
-    bridge.publish({
-      username: "CC:22:3D:E3:CE:27",
-      port: 51826,
-      pincode: "031-45-154",
-      category: Accessory.Categories.OTHER
-    });
+    // publish now unless we're waiting on anyone
+    if (asyncCalls == 0)
+      publish();
 }
 
 function loadAccessories() {
@@ -81,6 +82,7 @@ function loadAccessories() {
 function loadPlatforms() {
 
     console.log("Loading " + config.platforms.length + " platforms...");
+    
     for (var i=0; i<config.platforms.length; i++) {
 
         var platformConfig = config.platforms[i];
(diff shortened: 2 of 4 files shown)


SphtKr:
I'm catching this now and then...possibly when opening a HomeKit app (Eve) and checking accessory status. I have LiftMaster and YamahaAVR configured (could be my YamahaAVR module :-) ).

    events.js:85
          throw er; // Unhandled 'error' event
                ^
    Error: read ECONNRESET
        at exports._errnoException (util.js:746:11)
        at TCP.onread (net.js:559:26)

nfarina:
Hm, is that the complete stacktrace?

SphtKr:
Yeah, that was all that was in the log file that time, at least. I think there may have been more the first time but I dismissed it at the time. On the plus side, I haven't hit it again since yesterday. This has happened two or three times, but it's hit in the instance on my server so I don't have node-inspector hooked up to it or anything, sorry. I'll keep an eye out, could have been unrelated.

nfarina:
Just got this myself. Investigating.

KhaosT:
It seems that we're trying to read/write to a socket that has already been closed, this might be caused by iOS closed that side's socket and we didn't teardown the socket properly.

nfarina:
This just happened again after running for a couple days - I used "longjohn" as recommended [here](http://stackoverflow.com/questions/17245881/node-js-econnreset) to attempt to get a longer stacktrace; here's what I got:

```sh
/home/pi/homebridge/node_modules/longjohn/dist/longjohn.js:195
        throw e;
              ^
Error: read ECONNRESET
    at exports._errnoException (util.js:746:11)
    at onread (net.js:559:26)
---------------------------------------------
    at fireErrorCallbacks (net.js:458:15)
    at Socket._destroy (net.js:495:3)
    at onread (net.js:559:17)
```

nfarina:
Good news - I can repro this in less than 2 days ;) If you simply `telnet` to localhost on the HAP port, then kill the telnet process, about 50% of the time it will cause an `ECONNRESET`. The difference between HAP-NodeJS master and this refactor is that the master branch signs up for the `error` event from the socket, which causes the error to be "caught" instead of killing the node process entirely.

The fix is simply to handle the error event, which I'll apply a patch for in the refactor branch.

nfarina:
OK this branch should contain the fix for `ECONNRESET`.

SphtKr:
Awesome! I've got it pulled down and running on my Mini now. Will let you know if I hit further problems.

SphtKr:
A warning, the downside to the single bridge device implementation is that if you advertise a device that HomeKit decides is not compliant, it will kill the whole bridge and all attached devices! I'm working on a bridge platform where I can't test all the devices someone could possibly connect, so if my code advertises an illegal combination of characteristics, it will take all the user's devices "unreachable"! 

For example, I have some thermostats that don't report the current actual temperature...both characteristics are required for a HomeKit thermostat so I had to build in a specific workaround for that situation. But I don't know if someone else's device may produce a similar scenario (e.g., apparently a garage door opener ''must'' have an obstruction detector device to go with it).

On top of that, I've managed to put a device out of compliance in such a way that HomeKit does not want to talk to it anymore...somehow. That would be pretty bad for users, right now the only workaround is to change the "username" in app.js so it looks like a new device.

I opened #104, that would give a "break glass" kind of workaround for that wedged situation.

nfarina:
@SphtKr Very very excellent points. I've seen all these things happen to me while building this out. I still think the pairing process is painful enough that the bridge approach is worth it personally but I'd be curious to know others' thoughts.

SphtKr:
Right, I think this is a good way to go, just gotta keep in mind ways to make it work. E.g. that issue I opened, and also I'm going to try to make it easy for folks to exclude devices if needed. Tough when HomeKit is designed around only adding known devices that have been through a rigorous testing process :grimacing:.

nfarina:
Alright, let's do this!

maddox:
Oh snap!

cflurin:
@nfarina Thanks for the new homebridge version.
I've just uploaded a shim for Temperature Sensors here:

https://gist.github.com/cflurin

The temperature will be updated from the FHEM-Server (longpoll) similar to justme-1968's FHEM.js

nfarina:
@cflurin Awesome! Hopefully this will get even easier with the new API - see my [temperature sensor example in HAP-NodeJS](https://github.com/KhaosT/HAP-NodeJS/blob/master/accessories/TemperatureSensor_accessory.js)

PR #96, opened 2015-08-07
update FHEM platform to the current version

justme-1968:
lots of internal cleanups 
more supported device types
query of values to fhem is nearly gone, initial device state (and longpoll) is used instead
optimized caching


maddox:
* edited title to reflect what the PR was updating "to the current version"

PR #116, opened 2015-08-23
Support new API in getServices()

### LINKED ISSUE #57: Higher level API
I think we should consider adding some more abstraction on top of HAP-NodeJS for ease of development. I think [this project](https://github.com/Torsten85/node-homekit) could be a good starting point for design. Using chaining with getters, setters, and event emitters would greatly reduce the amount of code required to create an accessory. It also reduces the HomeKit specific knowledge required for a new developer to start contributing to the project. We might also want to consider some higher level accessory base classes such as Light, Switch, Lock, Door, Alarm, Sensor, etc. which provide defaults and create characteristics based on what parts of the interface the developer chooses to implement. 

Here is an example of a potential interface using EventEmitters with callbacks. The underlying API would create permissions, cTypes, and other HomeKit specific attributes for HAP-NodeJS based on the event listeners that are implemented.

```javascript
var accessory = new Accessory('First Garage Liftmaster');

var safetyLight = new Accessory.Light('Overhead Light');
accessory.addService(safetyLight);

safetyLight
    .on('setPowerOn', function (callback) {
        // Liftmaster API call for turning light on
    })
    .on('setPowerOff', function () {
        // Liftmaster API call for turning light off
    });

var backlight = new Accessory.Light('Information Display Backlight');
accessory.addService(backlight);

backlight
    .on('setPowerOn', function () {
        // Liftmaster API call for turning light on
    })
    .on('setPowerOff', function () {
        // Liftmaster API call for turning light off
    });

var securityLock = new Accessory.Lock(
    'Control Panel Lock',
    // We could add accessory type specific options such as Apple's auto lock feature like this
    {
        autoLock: true,
        lockTimeout: 10000
    }
);

accessory.addService(securityLock);

securityLock
    // Auto lock also calls this to simplify the interface
    .on('setLocked', function (isLocked, done) {
        // Fake Liftmaster API call for locking garage controls
        liftmasterApi.setControlLock(isLocked, function (err) {
            done(err);
        });
    })
    .on('getLocked', function (done) {
        // Fake Liftmaster API call for getting lock state of garage controls
        var isLocked = liftmasterApi.getControlLock(function (err, locked) {
            // Node js style callback with err, result
            done(null, isLocked);
        });
    });

// Example fake Liftmaster API event callback for updating a read-only characteristic
liftmasterApi
    .obstructionDetected(function () {
        securityLock.emit('obstructionDetected');
    });

// Run this when any discovery and setup is done to start the server
accessory.publish();
```

### LINKED ISSUE #114: New querying of types
With the enhanced HAP-NodeJS a new way for configuring accessories can be done. Shown in Lock_accessory.js and TemperatureSensor_accessory.js. But I can not figure out how to franslate this in a fake accessory for homebridge. Could you provide some guidance or also a fake example?

nfarina:
See [`getServices()` implementation in `accessories/Http.js`](https://github.com/nfarina/homebridge/blob/support-new-api/accessories/Http.js#L72) for an
example of how to use.

Fixes #114
Fixes #57

/cc @madmod, @volschin

### DIFF
--- a/accessories/Http.js
+++ b/accessories/Http.js
@@ -1,5 +1,10 @@
-var types = require("HAP-NodeJS/accessories/types.js");
+var Service = require("HAP-NodeJS").Service;
+var Characteristic = require("HAP-NodeJS").Characteristic;
 var request = require("request");
+
+module.exports = {
+  accessory: HttpAccessory
+}
 
 function HttpAccessory(log, config) {
   this.log = log;
@@ -9,9 +14,6 @@ function HttpAccessory(log, config) {
   this.off_url = config["off_url"];
   this.brightness_url = config["brightness_url"];
   this.http_method = config["http_method"];
-
-  // device info
-  this.name = config["name"];
 }
 
 HttpAccessory.prototype = {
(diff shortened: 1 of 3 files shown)


volschin:
Looks very nice. I will try it today.

cflurin:
My first shim using the new API > FhemSwitchN.js:
https://gist.github.com/cflurin

The value for callback(null, value) inside getPowerState has to be a string, I don't know the reason, it should be a boolean.

nfarina:
@cflurin Very cool! You shouldn't have to coerce values to string - did you try `true`/`false` instead of 0/1?

cflurin:
Yes, I did. Using true/false or 0/1 EVE delivers N/A
I also get 0/1 instead of false/true from EVE, Insteon+ sends false/true! 
````javascript
.on('get', this.getPowerState.bind(this));
````

**Comment by volschin** [comment]
Nick, is there a reason why do you inherit from a special service and not from Accessory? Normally an acessory willl have at least two services, Service.AccessoryInformation and a specific one. Or did I missed up something?

**Comment by nfarina** [comment]
@volschin Don't 100% follow - you mean why don't we inherit from Accessory class in HAP-NodeJS?

**Comment by volschin** [comment]
@nfarina In your HAP temperature example you require 
var Accessory = require('HAP-NodeJS').Accessory;
and then work with new Accessory() and addService(Service.TemperatureSensor)
in http.js you dont do so and return directly the service.
So in HAP I can combine multiple services in one accessory.

Can this be done also in Homebridge?

**Comment by nfarina** [comment]
Yes, I tried to simplify things in Homebridge's API - especially since we're already calling our objects "accessories" even though they don't inherit from HAP-NodeJS's Accessory class (and I'd like to not inherit from that). So this way you can just return an array of (possibly multiple) services in `getServices()`. And you can ignore `AccessoryInformation` service since it's added for you. Idea is to make things as easy as possible.

**Comment by volschin** [comment]
Means no chance to set serial numbers, firmware versions ... ?

**Comment by nfarina** [comment]
What if you were optionally able to return an `AccessoryInformation` service as well?

**Comment by volschin** [comment]
This sounds good.

**Comment by nfarina** [comment]
OK try that!

**Comment by nfarina** [comment]
@cflurin When I use true/false in the Http sample, it works in Eve, MyTouchHome, and HomeKit Catalog - all show "On" correctly. Modification you can try to Http.js in this PR:

```js
    lightbulbService
      .getCharacteristic(Characteristic.On)
      .on('get', function(callback) { callback(null, true); })
      .on('set', this.setPowerState.bind(this));
```

cflurin:
@nfarina, It was a stupid bug in getPowerState(), I've fixed it.
https://gist.github.com/cflurin

nfarina:
@cflurin Cool. Might want to check one of your callbacks… should be `callback(err)` not `callback(err, null)`

cflurin:
@nfarina, Thanks I'll fix it.
I'm trying the Service.AccessoryInformation, the Manufacturer, Model and SeriaNumber are OK
but I'm not sure how to implement the Identify Service:

```javascript
    var informationService = new Service.AccessoryInformation();
    
    informationService
      .setCharacteristic(Characteristic.Manufacturer, "FHEM Manufacturer")
      .setCharacteristic(Characteristic.Model, "FHEM Model")
      .setCharacteristic(Characteristic.SerialNumber, "FHEM Serial Number");
      
    informationService
      .getCharacteristic(Characteristic.Identify)
      .on('set', function(paired, callback) { 
        this.log("setIdentify: " + paired);
        callback();
      }.bind(this));
````
What is wrong?

nfarina:
@cflurin check the http sample on the PR - just implement `identify()`

cflurin:
Great! I think my example is now almost completed. https://gist.github.com/cflurin

cflurin:
The temperature is not sent to the iOS device:

```javascript
  getServices: function() {
  
    var informationService = new Service.AccessoryInformation();
    
    informationService
      .setCharacteristic(Characteristic.Manufacturer, "FHEM Manufacturer")
      .setCharacteristic(Characteristic.Model, "FHEM Model")
      .setCharacteristic(Characteristic.SerialNumber, "FHEM Serial Number")
      .setCharacteristic(Characteristic.Name, this.name);
      
    var FhemTemperatureSensorService = new Service.TemperatureSensor();
    
    this.currentCharacteristic = FhemTemperatureSensorService
      .getCharacteristic(Characteristic.CurrentTemperature)
      .on('get', function(callback) { 
        this.log('Temperature requested');
        callback(null, 22.2); 
        }.bind(this));

    return [informationService, FhemTemperatureSensorService];
  }
````
The Message 'Temperature requested' is OK but EVE shows 0.0

nfarina:
Are you running iOS 9? Temperature sensor is a new service type and iOS 8 may not be handling it correctly.

cflurin:
No, I'm running iOS 8.4.1 but with the Master Branch (old API) it works. There is an example on my Gist.

cflurin:
I updated the HAP-NodeJS today, so it may be the difference?

cflurin:
@nfarina 
The difference is in Accessory.js, removing "status=0" works:
````javascript
  characteristic.getValue(function(err, value) {
    
    debug('[%s] Got Characteristic "%s" value: %s', this.displayName, characteristic.displayName, value);

    if (err) {
      debug('[%s] Error getting value for Characteristic "%s": %s', this.displayName, characteristic.displayName, err.message);
      callback(err);
    }
    
    // all done - compose the response and send it to the callback
    callback(null, {
      aid: aid,
      iid: iid,
      value: value,
      // status: 0
    });
    
  }.bind(this), context);
````

nfarina:
@KhaosT what is status used for here?

cflurin:
Test results:  (new API) Lightbulb including dim, hue and setValue() working properly:
https://gist.github.com/cflurin

KhaosT:
Status is used to indicate if this read is success or not but I guess iOS 8.4 doesn't like we return status if HTTP status is 200…

cflurin:
In a month I'll update to iOS 9. There is no need to change anything, I've removed "status: 0" and it's OK for me.

nfarina:
OK merging this, let's move forward!

PR #124, opened 2015-08-24
Added usescenes parameter and fixed dimmer levelrange 1-100 => 1-16

EddyK69:


nfarina:
Note that you may need to fix this to avoid the "multiple callback" issue being attempted-fixed in #125. Also, cc'ing @joepv @lukeredpath @gizmocuz since they worked on the Domoticz platform (which I can't run, not having the hardware myself)

EddyK69:
Ok, i am closing this request and will make a new one after more testing of the #125 fix... thx!

PR #127, opened 2015-08-26
Updated MiLight Shim

dotsam:
Now uses new API and now actually supports lamp types other than rgbw in a meaningful way. Also added milight to readme.


nfarina:
Looks great!

PR #130, opened 2015-08-27
[Domoticz] Fixed misc. dimmer issues (SwitchType instead of HaveDimmer etc) 

EddyK69:
Fixed issue that 'on/off'-type lights showed as dimmers in HomeKit. Checking now on SwitchType instead of HaveDimmer
Fixed issue that 'on-off'-type lights would not react on Siri 'Switch on/off light'; On/Off types are now handled as Lights instead of Switches


snowdd1:
Just curious: Why don't you have `lightbulb` types without dimmer? 
I used lighbulb types for both with and without dimmers. I just didn't assign them the brightness characteristic. Siri still can turn those on and off.

EddyK69:
Thnx for the suggestion; i will look into it

PR #135, opened 2015-08-31
[MiLight] Update to address correct bulbs and implement all features

dotsam:
This should be feature-complete for all types of lamps now. Except the disco features. Can't think of a way to cleanly implement them, and who really wants them anyway?


nfarina:
Cool, thanks!

PR #147, opened 2015-09-06
Update config-sample-knx.json

Netsrac80:
forgotten comma


snowdd1:
Did I write there that it would help to paste it into jsonlint.com before saving? That would have been a great idea to do...

snowdd1:
already done, please close.

PR #152, opened 2015-09-07
[MiLight] Fix scope issue that prevented config from working

### LINKED ISSUE #143: MiLight accessory does not pass IP address
Thanks for your amazing work!

The IP address of the bridge does not flow through from the config.json file.  The zone is correctly passed, but the IP appears to be 255.255.255.255 (the constant DEFAULT_IP defined in node_milight_promise) this results in the lights associated with the zone on each of the MiLight gateways to turn on.  Example:  I have three MiLight gateways (192.168.69.51-53).  My config.json defines the accessory as follows:

{
"accessory":"MiLight",
"name": "Entryway",
"ip_address": "192.168.69.51",
"port": 8899,
"zone": 1,
"type": "rgbw",
"delay": 35,
"repear": 3
}

When this executes, all three gateways turn one RGBW zone 1.

I wish I could figure out the code myself, but it is beyond my ability.  Any help would be amazing.

dotsam:
Also added documentation and a pause command to help with timing. Fixes #143

### DIFF
--- a/accessories/MiLight.js
+++ b/accessories/MiLight.js
@@ -69,60 +69,62 @@ function MiLight(log, config) {
   this.delay = config["delay"];
   this.repeat = config["repeat"];
 
-  var light = new Milight({
+  this.light = new Milight({
     ip: this.ip_address,
     port: this.port,
     delayBetweenCommands: this.delay,
     commandRepeat: this.repeat
-});
+  });
 
 }
 MiLight.prototype = {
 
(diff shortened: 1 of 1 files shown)


dotsam:
Nope, something else broke here. Please don't merge this yet.

dotsam:
This is all fixed now, can be merged. (Won't it be great when we all manage our own npm modules?)

nfarina:
Agreed!

PR #155, opened 2015-09-09
[MiLight] Convert to platform

dotsam:
Convert accessory to platform and fix a bug due to a missing successful callback. All users of the accessory will have to convert their config to a platform.

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -71,7 +71,17 @@
             "platform": "YamahaAVR",
             "play_volume": -35,
             "setMainInputTo": "AirPlay" 
-        }
+        },
+        {
+            "platform": "MiLight",
+            "name": "MiLight",
+            "ip_address": "255.255.255.255",
+            "port": 8899,
+            "type": "rgbw",
+            "delay": 30,
+            "repeat": 3,
+            "zones":["Kitchen Lamp","Bedroom Lamp","Living Room Lamp","Hallway Lamp"]
+         }
     ],
 
     "accessories": [
@@ -151,16 +161,6 @@
             "host": "192.168.1.200", // IP address of the SER2SOCK service
             "port" : 4999, // Port the SER2SOCK process is running on
             "pin": "1234" // PIN used for arming / disarming
-        },
-        {
-            "accessory":"MiLight",
-            "name": "Lamp",
-            "ip_address": "255.255.255.255",
-            "port": 8899,
-            "zone": 1,
-            "type": "rgbw",
-            "delay": 35,
-            "repeat": 3
         },
         {
             "accessory": "Tesla",
--- a/platforms/MiLight.js
+++ b/platforms/MiLight.js
@@ -27,7 +27,7 @@ Where the parameters are:
  *ip_address (optional): The IP address of the WiFi Bridge. Default to the broadcast address of 255.255.255.255 if not specified
  *port (optional): Port of the WiFi bridge. Defaults to 8899 if not specified
  *type (optional): One of either "rgbw", "rgb", or "white", depending on the type of bulb being controlled. This applies to all zones. Defaults to rgbw.
- *delay (optional): Delay between commands sent over UDP. Default 30ms
+ *delay (optional): Delay between commands sent over UDP. Default 30ms. May cause delays when sending a lot of commands. Try decreasing to improve.
  *repeat (optional): Number of times to repeat the UDP command for better reliability. Default 3
  *zones (required): An array of the names of the zones, in order, 1-4. Use null if a zone is skipped. RGB lamps can only have a single zone.
 
@@ -202,9 +202,9 @@ MiLightAccessory.prototype = {
     } else if (this.type == "white") {
       // Again, white lamps don't support setting an absolue colour temp, so trying to do warmer/cooler step at a time based on colour
       if (value >= 180) {
-        this.light.sendCommands(commands.white.warmer());
-      } else {
         this.light.sendCommands(commands.white.cooler());
+      } else {
+        this.light.sendCommands(commands.white.warmer());
       }
     }
     callback();
--- a/platforms/MiLight.js
+++ b/platforms/MiLight.js
@@ -65,7 +65,6 @@ function MiLightPlatform(log, config) {
 
 MiLightPlatform.prototype = {
   accessories: function(callback) {
-    var that = this;
     var zones = [];
 
     // Various error checking    
@@ -208,7 +207,7 @@ MiLightAccessory.prototype = {
         this.light.sendCommands(commands.white.cooler());
       }
     }
-
+    callback();
   },
 
   identify: function(callback) {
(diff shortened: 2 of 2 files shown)


nzhome:
awesome!! will do.

nzhome:
you probably shouldn't have the word "lamp" or "Room"  in the zone name should you? this would be better for siri?

"zones":["Kitchen","Bedroom","Living","Hallway"]

nzhome:
QA Tested.. PASSED. new platform Works perfectly.  including multi-platforms.  I have multiple bridges and a mix of rgbw and white.

although I found a huge message queue lag with the defaults. I could gather up to 20-30 second lag between commands was easy to build up when changing hue and brightness, then trying to toggle the on/off.

this gave the perfect response with no message queue lag, please make this the default 
            "delay": 1,
            "repeat": 1,

(I tried using 0's but it made them worse)

Also its quite hard to get the slider to 2% for nightmode, can you make it anything <= 10% for whites to be night mode.  thanks.  your idea of <50 drive whites down, and >50 drive whites up.. legendary.  it works soo well. because as I slide more it increases brightness anyway. ohh buy it did feel a little deadzone'ish near  50 and weird as you come back down from 100 though...  I have another better idea.  if you are at <100 and decreasing since last brightness command then decrease.   (but make brightness steps of 10% only drive  commands down)    then on the way up.. increase.. again only increase when the brightness increases 10% steps from the last brightness command.  that should do it.

nfarina:
Ok to merge?

dotsam:
Let's go ahead and merge this in now @nfarina  

@nzhome I've done the quick swap for warmer/cooler on white bulbs, but I'm going to have to think on how I'd like to implement your other suggestions. As for the delay/repeat defaults, these are coming from the npm module I'm using, so I'd rather not have to worry about two sets of defaults. After reading some details on the RF protocol the lights are using (https://hackaday.io/project/5888-reverse-engineering-the-milight-on-air-protocol) I'm actually a little surprised things are working that reliably for you with no delay. There must be something else (maybe Homebridge) introducing enough delay to make things work.

nzhome:
BEST:
            "delay": 2,
            "repeat": 1,

GOOD
            "delay": 1,
            "repeat": 1,

BAD
            "delay": 2,
            "repeat": 2,

WORST
            "delay": 2,
            "repeat": 3,

nzhome:
as soon as you put your finger on the MyTouchHome slider, the the brightness starts at 12%  and on the other end if the brightness is at 100% and you touch it it jumps to 88%   so it would be better to leave  1-11 and  89-99  as deadzones. and put the nightmode and brightness increments in between 12 and 88.

nzhome:
TESTING FAILED There is a major problems going on here...  although I am seeing perfect responses on homebridge itself... but the commands are getting jumbled and garbled by the dependent nodejs called node-milight-promise...  is there another milight library to use instead? this one is rubbish.


  Accessory [Homebridge] Setting Characteristic "On" to value false +2ms
[MiLight] [Dining] Setting power state to off
  Accessory [Homebridge] Setting Characteristic "On" to value false +4ms
[MiLight] [Rumpus Wall] Setting power state to off
  Accessory [Homebridge] Setting Characteristic "On" to value false +7ms
[MiLight] [Rumpus Ceiling] Setting power state to off
  Accessory [Homebridge] Setting Characteristic "On" to value false +1ms
[MiLight] [Lounge] Setting power state to off
  Accessory [Homebridge] Setting Characteristic "On" to value false +1ms
[MiLight] [Kitchen] Setting power state to off
  EventedHTTPServer [fe80::1c6e:9398:f4d6:5c32] HTTP Repsonse is finished +3ms

dotsam:
I haven't had any problems with the library over the past 2+ months, and even had it updated to include all of the correct commands for all bulb types. Can you run Homebridge with the command "MILIGHT_DEBUG=1 node app.js" and provide the output of that along with a better description of what does/doesn't happen?

nzhome:
My friends who have used the same library also comment the same. they said the problem is more noticable on Raspberry PI 2 than raspberry PI 1. thanks I will do some debugging and further testing to see where the issue lies.  the LimitlessLED app and other Music Apps are working perfectly. so I know my bridge is all good.

nzhome:
umm you do know that the post above was already with debugging ON right?  thats why the [Milight] lines in the log I posted.

nzhome:
only half the lights in the house actually turned off in that case.. it could be cause you don't have white lights, or you don't have as many lights as us?

nzhome:
what should we try next?

dotsam:
The MILIGHT_DEBUG=1 line enables the debugging output from the node-milight-promise module, which should show the exact commands it's sending. Also, if you're only getting some light turning off, I would think that this might be a timing issue, so you should try increasing the delay value. I believe the "official" milight spec is 100ms between commands, so the 30ms timing of the node-milight-promise module is already aggressive.

nzhome:
if it is set to 100ms in my config.jason.  then why does it always try to do my lights so fast (see below +2ms?  and sometimes the command is missed. but I know that I can use a music app at a beats of 20ms with no problen.  why is 30ms agressive? I don't think so..  I think there is still some problem here..

  Accessory [Homebridge] Setting Characteristic "Brightness" to value 10 +0ms
[MiLight] [Rumpus Ceiling] Setting brightness to 10
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 10 +2ms
[MiLight] [Dining] Setting brightness to 10
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 10 +2ms
[MiLight] [Kitchen] Setting brightness to 10
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 10 +8ms
[MiLight] [Lounge] Setting brightness to 10
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 10 +1ms
[MiLight] [Rumpus Wall] Setting brightness to 10

dotsam:
This is the "-promise" part of the milight library. Since the MiLight protocol is UDP without any sort of response, there's not really any point in waiting for a callback to see if the command was successful or not, because we can't know that. So instead, the library returns immediately and keeps its own queue of commands to be sent out at the intervals defined in the config.

So even though Homebridge is sending out commands at 0-8ms (in your example above), the node-milight-promise library isn't going to send commands any less than 30ms (or whatever the configured number) apart. With the repeat parameter set to more than 1, it will be multiples of this number.

Again, this delay is intentional to be compatible with the MiLight protocol. As detailed in this reverse engineering of the protocol (https://hackaday.io/project/5888/logs), the bridge is sending out the RF commands ~40 times across 3 different frequencies. Presumably the bulbs are always hopping between these frequencies, and they also presumably implement a way of ignoring commands that aren't for them and other RF interference. The 100ms value if therefore a worst-case scenario for the bulb to be on the right frequency at the right time and not be ignoring commands.

So yes, 30ms is aggressive timing, as a bulb may not always be ready to receive a command that quickly in some environments. But it does seem to work most of the time. But as you start lowering that delay, you're going to see the problems of not all bulbs responding to all commands.

PR #166, opened 2015-09-14
Home Assistant Shim

maddox:
This is a first pass at a shim for [Home Assistant](https://home-assistant.io).

Right now this only supports lights, but I'd like it to at least manage switches too before it lands.

//cc @nfarina @balloob

### DIFF
--- a/platforms/HomeAssistant.js
+++ b/platforms/HomeAssistant.js
@@ -98,15 +98,22 @@ HomeAssistantPlatform.prototype = {
     var that = this;
     var foundAccessories = [];
     var lightsRE = /^light\./i
+    var switchRE = /^switch\./i
 
 
     this._request('GET', '/states', {}, function(error, response, data){
 
       for (var i = 0; i < data.length; i++) {
         entity = data[i]
+        var accessory = null
 
         if (entity.entity_id.match(lightsRE)) {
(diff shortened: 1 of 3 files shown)


maddox:
Before this lands, I'd like to have it support switches too. But I'm not exactly sure how to support 2 types of accessories inside here. I'd love some suggestions from anyone that might have done this in another shim. I couldn't really tell if any other shim has done that :/

platforms/HomeAssistant.js
```diff
@@ -0,0 +1,301 @@
...
+    })
+
+  }
+}
+
+function HomeAssistantAccessory(log, data, client) {
+  // device info
+  this.data = data
+  this.entity_id = data.entity_id
+  this.name = data.attributes.friendly_name
```
balloob: Not every entity will have this attribute. In [the HA frontend](https://github.com/balloob/home-assistant-js/blob/master/src/modules/entity/model.js#L40) I fall back to the object_id (part right of the `.` in entity_id) with `_` replaced by `<space>`

maddox: OK awesome, good to know. thanks.

maddox:
Woot woot, this is ready to go!

maddox:
Switch got added.

PR #171, opened 2015-09-15
Add Background Launching

maddox:
:construction: :construction: :construction::construction: :construction: :construction:  Pushing this up early for some discussion :construction: :construction: :construction::construction: :construction: :construction:

This is the start of allowing homebridge to launch at boot. This first push supports only OS X by way of `launchd` and [forever](https://github.com/foreverjs/forever). It basically uses `launchd` to launch `forever` to keep it alive and push out logs.

The app will now log to `$approot/log/logs.log` and `$approot/log/errors.log`.

In addition to all of this, I've added [Scripts to Rule Them All](http://githubengineering.com/scripts-to-rule-them-all/), a really nice pattern for maintaining and running things in any language.

### Scripts

The new scripts included allow us the bootstrap the app, upgrade the app, boot the app, install the app, uninstall the app, and restart the app. Here's some brief descriptions:

* `script/bootstrap` - Guards to make sure `forever` is installed, installs node packages, and creates the `config.json` file if it doesn't already exist.
* `script/server` - This boots homebridge using `forever`. If the environment is `development` the app loads in place and logs to `stdout`. You can simply If the environment is `production`, it daemonizes and is monitored and kept alive by  `forever`.
* `script/install` - This generates (based on the template `org.homebridge.plist`) a `launchd` script to load into your system to have it start right away and on boot. The script simply launches `script/server` with the `production` environment, causing it to daemonize.
* `script/uninstall` - Stops the app with `forever`, stops the process in `launchd`, and removes the `launchd` script from the installed place.
* `script/restart` - runs `script/uninstall` and then `script/uninstall`.
* `script/upgrade` - does a `git pull` on the current branch, an `npm update`, and then runs `script/restart`.

### Arch Support

As this pull is being made, only OS X (darwin) is supported for installing. Though I am going to have a linux version of all of this submitted before this should land.

But as of now, this works on OS X.

/cc @nfarina @lukeredpath


maddox:
I've added the linux support now. I'm still waiting for the npm packages to install on my pi to verify that it works though :grimacing:

maddox:
Well so far this is a disaster. The pi has been trying to install foreverrrrrr. And installing this on an Ubuntu VM is erroring like crazy. Damn npm.

Pi on the left and VM on the right, lol. So yeah, I could use help testing this probably to make sure it works on linux.

screen shot 2015-09-15 at 12 15 49 am

dborup:
Hi Maddox, I downloadede the script to rule them all package. 
As I am a n00b with this NPM/github stuff, I am not sure how I run the scripts ?

realdannys:
How do I actually roll this out into my Homebridge install? I feel stupid, but I don't see an obvious way documented to incorporate this pull...

snizzleorg:
I'm using the normal linux way which also described here: https://github.com/nfarina/homebridge/wiki/Running-HomeBridge-on-a-Raspberry-Pi

This is running very nice so far.

PR #173, opened 2015-09-15
File-based motion or contact sensor

SphtKr:
This module creates a motion sensor accessory based on watching a directory or file. This is useful because many inexpensive IP cameras (Foscam, D-Link, others) have an option to FTP an image or video file when they detect motion. If you use this accessory to monitor that FTP drop location, you can expose that camera as a motion sensor to HomeKit.

Config:

* `path`: the dir, file, or glob to watch (see docs for the chokidar module for what's possible)
* `window_seconds`: A file modification date has to be within this many seconds of now to trigger the sensor. This is also the length of time that the sensor will stay triggered before falling back to it's default state.
* `inverse`: if needed, you can invert the behavior of the sensor, for instance so that finding a file means there is ''no'' motion or contact detected.
* `sensor_type`: currently either `"m"` for motion sensor or `"c"` for contact sensor. Contact sensor support is a little wonky right now, it doesn't seem to want to fall back to the untriggered state for some reason--also app support seems to be lacking.

### DIFF
--- a/accessories/FileSensor.js
+++ b/accessories/FileSensor.js
@@ -1,0 +1,76 @@
+var Service = require("HAP-NodeJS").Service;
+var Characteristic = require("HAP-NodeJS").Characteristic;
+var chokidar = require("chokidar");
+var debug = require("debug")("FileSensorAccessory");
+var crypto = require("crypto");
+
+module.exports = {
+  accessory: FileSensorAccessory
+}
+
+function FileSensorAccessory(log, config) {
+  this.log = log;
+
+  // url info
+  this.name = config["name"];
+  this.path = config["path"];
+  this.window_seconds = config["window_seconds"] || 5;
+  this.sensor_type = config["sensor_type"] || "m";
+  this.inverse = config["inverse"] || false;
+  
+  if(config["sn"]){
+      this.sn = config["sn"];
+  } else {
+      var shasum = crypto.createHash('sha1');
+      shasum.update(this.path);
+      this.sn = shasum.digest('base64');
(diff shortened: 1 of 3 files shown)


nfarina:
Very cool!

tommyd75:
@SphtKr Do you think I would I be able to use this with a simple text file?

Like this for the path

/192.168.1.100/motion.txt

SphtKr:
@tommyd75 Absolutely, any file local on the file system. Two notes: if you're monitoring a single file, it needs to exist before the monitor starts, it seemed from my testing at least. And, you may wish to tweak the window_seconds parameter from the default of five...I'm finding the default should probably be a bit longer.

tommyd75:
@SphtKr So would the text file just have anything in there and that gets passed to the value, or does the camera have different things in there, would you be able to post what your local file looks like?

SphtKr:
@tommyd75 It's only looking at the file metadata, namely modification time (and, well, existence). So it doesn't matter what's in the file, only that the file's `mtime` falls within `window_seconds`. In my case, it's a JPEG file that's dropped in the watched directory by an IP camera. There'd be no data to pass through anyway, it's just a boolean true/false value. Make sense? I should do a wiki page...

tommyd75:
ahh ok, so as long as there's something written to the file it would note the time changed?

SphtKr:
@tommyd75 Exactly. Or just use something like `touch` on the terminal/shell...the contents don't actually have to change.

tommyd75:
@SphtKr ok, thanks!

PR #182, opened 2015-09-17
Add shim for devices with serial connection

Cosmo:
There are still a lot of devices which are not connected to a network, but can be controlled with a serial connection -- a video projector for example.

I've created a generic accessory shim those devices.
All you need is a serial port or usb-to-serial adapter.

The config values are pretty straight forward.
You set the name, id (a unique string), baud rate, manufacturer, model_name, the two commands for on and off for your device.

Set a Siri-name (e.g. Home Cinema) in your HomeKit app (Eve, Insteon+, ..) and you are good to go.

For example:
```
{
  "accessory": "GenericRS232Device",
  "name": "Projector",
  "id": "TYDYMU044UVNP",
  "baudrate": 9600,
  "device": "/dev/tty.usbserial",
  "manufacturer": "Acer",
  "model_name": "H6510BD",
  "on_command": "* 0 IR 001\r",
  "off_command": "* 0 IR 002\r"
}
```

### DIFF
--- a/accessories/GenericRS232Device.js
+++ b/accessories/GenericRS232Device.js
@@ -1,4 +1,5 @@
-var types = require("HAP-NodeJS/accessories/types.js");
+var Service = require("HAP-NodeJS").Service;
+var Characteristic = require("HAP-NodeJS").Characteristic;
 var SerialPort = require("serialport").SerialPort;
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


maddox:
Oh this is great!

Cosmo:
Sure thing!

accessories/GenericRS232Device.js
```diff
@@ -0,0 +1,58 @@
...
+          } else {
+            callback()
+          }
+        });
+      }
+    });
+  },
+  
+  getServices: function() {
+    var switchService = new Service.Switch();
```
nfarina: I think this Service should have a name (probably the same name as the accessory) to give Siri a better chance of working. Example:

```js
var switchService = new Service.Switch(this.name);
```

Cosmo: Oh, didn't know this was possible :+1: Thanks!

nfarina:
Awesome thanks!

PR #184, opened 2015-09-17
Print scannable setup code to terminal on startup

Cosmo:
This prints the pin code from the config to the terminal in a scannable format.
I was missing this from the first time I've installed homebridge ;)

Here's a example video of what it does:
[Setup Code](http://www.youtube.com/watch?v=JYHM4FtCSbw)

### DIFF
--- a/app.js
+++ b/app.js
@@ -191,6 +191,16 @@ function createAccessory(accessoryInstance, displayName) {
   }
 }
 
+// Returns the setup code in a scannable format.
+function printPin(pin) {
+  console.log("Scan this code with your HomeKit App on your iOS device:");
+  console.log("\x1b[30;47m%s\x1b[0m", "                       ");
+  console.log("\x1b[30;47m%s\x1b[0m", "    ┌────────────┐     ");
+  console.log("\x1b[30;47m%s\x1b[0m", "    │ " + pin + " │     ");
+  console.log("\x1b[30;47m%s\x1b[0m", "    └────────────┘     ");
+  console.log("\x1b[30;47m%s\x1b[0m", "                       ");
+}
+
 // Returns a logging function that prepends messages with the given name in [brackets].
 function createLog(name) {
   return function(message) {
@@ -210,3 +220,5 @@ function publish() {
 }
 
 startup();
+
+printPin(bridgeConfig.pin);
--- a/app.js
+++ b/app.js
@@ -211,6 +211,7 @@ function createLog(name) {
 }
 
 function publish() {
+  printPin(bridgeConfig.pin);
   bridge.publish({
     username: bridgeConfig.username || "CC:22:3D:E3:CE:30",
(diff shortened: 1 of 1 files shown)


nfarina:
OMG BRILLIANT

nfarina:
I wonder if the call to `printPin` would be better placed in the `publish()` method just before publishing the bridge - that way it won't scroll out of view while all your accessories/platforms are (verbosely) loading.

justme-1968:
great.

yes. it would be better at behind publish. there can be many many lines of output if a platform shim is used.

maddox:
THIS IS WONDERFUL!! Thanks!

snowdd1:
:+1:

> Am 17.09.2015 um 17:19 schrieb Jon Maddox - notifications@github.com <snowdd1.theelk.6452e6e747.notifications#github.com@ob.0sg.net>:
> 
> THIS IS WONDERFUL!! Thanks!
>

Cosmo:
Done!

KhaosT:
Awesome 👏

nfarina:
So, so cool. Thanks for this!

cflurin:
@Cosmo: not only cool but also useful. Thanks.

For OS X I had to set the color this way:
```javascript
  var color = "\x1b[30\x1b[47\x1b[0m"; // "\x1b[30;47m%s\x1b[0m"
  console.log("Scan this code with your HomeKit App on your iOS device:");
  console.log(color, "                       ");
  console.log(color, "    ┌────────────┐     ");
  console.log(color, "    │ " + pin + " │     ");
  console.log(color, "    └────────────┘     ");
  console.log(color, "                       ");
```

Cosmo:
@cflurin
Now I feel stupid for not using a variable for the escape sequences :grinning:

No color is needed? How does your terminal look like?
I couldn't get my iPhone to scan the setup code on a black background with green text.
Tried different angles, font sizes, no luck :(

cflurin:
@Cosmo: I got a gray background

terminal

I'm using the Basic-Profile for the Terminal, which has white-background.

Cosmo:
That explains why you don't need color :)

cflurin:
I didn't try but maybe "Light White" could be better:
```javascript
var color = "\x1b[107;30m%s\x1b[0m";
```

SphtKr:
Awesome, I thought of something like this but figured you'd have to have the little logo, which would run afoul of trademarks...didn't even think of trying it without!

Cosmo:
hey @cflurin, your white looks much better! is this a standard?

cflurin:
@Cosmo: take a look at: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
I don't really unterstand the difference between normal and light white! White should be RGB FF,FF,FF.
Anyway on my RPi the code 107 instead of 47 is actually "white".

PR #186, opened 2015-09-18
Refactor HA Switch to Support Being Stateless

maddox:
We do this, because the `HomeAssistant` shim supports Home Assistant scenes to be bridged over. 

Scenes don't REALLY have `on`/`off` state, at least not a reliable kind. So having a single push button would be best for firing off scenes. This means we just have an event that is fired.

### DIFF
--- a/platforms/HomeAssistant.js
+++ b/platforms/HomeAssistant.js
@@ -493,34 +493,31 @@ HomeAssistantSwitch.prototype = {
     }
   },
   getServices: function() {
-    var switchService = new Service.Switch();
     var informationService = new Service.AccessoryInformation();
+    var switchService;
     var model;
 
-    switch (this.domain) {
-      case "scene":
-        model = "Scene"
-        break;
-      default:
-        model = "Switch"
+    if (this.domain == 'switch') {
+      model = "Switch"
+      switchService = new Service.Switch();
+
+      switchService
+        .getCharacteristic(Characteristic.On)
+        .on('get', this.getPowerState.bind(this))
+        .on('set', this.setPowerState.bind(this));
+    }else{
+      model = "Scene"
+      switchService = new Service.StatelessProgrammableSwitch();
+
+      switchService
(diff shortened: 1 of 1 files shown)


maddox:
It looks like we can't directly use the stateless switch in a functional manner besides triggering native HomeKit events.

Since the `HomeAssistant` shim is meant only for using Siri as a front end, this doesn't work for it. 

Closing.

snowdd1:
@maddox 
The event riggers I learnt are just sensors, i.e. you can have a HK compliant pushbutton that updates the HK (i.e. fire the event) and then the event listeners in HK can react to that.
The thing you need is an actuator (here software in kind) the can be triggered from inside HK and actually DOES something (start the HA scene in your case).

I think it's best to use the switch (as you did before) and reset it's state to off after the scene has been started.

That is why I first got confused and needed https://github.com/KhaosT/HAP-NodeJS/issues/134#issuecomment-141184460 for clarification myself.

justme-1968:
are the ha scenes related and linked together? then you could switch all other scenes off if one scene is activated. if not then i think the best is to automatically switch the homekit state back to off using the switch in a push button kind of way. that schould be possible in the newValue parameter of the set callback.


for the fhem harmony activity integration i 'm using switches that are linked so that only one of them can be active. if you activate one activity all others go to off.

currently i'm using one accessory per activity but i will change this to one accessory with multiple switch services.

but anyway this is only a workaround. a characteristic with a 1 of many selection would be much better suited...

maddox:
@justme-1968 @snowdd1 great ideas. To be honest, scenes in HA actually DO have state. They ARE set to on/off. There's just a sort of bug with Hue specific lights due to some floating point issues on the Hue Bridge side. 

So technically, not having state is really a bug, in a single type of system. It's looking to get resolved as well. Since HA does maintain state for the Scene, I should probably just implement the getter so state can be updated. 

The hue bug will propogate over, but at least everything will work like it should once the bug is fixed.

PR #188, opened 2015-09-18
KNX cleanup

snowdd1:
I found that the description was wrongly named and in the wrong folder.
It is now named after the platform shim to appear next to it in
listings, also the config-sample that comes along with it.

As long as we do not go for own repositories and npm packages, should we
consider folders for additional files if required?


nfarina:
For now folders is OK - but I definitely want to move to separate npm packages soon.

nfarina:
OK to merge?

snowdd1:
Ready to merge!

PR #189, opened 2015-09-19
KNX Platform: "Garage Door Opener" and "Motion Sensor" services

snowdd1:
Two new services for the KNX platform.


snowdd1:
Thanks!

PR #191, opened 2015-09-20
fixing thermostat.

luxus:
switch the order of the characteristics
now my stellaZ thermostat is shown correct in apps like eve and insteon+

### DIFF
--- a/platforms/ZWayServer.js
+++ b/platforms/ZWayServer.js
@@ -89,11 +89,11 @@ ZWayServerPlatform.prototype = {
         //TODO: Unify this with getVDevServices, so there's only one place with mapping between service and vDev type.
         //Note: Order matters!
         var primaryDeviceClasses = [
-            "switchBinary",
             "thermostat",
-            "sensorBinary.Door/Window",
             "sensorMultilevel.Temperature",
-            "switchMultilevel"
+            "switchMultilevel",
+            "switchBinary",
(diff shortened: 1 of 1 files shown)


SphtKr:
@luxus Heh...well, that was almost what I was going to try...can you try this order for me though, instead:

            "thermostat",
            "switchBinary",
            "sensorBinary.Door/Window",
            "sensorMultilevel.Temperature",
            "switchMultilevel"

(and the same in `getVDevServices`)? The issue is that I have both a `switchBinary` device and a `sensorBinary.Door/Window` device that have additional `sensorMultilevel.Temperature` sensors as extras...and with the order you've used the temperature sensors would become the primary.

Odds are we'll find exceptional devices that make a single order for the whole Z-Wave ecosystem impossible to choose. I already have a solution in mind and partially implemented for the unusual cases...

In the existing code, you can add a Tag in ZWay's home automation interface called `Homebridge:Skip`, which would let you exclude your `switchBinary` device entirely, which would make it choose the `thermostat` as the primary. I also want to add a few more like `Homebridge:IsPrimary` and/or `Homebridge:IsNotPrimary` to give the Homebridge module hints as to how it should treat the devices. Also a tag to force Homebridge to split the Z-Wave device into multiple accessories would probably be useful, but I haven't figure out how that should work yet.

PR #204, opened 2015-09-25
Noted updated for requirement to successfully get HAP-NodeJS installed.

derekforeman:
HAP-NodeJS was failing to build with npm install. This might help clarify for others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,13 @@ You'll also need some patience, as Siri can be very strict about sentence struct
 
 # Getting Started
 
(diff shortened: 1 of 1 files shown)


nfarina:
Awesome thanks!

cflurin:
It depends of the OS. It isn't a requirement for Mac OS X.

maddox:
Yeah, this message is pretty confusing in the README all on it's own. It has no context. "Package" has is pretty vague these days. 

It should mention that this is a Linux dependency for those using Linux (probably a pi), and should provide installation instructions: `sudo apt-get install libavahi-compat-libdnssd-dev `

nfarina:
Good points. I hit Merge a bit too quickly there. Updated the readme in master

Issue #193, opened 2015-09-21
Z-Way improvements (thermostat etc.)


luxus:
follow up of this https://github.com/SphtKr/homebridge/commit/656a8057acbd177995f5a3dd9b63262082387d2b
and
https://github.com/nfarina/homebridge/pull/191
and 
json of stellaZ radiator valve thermostat
https://gist.github.com/luxus/3232fbedc09e8f1eff09#file-devices-json-L82

@SphtKr 
the:
```
        "thermostat",
        "switchBinary",
        "sensorBinary.Door/Window",
        "sensorMultilevel.Temperature",
        "switchMultilevel"
```
works as well
![img_0356](https://cloud.githubusercontent.com/assets/7449/9990870/2aca46fe-6065-11e5-9ece-48182d5fe479.jpg)
now it shows even the battery level :dancer: 
eve understands all characteristics

there is still one metric i don't see.
```
      {
        "creatorId": 1,
        "deviceType": "switchMultilevel",
        "h": -1703465104,
        "hasHistory": true,
        "id": "ZWayVDev_zway_11-0-38",
        "location": 1,
        "metrics": {
          "icon": "multilevel",
          "title": "Wohnzimmer Ventil",
          "level": 44
        },
```

 if i am in heat mode i can't even change it. but it would nice if could see it because it shows how open the valve is.

i don't have any additional z-wave items(sending back my lc-13 thermostat from danfoss). so thats the only one i can test!

SphtKr:
Aha! Well I think you have the Eve app update that came out today to thank for the battery status...finally! I've been trying to get that to work for weeks. I guess it wasn't me!

What...on earth does that `switchMultilevel` do? Is that fan speed? My German is very rough.

luxus:
hi @SphtKr,
sorry for the late answer.
i see you did a lot work on the z-way integration. the wiki sounds awesome.. 
i wish there would be such a integration in home-assistant.io (i learn to program at the moment to do just this)
 back to topic.
the `switchMultilevel` change the level of the valve that controls how much heated water go trough
![german heater]
(http://www.heizsparer.de/wp-content/uploads/images/heizkoerper-glieder-irina-fischer-fotolia.jpg)
for the stellaz it just works in the right operation mode

```
Operating the device
The device does not allow any local operation on the device itself but control via Z-Wave radio from a Z- wave controller only. The thermostat has four operating modes. They are switched wirelessly as well:
[...]

PR #223, opened 2015-10-02
RGB(+W+W) support

SphtKr:
Initial read/write support for RGB bulbs complete . Needs testing, but seems to work with my Aeon bulb…taking into account the wonkiness of that bulb with Z-Way, at least.

References #193 .


SphtKr:
@nfarina I think this is ready to merge, any feedback?

nfarina:
Just waiting for the word :)

PR #225, opened 2015-10-02
Update of handling of binary and percentage type KNX values

snowdd1:
The use of 0/1 for binary objects depends on installation, a contact
sensor can have contact=1 or open=1
The same applies to percentages:  a window can be 100% closed or 100%
open.

For a more general usage of reversed values, a suffix ("R") can now be
attached to all DPT1 and DPT5 group addresses.
Please see updated KNX.md for examples and details.

Note: outdated characteristic-workaround throw an exception at boot-up:
-  ContactSensorStateContact1
-  LockCurrentStateSecured0
-  LockTargetStateSecured0
Please see error message and KNX.md for reference.

### DIFF
--- a/accessories/knxdevice.js
+++ b/accessories/knxdevice.js
@@ -14,6 +14,9 @@ New 2015-09-18:
 New 2015-09-19:
 -  GarageDoorOpener Service
 -  MotionSensor Service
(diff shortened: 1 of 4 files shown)


snowdd1:
Thanks!

Issue #203, opened 2015-09-24
accessory.name displayName and uuid generation


justme-1968:
in the old platform support issue #95 there was a short discussion about the generation of the uuid for each accessory. unfortunately we have not followed this to the end.

for devices that have a real unique serial number it would be better to create the uuid based on this serial number an not on the (display) name which might change during the lifetime of the device.

i would like to suggest creating the (platform) accessories with something like this:
```
var accessoryUUID = uuid.generate(accessoryInstance.constructor.name + ":" + (accessoryInstance.uuid_base || displayName));
```

instead of a new variable uuid_base we could also use the serialNumber from the AccessoryInformation service.

snowdd1:
You mean here
https://github.com/nfarina/homebridge/blob/master/app.js#L162
?
I would support that. During modeling my house I had to change names quite often, as in a global (or house-wide) context the name was not distinctive enough, loosing room asignments and such.
I assume with an constant UUID that would not have happened.

Did you test your change in you own fork/locally?

AccessoryInformation service is quite global in my platform, I would rather configure them as a new property with the accessories within.

justme-1968:
yes. in app.js.

i will check if it works as intended.

desvdp:
see also #148

justme-1968:
what exactly is accessoryInstance.constructor.name in the call to uuid.generate supposed to do?
as far as i can see it just prefixes everything with an 'Object' string. so it looks redundant.

so this:
```
var accessoryUUID = uuid.generate(accessoryInstance.uuid_base || displayName);
```
should be enough?

desvdp:
Actually it prefixes with the name of the accessory type you're generating a UUID for, making accessories unique by their accessory shim by getting the name of the constructor object

> function MyCoolAccessory(n) { this.name=n;}
> var myThing = new MyCoolAccessory("home");
> console.log(myThing.constructor.name);

--> MyCoolAccessory

justme-1968:
hmmm... i'm not sure if this works as intended. if i put a debug message into uuid.generate it just prefixes everything with 'Object'.

snowdd1:
Are you using object literals?

justme-1968:
what exactly do you mean ?

nfarina:
[...]

PR #233, opened 2015-10-06
New UUID generation

snowdd1:
addresses #203

# This is a breaking change for all configured homekit databases, as it will remove room/area assignments and other things configured in iOS apps
... unless your accessory type was called `Object` before

Changes:
- Without any changes to the configuration file all accessories UUIDs will be generated by  `uuid.generate(accessoryType + ":"` instead of the constructor name which turned out to be unreliable in that context.  So about all UUIDs will change.
- For platforms the platformType will be used, as configured in config.json at eg. `"platform":"Nest"`
- new parameter uuid_base can be used for accessories in config.json to use a different base for UUID generation than the displayName, which might not be unique
- platforms can add the same property to their accessories before returned, just as they have a name property right now.

### DIFF
--- a/accessories/knxdevice.js
+++ b/accessories/knxdevice.js
@@ -17,6 +17,8 @@ New 2015-09-19:
 New 2015-10-02:
 - Check for valid group addresses
 - new "R" flag allowed for Boolean addresses: 1/2/3R is the boolean not(1/2/3), i.e. 0 and 1 switched on read and write
+New 2015-10-07:
+- Accept uuid_base parameter from config.json to use as unique identifier in UUIDs instead of name (optional)
  * 
  */
 var Service = require("HAP-NodeJS").Service;
@@ -37,6 +39,9 @@ function KNXDevice(log, config) {
 	log("Accessory constructor called");
 	if (config.name) {
 		this.name = config.name;
+	}
+	if (config.uuid_base) {
+		this.uuid_base = config.uuid_base;
 	}
 	if (config.knxd_ip){
 		this.knxd_ip = config.knxd_ip;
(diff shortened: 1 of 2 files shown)


snowdd1:
Compiles and runs fine with my installation (node 4.1.1 on raspi wheezy). Tested together with https://github.com/KhaosT/HAP-NodeJS/pull/144
UUIDs are built on platform name ("KNX" in my test), display names or if present, on uuid_base from accessory.

I wouldn't expect any surprises (well that's sounding obvious) but if somebody could test that with node 0.12 the confidence level would be even higher.

nfarina:
Tested and works great. Thanks for patching this up!!

snowdd1:
Your welcome, I still have to rename my platform to "Object" to keep my database ...
Just kidding.

maddox:
Is this still going to break things?

nfarina:
Yes, but only for platforms you're already using that return accessories which use the "new API" style syntax.

maddox:
@nfarina but we should update platforms to use this new style though right? Hence we'll lose our stuff?

nfarina:
Yes, this was the case before but unfortunately platforms that are upgraded to the new API will end up with different UUIDs. HomeKit won't care but it will remove those accessories from any assigned rooms, scenes, etc.

maddox:
I wasn't following the other couple of threads very closely. What is this providing? Does this have to do with being able to auto-detect new accessories without restarting? (did I read that correctly somewhere?)

nfarina:
No this is unrelated - this fixes the auto generation of UUIDs from platforms which was previously bugged - it was prepending the string "Object" instead of the accessory constructor name.

maddox:
:thumbsup: thanks!

PR #235, opened 2015-10-06
Add Dockerfile

adrienbrault:
https://github.com/nfarina/homebridge/pull/87

This is working for me using compose:

``` yaml
homebridge:
  build: homebridge/
  ports:
  - "51826-51999:51826-51999"
  volumes:
  - /var/run/dbus:/var/run/dbus
  links:
  - home-assistant
  restart: always
```

PS: To use port range with docker-compose, you need the soon to be released 1.5.0 version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:4.1-wheezy
+
+RUN apt-get update -y && \
+    apt-get upgrade -y && \
+    apt-get install -y libavahi-compat-libdnssd-dev && \
+    apt-get -y autoremove && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY package.json /usr/src/app/
+RUN npm install
+COPY . /usr/src/app
+
+EXPOSE 51826:51999
+
+CMD [ "npm", "start" ]


nfarina:
you shouldn't need to expose any ports besides 51826 anymore I think?

adrienbrault:
If I only expose 51826, it doesn't work, I can't see the devices from my iphone.

https://github.com/KhaosT/HAP-NodeJS/blob/master/Core.js#L33
I think every device is bound to a separate port

nfarina:
If you're using the latest Homebridge, it's not using the same system as in `Core.js`, it's using Bridged mode like `BridgedCore.js` (but its own version). All accessories are hosted "behind" a single bridge accessory that is hosted on port 51826 by default.

dgarozzo:
This Dockerfile isn't working for me. I'm curious to know how you have it working. When I go to build, it gives an error. 

npm ERR! telldus@0.0.9 install: `node-gyp configure build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the telldus@0.0.9 install script.
npm ERR! This is most likely a problem with the telldus package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp configure build
npm ERR! You can get their info via:
npm ERR!     npm owner ls telldus
npm ERR! There is likely additional logging output above.

prbreezy:
Hi Guys,

Probably you can use:
> 
--net=host — Tells Docker to skip placing the container inside of a separate network stack. In essence, this choice tells Docker to not containerize the container’s networking! While container processes will still be confined to their own filesystem and process list and resource limits, a quick ip addr command will show you that, network-wise, they live “outside” in the main Docker host and have full access to its network interfaces. Note that this does not let the container reconfigure the host network stack — that would require --privileged=true — but it does let container processes open low-numbered ports like any other root process. It also allows the container to access local network services like D-bus. This can lead to processes in the container being able to do unexpected things like restart your computer. You should use this option with caution.

--> https://docs.docker.com/v1.8/articles/networking/

MrZoidberg:
Is it possible to run it on Raspberry 3?

nfarina:
Closing this issue as it's super old and I can't tell if it is working for people (I haven't tried this myself) - if anyone is having success with this docker config then will reopen!

PR #242, opened 2015-10-08
Simple fix for undefined Nest device names

ethansinjin:
Nest is no longer returning device names in all cases.


nfarina:
Cool thanks!

PR #248, opened 2015-10-09
Add support for local Telldus control

theodorton:
Supported features: Lights On/Off and Dimming.

Adds `telldus` as a dependency.


nfarina:
Thanks!

PR #273, opened 2015-10-16
Add platform support for Indigo server (http://www.indigodomo.com/)

webdeck:
Adds platform support for Indigo (http://www.indigodomo.com/)


nfarina:
Awesome thanks for the PR!

PR #275, opened 2015-10-16
Support for case-sensitive file systems

MSchmidt:
When installing homebridge on a case-sensitive file system it fails to start because it cannot find ``HAP-NodeJS`` even though it is installed correctly.
This is because HAP-NodeJS is located in ``node_modules/hap-nodejs`` and not ``node_modules/HAP-NodeJS``.
Maybe this issue only arises with the current node and npm versions, but here is my pull request to fix this. I've tested this with node 4.2.1 and npm 3.3.6 (current stable versions).


nfarina:
On my Mac, HAP-NodeJS was check out by npm as `HAP-NodeJS` so this will cause that to break.

If you switch the dependency in `package.json` to use the new nom-published HAP-NodeJS, npm should create the directory as `hap-nodejs` since that's the name of the new npm package. Can you try that instead?

MSchmidt:
I've updated the pull request to use the npm release of hap-nodejs now. Everything is checked out in lowercase.

nfarina:
Awesome thanks!

PR #282, opened 2015-10-17
Update app.js

tommasomarchionni:
Hello,
with this string HAP-NodeJS/lib/util/once server doesn't start.
Use hap-nodejs/lib/util/once instead.


nfarina:
Thanks; looks like we missed a spot.

PR #295, opened 2015-10-19
Handle where no model is returned from Telldus Live

aholen:
This will prevent homebridge exiting with stacktrace when there is not "model" for the device returned from the Telldus Live API.


nfarina:
This PR seems to be empty now :) Closing.

PR #298, opened 2015-10-19
this is the latest update to the FHEM platform shim

justme-1968:
i have changed the handling of hap-nodejs vs. HAP-NodeJS to allow both as i and some users are still on the old version and can not update immediately.

the changes:

check for hap-nodejs and HAP-NodeJS
allow longpoll to multiple fhem servers
increase reconnect timeout after subsequent longpoll disconnect and error
added first version of HM-SEC-KEY-S
added target door state for HM-SEC-KEY -> allows open to be send
added current door state to contact sensor -> allows to ask for open windows with siri
added EnOcean blinds
added ROOMMATE devices
added timestamp to longpoll console output
added CurrentAmbientLightLevel characteristic
added firmware revision characteristic
set uuid_base to real device serial number if available
changed harmony device to a single accessiory with mutliple switches
added hue colormode & xy to hsv conversion
added MAX shutter contacts
parse min/max/step values for thermostats
use this/bind instead of that
added genericType ignore
added batter state charactersitic
use new characteristic.setProps() method
lots of cleanups


nfarina:
Can you pull in latest from master? There are some conflicts, most likely package.json or something.

justme-1968:
the conflict was my different approach to the hap-nodejs vs. HAP-NodeJS handling. should be resolved.

PR #313, opened 2015-10-24
Fix MyQ Garage Opener and multiple Nests configuration

fcarucci:
This patch checks the MyQDeviceTypeName property against VGDO, which is used by MyQ to identify a "virtual garage door opener".
It also fixes the scenario where multiple Nests are present: the accessory name needs to be unique.


nfarina:
Thanks!

PR #320, opened 2015-10-24
LiftMaster: return garage door state

fcarucci:
This patch will return the gare door state to HK. Siri can now query the state of the door.

### DIFF
--- a/accessories/LiftMaster.js
+++ b/accessories/LiftMaster.js
@@ -258,7 +258,7 @@ LiftMasterAccessory.prototype = {
         onUpdate: null,
         perms: ["pr"],
         format: "string",
-        initialValue: "Garage Door Opener Control",
+        initialValue: "Garage Door",
         supportEvents: false,
         supportBonjour: false,
         manfDescription: "Name of service",
--- a/accessories/LiftMaster.js
+++ b/accessories/LiftMaster.js
@@ -16,7 +16,15 @@ LiftMasterAccessory.prototype = {
 
   setState: function(state) {
     this.targetState = state;
+    this.callback = undefined;
     this.login();
+  },
+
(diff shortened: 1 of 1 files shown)


nfarina:
Hey nice! I'm a bit worried about the shared `this.callback` variable though - if two devices call `getState` within a few seconds of each other, the first callback will be lost. Can the `callback` parameter simply be passed on to the other methods and not stored as a (essentially global) instance variable?

fcarucci:
Sure, I’ll give it a go this evening.
Doesn’t the original code have the same problem with targetState? I followed the same pattern.

- Fran

nfarina:
Hah, indeed it does. And of course it was me who did it that way :) Shame on my younger self!

I'm OK merging this as-is if you'd like - if you have time to fix both `targetState` and `callback` though that would be awesome.

fcarucci:
Cool, let’s do that. I’m very rusty in JS (  euphemism :) ) but I’ll try.

nfarina:
Merging this for now.

fcarucci:
Thanks! I have a question if you don’t mind, do you know if there is a way to expose a scene to HomeKit?
I’m trying to bridge Lutron Caseta scenes to HK.
Thanks!

nfarina:
There is no way to expose scenes to HomeKit that I'm aware of unfortunately. You could simulate something though by creating a simple "Switch" accessory that, when enabled, turns on the Lutron scene. But you can't expose "native scenes" directly.

fcarucci:
Yep, that’s how I’m doing it at the moment. Feels hacky, but it’s the only way…
I’m looking into exposing the Automatic car device next to ask questions like “Is my car on?”.

PR #337, opened 2015-10-28
Fixed issue on Philips Hue where initial state was not being loaded

KraigM:
Fixed issue on Philips Hue where initial state was not being loaded.  (As noted in comments) it needs a better way to load/refresh the state, but this is a quick fix.


KraigM:
That was quick, lol

nfarina:
😎

Issue #345, opened 2015-10-29
Settings multiple values once, makes the PhilipsHue platform crash


patricks:
I am using an app, where i can change a whole set of hue lamps and there settings.

If I start such a scene. I get the following crash:

```
[Phillips Hue] Set Bedroom Lamp, characteristic: brightness, value: 70.
  EventedHTTPServer [::ffff:10.0.1.4] Muting event '18.9' notification for this connection since it originated here. +29ms
  EventedHTTPServer [::ffff:10.0.1.11] Sending HTTP event '18.9' with data: {"characteristics":[{"aid":18,"iid":9,"value":true}]} +1ms
[Phillips Hue] Set Living Room Lamp, characteristic: power, value: true.
/Users/patrick/Developer/homebridge/platforms/PhilipsHue.js:241
            this.executeChange(characteristic, value, callback);
                 ^

TypeError: this.executeChange is not a function
    at null._onTimeout (/Users/patrick/Developer/homebridge/platforms/PhilipsHue.js:241:18)
    at Timer.listOnTimeout (timers.js:92:15)
```

yhellowt:
I am having the same issue. Everything works fine until >7 actions need to be processed.

andersboe:
I am experiencing the same issue as well. As @yhellowt said, it seems to happen once too many actions are called.

andersboe:
Fixed the issue, made a pull request, check it out :)

dvinc:
Commit 70f2dc62 seems to be related to solving the issue.

Issue #349, opened 2015-10-30
Homebridge crash with too many 'Philips Hue' actions


andersboe:
First off: Love this project! 

I have got stuck into an error with the Philips Hue shim, that seems to fail if too many actions are called at a time. 

The homebridge server crashes when I try to set 'Power', 'Color' and 'Brightness' of three lights. 
Most of the scene get set correctly before the server crashes. 

(Setting 'Power', 'Color' and 'Brightness' for two lights works fine, same goes for setting two of the properties for all three lights. The limit seems to be somewhere around 8-9 "actions")

Below is the log from the call until the crash:
```
EventedHTTPServer [::ffff:10.0.1.4] HTTP request: /characteristics +5s
  HAPServer [XX:XX:XX:XX:XX:XX] HAP Request: PUT /characteristics +0ms
  Accessory [Homebridge] Processing characteristic set: {"characteristics":[{"aid":5,"iid":11,"value":0},{"aid":4,"iid":11,"value":0},{"aid":4,"iid":9,"value":1},{"aid":3,"iid":11,"value":0},{"aid":5,"iid":9,"value":1},{"aid":3,"iid":9,"value":1},{"aid":3,"iid":10,"value":56},{"aid":4,"iid":10,"value":56},{"aid":5,"iid":10,"value":56}]} +0ms
  Accessory [Homebridge] Setting Characteristic "Hue" to value 0 +1ms
  Accessory [Homebridge] Setting Characteristic "Hue" to value 0 +2ms
  Accessory [Homebridge] Setting Characteristic "On" to value 1 +1ms
  Accessory [Homebridge] Setting Characteristic "Hue" to value 0 +0ms
  Accessory [Homebridge] Setting Characteristic "On" to value 1 +0ms
  Accessory [Homebridge] Setting Characteristic "On" to value 1 +0ms
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 56 +0ms
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 56 +1ms
  Accessory [Homebridge] Setting Characteristic "Brightness" to value 56 +0ms
[Phillips Hue] Set Hue Spot 2, characteristic: power, value: 1.
[Phillips Hue] Set Hue Spot 2, characteristic: hue, value: 0.
[Phillips Hue] Set Hue Spot 3, characteristic: hue, value: 0.
[Phillips Hue] Set Hue Spot 3, characteristic: power, value: 1.
[...]

PR #358, opened 2015-11-01
Philips hue fix

andersboe:
This pr fixes issue #345 and #349 

Solution was to add `var that = this` logic in the `executeChange` and `getState` functions.
Not sure why it's not already done (maybe there is some reason for that), as it is implemented in other parts of the code. 

Also added some automatic code style cleanup in the lower half to keep the file uniform.

### DIFF
--- a/platforms/PhilipsHue.js
+++ b/platforms/PhilipsHue.js
@@ -274,15 +274,15 @@ PhilipsHueAccessory.prototype = {
         if (newValue != undefined) {
           callback(null, newValue);
         } else {
-          //  this.log("Device " + that.device.name + " does not support reading characteristic " + characteristic);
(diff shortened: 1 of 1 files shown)


dvinc:
There was a missing bind(this) at l.242, but it has been fixed up in 70f2dc6. This caused error when then bridge message rate limiting caused the retry timeout callback to be called. Did you check if this fixed your issues #345 and #349?

andersboe:
That's correct. Testet out the latest repo, and it seems like everything runs ok. 

I cloned the repo two times at saturday, must be just before and after 70f2dc6, as one of them has the bind, and the other does not - hence I also added it in 8ee289e thinking it was something I randomly deleted.. :)

nfarina:
Is this OK to close? Trying to tie up loose PRs before some major changes.

andersboe:
Yep!

PR #359, opened 2015-11-01
Fixed humidity on Nest

KraigM:
Fixed humidity and added change handling of settings (like Celsius/Fahrenheit).

### DIFF
--- a/platforms/Nest.js
+++ b/platforms/Nest.js
@@ -41,14 +41,26 @@ NestPlatform.prototype = {
                         }
                     }
                     function subscribe() {
-                        nest.subscribe(subscribeDone, ['shared']);
+                        nest.subscribe(subscribeDone, ['device','shared']);
                     }
 
                     function subscribeDone(deviceId, data, type) {
                         // data if set, is also stored here: nest.lastStatus.shared[thermostatID]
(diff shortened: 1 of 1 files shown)


pedroserano:
Hi KraigM,

These fixes are very much welcome. I wanted to build on this to add home/away status (which could be used as a trigger in ios9) and fan on/auto read/write support, both of which seem to be exposed in the unofficial nest api, I think. I started working on the fan system and despite my best efforts have come up short. If I pasted the code, could you look at it and point me in the right direction? Would it be easier for you to add those things without also trying to debug?

KraigM:
Awesome. Yes you can paste code and I will look at it and see if I can help you out, no problem.

KraigM:
@pedroserano ^. Also, I don't know if you have switched to the new plugin system yet or not.  If you haven't, don't worry about it, I can help you integrate later.  But if you have, I split out nest into its own plugin: https://github.com/KraigM/homebridge-nest.  So we can maintain it better.  Feel free to send a pull request to that and I can look at it and hopefully help you out.

pedroserano:
Yes totally! I have the new plugin system installed on my system. I will use the new address to submit a pull request.

PR #387, opened 2015-11-08
Provide and output proper existing error message

knalli:
If the found plugin / package does not fulfill some specs (i.e. does not contain the keyword `homebridge-plugin`), it will be rejected with an error. Without that change nobody will be notified WHY the plugin has been rejected/skipped.

That is only a proposal, perhaps we can solve the problem in another way. But that specific information forced spending too much time.


nfarina:
I think it's a good idea. In this case you're helping users identify packages that are "trying" to be a homebridge plugin but fail one of the tests.

PR #398, opened 2015-11-18
Smartthings

lucaschristian:
[Review on Reviewable](https://reviewable.io/reviews/nfarina/homebridge/398)


nfarina:
Hello, this looks like good work but Homebridge "core" doesn't contain anything related to actual devices - instead this should be made into a Plugin. Check the new README for more details.

Issue #385, opened 2015-11-07
IOS Won't pair ... but the fix given do not work


sangoku71:
Hello everybody,

I have a little problem with home bridge since this morning : my IOS won't pair.
I tried the 2 methods given in the issue 381 : 
I mean deleting persist folder and changing the name in config,json but nothing happened ...

Can someone help me ?

in advance THX

Sango

askme233:
Check to make sure you do not have other devices with the same iCloud ID that may have replicated any issues (although changing the name should have helped).  I assume you have gone into settings/privacy/homekit and erased the homekit DB on your iDevice?

Be2daErtin:
when changing your bridge name, also change the username and see if that works.

sommertrip:
I have the same issue, changed everything (bridge name, username, tried as well issue 381: deleting persist folder. nothing helped.

Be2daErtin:
Turn keychain off and back on again....

sommertrip:
did this 3 times, can not find the homebridge anymore

williamboliver:
Same issue.   tried keychain restart and am only getting my hue bridge.   Dead in the water...

askme233:
I suggest downloading bonjour explorer, discovery or similar so that you can see the details of the bonjour advert.  Certainly reset everything (delete persist, wipe HomeKit DB, reset phone, etc). If you see the bonjour service advertised, but don't pick it up in the app, then you know it is an app recognition problem. If the app sees the accessory, but will not pair it is likely either a problem on the homebridge side (check all the logs). If you don't see homebridge advertised on bonjour, or it is eg on the wrong network, then you need to fix that first.

sommertrip:
o.k thats what I see in bonjour explorer. I don´know which kind of advertised service you mean?
![voila_capture 2015-11-25_09-22-44_am](https://cloud.githubusercontent.com/assets/15965995/11391907/4e01f7c4-9356-11e5-9f2e-628e52431ff8.png)

amitgandhinz:
im running into the same issue.  It was working fine until i made some changes to my config.  Those changes werent getting picked up so i tried deleting the device from the app.  Now it wont detect the homebridge using the insteon+ app (even after reverting my config changes).  HAP shows up from my RPi on Bonjour Discovery.  Have also reset the phone (several times), removed the Home from the device, and rm the /persist folder.

rudders:
[...]

PR #442, opened 2015-12-27
Add test for valid username

n8henrie:
Like a MAC address, a valid username can only contain 6 pairs of
hexadecimal characters, and for homebridge purposes must be
colon-separated. This adds a simple regex test for a proper username,
since I just spent nearly 24h trying to debug an issue that all came
down to not realizing that a MAC address could only have letters A-F
(see https://github.com/nfarina/homebridge/issues/385).

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -27,16 +27,16 @@ function Server() {
 }
 
 Server.prototype.run = function() {
-  
+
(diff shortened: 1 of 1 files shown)


nfarina:
Wonderful thanks!

PR #450, opened 2016-01-03
Trying 0.0.7 version of hap-nodejs

nachtsheim:


rudders:
If this brings us stability especially to RPi then whoohoo!!!

nfarina:
Oops missed this PR - I just bumped to 0.0.7 in c8c84c0.

PR #467, opened 2016-01-13
Updated README.md to explain Startup Errors.

rudders:
This issue is frequently discussed so let's see if people read the documentation...


nfarina:
Good call!

rudders:
Whoohoo - now formally a contributor to the project ;) If I can get the bloody wemo platform stable I'll try and help out a bit more with the doco.

Issue #477, opened 2016-01-19
Simple UI for managing / customizing plugins


epheterson:
While the CLI / config file system is functional, it's not the most approachable. It'd be nice to track a longer-term UI to streamline the process of using HomeBridge.

KhaosT:
@epheterson surprise to see you here 😄It's on the roadmap for a while but everyone is kinda busy right now so that will probably arrive later :)

epheterson:
Didn't know you were on here either! Looking forward to it as the customization requires you go back to each plugin's documentation instead of being intuitively discoverable.

cflurin:
I'm working on it at first for my plugin (homebridge-punt). It will be similar to the already implemented simulator.

nfarina:
It's something I'd love to have as well. I started working on this a while ago in the `provider-API` branch, but it has a ways to go.

KraigM:
Im actually working on something along those lines right now.  I have it so you can load your config and installed plugins, make modifications to the config, and save the config, all from anywhere on the network (secured by the access key, i.e. "174-33-129")

epheterson:
To be clear, when I suggested UI, I did not mean in-browser UI, or UI to edit the config file via text. Although that's a step, an ideal solution would be more akin to the UI to discover / install / customize plugins for browsers such-as Firefox or Chrome. 

I'd love to play with any half-baked solutions any of you have made!

bugs181:
I suppose what you're really looking for then is a sort of UI for `npm` that looks for the `homebridge-` prefix and then uses the `package.json` file. Inside that, you can have a `screenshotUrl` property that displays a screenshot, an install button, and perhaps some sort of social/rating platform. It all seems kind of elaborate but I think it could work.

epheterson:
Hey, that's really half of the puzzle. The other half is actual configuration of the plugin. For example, if I download the homebridge-cmd plugin, a UI to visually convey to the user that one accessory will exist, with whichever title, running whichever commands when the boolean goes to each state. Then a (+) button to add an additional toggle / accessory and configure it. 

I'd like for this to be more approachable for people who may never want to open a text editor, copy/paste lines to the correct section, modify and verify syntax before relaunching homebridge.

bugs181:
[...]

Issue #465, opened 2016-01-11
Dynamically reload platform accessories


johannrichard:
Homebridge platform plugins like OpenHAB or [the one I recently build](http://github.com/johannrichard/homebridge-mystrom) all connect to systems where the number of accessories can and will change dynamically over time. Unless I missed it, `homebridge` does not provide any means to dynamically reload the accessories a platform provides, and hence, one has to restart the process every time a new accessory is added to the platform. 

It would be nice if this was changed at some point.

snowdd1:
See #237, this would require plugins to be able to restart the bridge's advertiser.

AppleTechy:
@johannrichard My understand is that the homekit platform doesn't allow dynamic device. Though I believe if my understand is correct iOS 9.3 beta does allow dynamic device loading which then the homebridge plugin can be edited to support this new function in the newest version of HomeKit. If that makes any since...?

KhaosT:
@AppleTechy Apple allows bridge to add/remove bridged accessory. It's just hap-nodejs doesn't implement that initially because the upper stream package (mdns) doesn't support update TXTRecord. Now I fixed that so it's possible for homebridge to make necessary changes to adopt this :)

PR #497, opened 2016-01-31
Plugin 2.0

KhaosT:
🚧Work In Progress🚧
This PR tries to give plugin the ability to dynamically add/remove accessory, update reachability of the accessory and provide a way for user to configure platform plugin without having them to touch config.json

This PR shouldn't break the existing plugin but in order for plugin to use the new feature, there are some changes plugins need to make on their side. Please check ``./example-plugins/homebridge-samplePlatform`` for more details.

I made a simple configurator to help test this PR, you can find it at [here](https://github.com/KhaosT/HesperusBridgeConfigurator).

This PR should addressed #465 and partially addressed #477. I'm not sure what's the best solution for user to install plugins other than having them to run ``npm install``.

/cc @cflurin, @snowdd1, @SphtKr, @maddox, @nfarina, @KraigM, @rudders


nfarina:
Whoa this is crazy stuff!

pdlove:
I'll try this out with the wink platform plugin before the end of this week and provide feedback. I'm very excited about this!

AppleTechy:
@pdlove The one thing I have heard that I don't like about it is that you
loose the ability to change the Siri name/ accessory name to something
else. Though I might have misunderstood someone else's comment

KhaosT:
@AppleTechy config.json is not going anywhere. However, for Accessory/Service name, you should always change that in HomeKit instead of change it on the "physical" accessory...

AppleTechy:
@khaosT So even with the new dynamic loading you can still manually change
inside a HomeKit compatible app the name of a specific accessory as well as
the Siri name for the device?

KhaosT:
@AppleTechy You can always do that in HomeKit even before this PR…

pdlove:
@appletechy the only item is that if the plugin removes the device instead of marking it as inaccessible then you lose all on the HomeKit configuration for the Accessory. For example, the Wink platform will only mark a device as removed when it is actually removed from Wink but will mark it as inaccessible if the hub can't communicate with it or the service is down for some other reason.

AppleTechy:
@khaosT and @pdlove Ok thank you for clarifying. I think I got confused
with what Marci said about not being able to edit the name but that seemed
to be an issue with the plugin in they were running not actually the
Homebridge 2.0

PatchworkBoy:
Nope - twas an error in my code for the eDomoticz plugin - this.name not being passed to the new service instance. All cleared up now.

snowdd1:
Hi @KhaosT , while playing with the plugin-2 fork, I found you've been passing the API twice:
Once with the platform registering function
https://github.com/nfarina/homebridge/pull/497/files#diff-388d8509adb882d4cf13e988b38c5a90R4
(variable called "homebridge" here)
Once with the platform constructor
https://github.com/nfarina/homebridge/pull/497/files#diff-388d8509adb882d4cf13e988b38c5a90R23
(variable called "API" here)

Is this the same object or did I miss something?

KhaosT:
It's the same object but in different context. The first one is for plugin to register itself (no plugin instance yet) and the second time is during plugin instance being initialized.

snowdd1:
Ok. I used to save the API object in a module-global variable, and for that it's no difference. I just ended up with to references to the same object then.

BTW: Seems to be heck of work to re-connect the accessories with their implementations... Right now I am trying to figure our how to do that reliably. UUIDs seem to be the most reliable reference.

KhaosT:
@snowdd1 the amount of work to handle accessory restoration really depends on the platform I guess, some platform offers uuid (like philips hue) which can significantly reduce the complexity of restoration. You can also assign "context" to PlatformAccessory object and help you restore accessories.

snowdd1:
@KhaosT 
Still investigating this:

1.  The characteristics in each service have to be unique, that is each service must not have more than one characteristic of a given type, right?

1.  The services in each accessory can be non-unique, but then they have to have distinct sub-types, right?

So I need to iterate through all services that are restored by plugin-2 and look into my persistence to find a matching configuration, and bind it to the right events?

Second thing I am wondering about, is how to change the configuration. If I need to add a characteristic (or a service) to an existing accessory, can I simply invoke the add methods during runtime?

PatchworkBoy:
> This PR shouldn't break the existing plugin
Can confirm everything working correctly [homebridge-edomoticz](https://github.com/patchworkboy/homebridge-eDomoticz) after pulling plugin-2. Nothing broken.

PatchworkBoy:
As can’t get Hesperus to build, I manually hit up http://rasprf.local:18081/add, which immediately triggers the following with the samplePlatform code:

```
Add Accessory
/usr/local/lib/node_modules/homebridge/lib/platformAccessory.js:84
  this._associatedHAPAccessory._sideloadServices(this.services);
                               ^
TypeError: undefined is not a function
    at PlatformAccessory._prepareAssociatedHAPAccessory (/usr/local/lib/node_modules/homebridge/lib/platformAccessory.js:84:32)
    at Server._handleRegisterPlatformAccessories (/usr/local/lib/node_modules/homebridge/lib/server.js:399:15)
    at Server.<anonymous> (/usr/local/lib/node_modules/homebridge/lib/server.js:32:10)
    at API.emit (events.js:107:17)
    at API.registerPlatformAccessories (/usr/local/lib/node_modules/homebridge/lib/api.js:143:8)
    at SamplePlatform.addAccessory (/usr/local/lib/node_modules/homebridge/node_modules/homebridge-samplePlatform/index.js:188:12)
    at SamplePlatform.<anonymous> (/usr/local/lib/node_modules/homebridge/node_modules/homebridge-samplePlatform/index.js:31:12)
    at Server.emit (events.js:110:17)
    at HTTPParser.parserOnIncoming [as onIncoming] (_http_server.js:491:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:111:23)
error: Forever detected script exited with code: 1
```

KhaosT:
@PatchworkBoy Please make sure after the pull request, you run npm install again so all dependencies are installed correctly. This PR uses a newer version of hap-nodejs

PatchworkBoy:
Sorted! Thanks.

KhaosT:
@snowdd1 

> The characteristics in each service have to be unique, that is each service must not have more than one characteristic of a given type, right?

Yes, characteristic has to be unique within a service. 

> The services in each accessory can be non-unique, but then they have to have distinct sub-types, right?

Yep, you can have multiple services with the same type. Sub-type is an ideal used in hap-nodejs to help developer differentiate services.

> So I need to iterate through all services that are restored by plugin-2 and look into my persistence to find a matching configuration, and bind it to the right events?

Correct. I'd suggest you go with the inverse way. When homebridge asks plugin to configure the accessory, plugin save the accessory and mark as unreachable. After plugin receive didFinishLaunching event, it starts normal logic to discovery real accessories. Upon finding the accessories, lookup the saved accessories and configure from there. If the accessory is not there, invoke API.registerAccessories.

> Second thing I am wondering about, is how to change the configuration. If I need to add a characteristic (or a service) to an existing accessory, can I simply invoke the add methods during runtime?

Right now it's not supported but I can definitely add this if it's really useful in some cases. Can you describe the use case you're thinking for having this feature?

snowdd1:
@KhaosT 
Thanks for your answers. The inverse method sounds easier, will try that.

The use case for online editing of devices (that is adding services or characteristics) is that the system my platform (homebridge-knx) supports has no discovery. All configuration needs to be done either by text file (now - in the config file, you can find all accessories here in this section https://github.com/snowdd1/homebridge-knx/blob/master/KNX-sample-config.json#L22) or (planned) in a web UI which should be able to add devices, but also change them. Users who have a KNX enabled home usually have almost **everything** wired to that system, as it is usually built into the house at build or renovation time, and is not a add-on system. In my home, I have all lights (about 40), all thermostats (about 13), all rolling shutters (20), the sliding gates of the parking lot, etc. controlled by the system. 
As it is a legacy system with restrictions originating from the 1980's it is quite user unfriendly, and users will have to make corrections to the config quite often I guess. The only way I see with plugin-2 is to unregister the accessory and re-register it with new properties, which will remove the device from rooms and scenes and everything AFAIK, if a homekit enabled device queries /accessories in that time, or if the uid/aid etc. changes in that process.
BTW: I think I am going to assign an arbitrary globally unique subtype to each service for easier reference/lookup, do you see any obstacles to that?
Thanks!

snowdd1:
@KhaosT 
I tried (step by step) to migrate to plugin-2, but I am failing even before doing any changes... (platform is still on dynamic=false)
Is this fork completely incompatible with iOS 9.1.x? 
EVE lets me add homebridge, but shows me one accessory (the bridge) only, with 3 characteristics (Control Point [n/a], State [0], Version [1.0]) and tells me that the accessory has sent an illegal answer. I cannot see any accessories behind the bridge. 
Is this expected?

While typing things are changing:
After some minutes I can see the devices behind the bridge, seems to be some background sync iCloud thing...
But I still see the bridge with Control Point being "n/a" 

After even some time more, the bridge itself vanished from EVEs overview and the accessories become available and controllable. Strange. Now it's completely working...

PatchworkBoy:
> If I need to add a characteristic (or a service) to an existing accessory, can I simply invoke the add 
> methods during runtime?

> Can you describe the use case you're thinking for having this feature?

This part has me the most curious - with Domoticz one can frequently switch the type of a sensor to settle on one which contains the most characteristics (i.e.: I can flip a light in domoticz from just being a switch, to being a switch with dimmer, to being a switch with hue)

It would be good to be able to do this and on the next ‘get’ of that device via Homebridge, re-poll for new / removed services and add / remove these on demand. That way it’s dynamic end-to-end not only in the accessories available, but the characteristics of existing accessories... reflecting any changes to either.

One could fire up homebridge once, and never need to restart it regardless of changes at the platform - all could be detected. T’would be a very strong feature to have imo...

snowdd1:
@PatchworkBoy : If the characteristics change, those that have been changed cannot survive in scenes etc. Only the static elements can.

KhaosT:
@snowdd1 I don't have iOS 9.1.x to test right now but changes made in plugin-2.0 shouldn't break earlier iOS support. Can you try to reproduce this issue and pull error information from HomeKit on iOS and post at here?

KhaosT:
@snowdd1 @PatchworkBoy OK, I'll try to make changes in hap-nodejs and bring those to this PR for modifying service/characteristic.

snowdd1:
@KhaosT 
Looks like I messed it somewhat by turning dynamic=true previously, that made iOS a little dizzy under the hood. At least with non-dynamic platform settings everything is good with iOS 9.1.2.

snowdd1:
@KhaosT 
I am wondering which methods of the SamplePlatform.prototype belong to the interface (i.e. must be identical in signature for  all platforms) and which are not.
I found that you made `'didFinishLaunching'` an event, but `configureAccessory()` is a straight function, as well as the `configurationRequestHandler`.
May I suggest to push the interface functions to top and give them a javadoc-like header? I'll do if you point me to.

KhaosT:
@snowdd1 sorry about the confusion, I'm terrible at documenting things 😅

``'didFinishLaunching'`` is an event emitted by API object when homebridge finishes launching sequence. Plugin should only try to register new accessories after this event.
``configureAccessory()`` and ``configurationRequestHandler()`` are new interfaces that Platform.prototype needs to provide in order for homebridge to give plugins the ability to restore accessories and handle configuration request.

snowdd1:
@KhaosT : 
during my work at the adaptation of my dev branch of the KNX plugin I found that you can create multiple services (like lightbulbs) with same UUID and a different subtype (see questions above) in `PlatformAccessory.prototype.addService`  , but `PlatformAccessory.prototype.getService(name)` does not support returning such services. It would always return the first found only.
To be sure to hit the right service during configuration, I have to look for UUID *and* subtype. That wouldn't harm anyone, if subtype is empty the result should be the same as without subtype parameter.

May I add a `getServiceByUUIDAndSubType(UUID, subtype)` in a PR?

snowdd1:
@KhaosT RE 9.1, sorry I was confused. Was talking about 9.2 vs 9.3, not, as I wrote, 9.1 vs. 9.2. Should have looked it up before writing. I do not have 9.1 devices either. 9.2 is good, though.

snowdd1:
@KhaosT 
I just found the signature of `platformCharacteristic.addService(service)` https://github.com/nfarina/homebridge/blob/plugin-2/lib/platformAccessory.js#L36
does not match your invocation in the samle  `addService(service, string)` https://github.com/nfarina/homebridge/blob/plugin-2/example-plugins/homebridge-samplePlatform/index.js#L180
Though it does no harm, it is better understandable for fellow developers if they match.

KhaosT:
@snowdd1 we don't want to encourage developer to include service name unless they know what they're doing. HomeKit encourages app developer to use service name to decide whether it should be displayed to user or not.

snowdd1:
@KhaosT Sorry, did not get your last answer. The latter `string` is just ignored, as the parameter is not used by addService. So in your example the string "Test Light" will be ignored.
Or am I completely wrong?

example-plugins/homebridge-samplePlatform/index.js
```diff
@@ -0,0 +1,197 @@
...
+
+  uuid = UUIDGen.generate(accessoryName);
+
+  var newAccessory = new Accessory(accessoryName, uuid);
+
+  // Plugin can save context on accessory
+  // To help restore accessory in configureAccessory()
+  // newAccessory.context.something = "Something"
+
+  newAccessory.addService(Service.Lightbulb, "Test Light")
```
snowdd1: The second parameter is not used in platformAccessory.addService()

example-plugins/homebridge-samplePlatform/index.js
```diff
@@ -0,0 +1,197 @@
...
+// Function invoked when homebridge tries to restore cached accessory
+// Developer can configure accessory at here (like setup event handler)
+// Update current value
+SamplePlatform.prototype.configureAccessory = function(accessory) {
+  console.log("Plugin - Configure Accessory: " + accessory.displayName);
+
+  // set the accessory to reachable if plugin can currently process the accessory
+  // otherwise set to false and update the reachability later by invoking 
+  // accessory.updateReachability()
+  accessory.reachable = true;
```
snowdd1: I tried to set that to `false` and it had no effect on my test device (iOS 9.3b2, EVE app)

lib/platformAccessory.js
```diff
@@ -0,0 +1,177 @@
...
+  if (!displayName) throw new Error("Accessories must be created with a non-empty displayName.");
+  if (!UUID) throw new Error("Accessories must be created with a valid UUID.");
+  if (!uuid.isValid(UUID)) throw new Error("UUID '" + UUID + "' is not a valid UUID. Try using the provided 'generateUUID' function to create a valid UUID from any arbitrary string, like a serial number.");
+
+  this.displayName = displayName;
+  this.UUID = UUID;
+  this.category = category || Accessory.Categories.OTHER;
+  this.services = [];
+  this.reachable = false;
+  this.context = {};
```
snowdd1: @KhaosT : I thought the context object to be persistent over restarts, but it is not?

KhaosT:
@snowdd1 notice [here](https://github.com/nfarina/homebridge/blob/plugin-2/lib/platformAccessory.js#L40), we uses [arguments object](https://docs.nodejitsu.com/articles/javascript-conventions/what-is-the-arguments-object) to construct the service so the string is not ignored.

lib/platformAccessory.js
KhaosT: @snowdd1 right now, homebridge only guarantee to persist context object at the time you invoke API.registerAccessories(). Are you saying that the context you set on PlatformAccessory before invoking API.registerAccessories() is not persisted after restart?

snowdd1: Yes it was empty, but I am a bit tired and was chasing inexistent other errors. I changed identification to UUID and that is pretty stable now. 
https://github.com/snowdd1/homebridge-knx/blob/plugin-2.0/index.js#L148

Code is still ugly and full of debug infos but is is running and survives a restart. Late adding of new accessories is not in there yet, I first wanted to save the status quo before going for the new features.

snowdd1:
You are right, I overlooked that - and it's a nightmare for documentation, as the possible parameters are not listed in the signature...

example-plugins/homebridge-samplePlatform/index.js
KhaosT: @snowdd1 It appears to be a bug on iOS side. HomeKit ignores the original value of reachable characteristic after pair-verify. Sending event can correctly update the reachability. Please check the latest commit with "/reachability" endpoint in sample platform.

KhaosT:
@snowdd1 That's true, so you prefer to add the display name back to method signature and document it so developer won't misuse the API?

devbobo:
@KhaosT what's the intention of `context` ? To store arbitrary data or something else ?

KhaosT:
@devbobo yeah, to store arbitrary data that plugin may needed to restore accessories when homebridge restarts. It can be name, UUID, preferences or anything that can be serialized to JSON.

devbobo:
I've converted my Lifx plugin to the new format, I everything working except for a custom `Characteristic`. I'm not sure how to implement it correctly so it get instantiated correctly when it's reload.

Are you able to provide an example of how to do this in your sample plugin ?

https://github.com/devbobo/homebridge-lifx-lan/tree/plugin-2

KhaosT:
@devbobo which characteristic are you talking about? ``Kelvin``?

devbobo:
@KhaosT yeah, Kelvin

KhaosT:
@devbobo Try to add this line at [here](https://github.com/devbobo/homebridge-lifx-lan/blob/plugin-2/index.js#L51)? I haven't figure out a way to persist type constructor so currently if you defines a custom characteristic constructor, you'll need to provide a "UUID"
```
Kelvin.UUID = 'C4E24248-04AC-44AF-ACFF-40164E829DBA';
```

lib/platformAccessory.js
snowdd1: Hi @KhaosT , had time today to check that. 
Setting `context='SomeStringValue';` somehow failed, and I have not really an idea why, could still be my fault. It somehow came back as an Array instead of a String.
Setting `context.subobject='SomeStringValue'` worked perfectly.
So I guess it's only me failing at using it right!

KhaosT: @snowdd1 Glad to know you get it working. I didn't think about the case that developer will assign a string directly to context... It suppose to be a dictionary but setting a string shouldn't fail.

devbobo:
I've been trying to run the branch in my house for the past few days. I was experiencing an issue where my homebridge instance was being automatically removed from HomeKit after a period of time.

After a lot of hair pulling, I believe that I narrowed it down the `Control Point` characteristic that Homebridge server now adds.

This morning I disabled [this line of code](https://github.com/devbobo/homebridge/commit/cefcfb44af73a3e0969dca71f460b15eb9605813
) and my server has been running without incident since then (around 5 hours)

cc @KhaosT

KhaosT:
@devbobo interesting, were you able to get any logs from iOS device when this occurred? Not really sure why this characteristic would cause the problem …

devbobo:
@KhaosT I don't, I will try to reproduce on a different server

Should the `Control Point` characteristic be `undefined` ? Wouldn't that result in an `Out of Compliance` error ?

```
Wed, 10 Feb 2016 23:41:45 GMT Accessory [Homebridge] Getting value for Characteristic "Control Point"
Wed, 10 Feb 2016 23:41:45 GMT Accessory [Homebridge] Got Characteristic "Control Point" value: undefined
```

KhaosT:
@devbobo I'm not sure what will JSON.stringify would do to ``value: undefined``, if it removes "value" field in the JSON then it definitely causes problem. I made a few changes in hap-nodejs to better handle the "undefined" value, before you try to reproduce the issue, please make sure you update the hap-nodejs ^^

devbobo:
it works, except that the default value for the `Control Point` is `null`

Here's the logs when I ran that....

```
  EventedHTTPServer [::ffff:172.16.1.52] HTTP request: /characteristics?id=1.10 +4ms
  HAPServer [CD:12:3D:E3:CE:30] HAP Request: GET /characteristics?id=1.10 +0ms
  Accessory [Homebridge Test] Getting value for Characteristic "Control Point" +2ms
  Accessory [Homebridge Test] Got Characteristic "Control Point" value: null +0ms
  EventedHTTPServer [::ffff:172.16.1.52] HTTP Repsonse is finished +1ms
  EventedHTTPServer [::ffff:172.16.1.52] HTTP request: /pairings +630ms
  HAPServer [CD:12:3D:E3:CE:30] HAP Request: POST /pairings +1ms
  HAPServer [CD:12:3D:E3:CE:30] Removing pairing info for client +0ms
  Accessory [Homebridge Test] Unpairing with client 1DBB82E5-3CD9-4FAE-BA29-A06845EE9E3D +1ms
  EventedHTTPServer [::ffff:172.16.1.52] HTTP Repsonse is finished +2ms
  EventedHTTPServer [::ffff:172.16.1.52] Client connection closed +43ms
  EventedHTTPServer [::ffff:172.16.1.52] HTTP connection was closed +0ms
  EventedHTTPServer [::ffff:172.16.1.52] HTTP server was closed +1ms
  EventedHTTPServer [::ffff:172.16.1.24] New connection from client +431ms
  EventedHTTPServer [::ffff:172.16.1.24] HTTP server listening on port 54826 +0ms
  EventedHTTPServer [::ffff:172.16.1.24] HTTP request: /pair-verify +2ms
  HAPServer [CD:12:3D:E3:CE:30] HAP Request: POST /pair-verify +1ms
  EventedHTTPServer [::ffff:172.16.1.24] HTTP Repsonse is finished +1ms
  EventedHTTPServer [::ffff:172.16.1.24] Client connection closed +4ms
  EventedHTTPServer [::ffff:172.16.1.24] HTTP connection was closed +0ms
  EventedHTTPServer [::ffff:172.16.1.24] HTTP server was closed +1ms
```

Changing to default value to something other than `null` makes it work as expected.

KhaosT:
@devbobo ``null`` should be an acceptable value for empty data. Do you mean your bridge disappears when Control Point returns ``null`` ?

devbobo:
@KhaosT yes, it disappears...if you look at the log, the homebridge is unpaired after returning that characteristic

devbobo:
@KhaosT I've just retested and it seems to be working fine with `null`, not sure what happened that 1st time.

KhaosT:
@devbobo That's really strange. It seems that HomeKit Accessory Simulator will return "" for empty data. Just to be safe I updated hap-nodejs to handle that case.

cflurin:
@KhaosT: I started testing the plugin 2 branch today. I was able to run the samplePlatform and the BridgeConfigurator on my iPhone 6 (iOS 9.2). 

Adapting my plugin there was a conflict with the `this.accessories = []` and the `<platform>.prototype.accessories = function(callback)`. I changed `this.accessories = []` to `this.Accessories = []` and this solved the problem.

KhaosT:
I've been running this for a while without experiencing any issue. Since this PR doesn't break the backward compatibility with existing plugin system, if there is no further feature requests, I'm planning to merge this PR on Mar 1.

nfarina:
👍

devbobo:
:+1: 

Homebridge/HAP-NodeJS are currently the most stable they have ever been....and the new plugin framework is awesome.

cflurin:
Hi, I'm ready to go: [plugin-2 branch] (https://github.com/cflurin/homebridge-punt/tree/iris)
Apart the identify function.  
```
Accessory.prototype.identify = function (callback) {...}
```
It seems the function isn't called.

KhaosT:
@cflurin should be fixed in latest commit. PlatformAccessory will emit 'identify' when it receives identify request.

cflurin:
@KhaosT: Thanks for the fix, however it didn't work for me I had to add `inherits` in homebridge/lib/plattdormAccessory.js:
```
var inherits = require('util').inherits;
var EventEmitter = require('events').EventEmitter;

...

inherits(PlatformAccessory, EventEmitter);
```

cflurin:
@KhaosT: Thanks for the fix, however it didn't work for me, I add `inherits` in homebridge/lib/plattdormAccessory.js:
```
var inherits = require('util').inherits;
var EventEmitter = require('events').EventEmitter;

...

inherits(PlatformAccessory, EventEmitter);
```
but still it doesn't work.

KhaosT:
@cflurin I forgot the ``inherits`` part. Fixed in latest commit and I also added an example to show how identify event works...

cflurin:
@KhaosT: for me it works this way:
```
  accessory.on('identify', function(paired, callback) {this.identify(paired, callback)}.bind(this));
```
 [plugin-2 branch] (https://github.com/cflurin/homebridge-punt/tree/iris)

I think that we need the callback, otherwise the identify only works once.
In your example:

```
  accessory.on('identify', function(paired, callback) {
    console.log("Identify!!!");
    callback();
  });

KhaosT:
@cflurin sorry about that, I included the changes in the update :)

KhaosT:
Alright, it's March 1 now and no one reports any sort of ship blocker so far. I think I'll just merge this PR and publish a new version then.

justme-1968:
i'm looking to change the fhem platform to use the new api to provide dynamic accessory creation at run time.

but i don't get the idea behind configureAccessory. at the moment it looks like a duplication of code. for new accessories i have to create them so there is complete code to create and initialize an accessory. 

in configureAccessory basically the code to initialize values, event handlers and internals has to be duplicated.

why can't i just recreate the same accessory on restart. there is a waring in the code that this should be avoided...

PR #513, opened 2016-02-07
new getServiceByUUIDAndSubtype(UUID, subtype) function

snowdd1:
Some platforms may have accessories that contain more than one service of a given type, such as multiple lightbulbs.


snowdd1:
Back reference: https://github.com/nfarina/homebridge/pull/497#issuecomment-181135461

snowdd1:
Cool, thanks.

PR #515, opened 2016-02-08
Update platformAccessory.js

snowdd1:
Small typo, big result.
"name" parameter does not exist, is called UUID in that function.


snowdd1:
Hold, it's still broken.

snowdd1:
Ok, it's working now. So many errors in so few lines of code. Damn, has cost me the complete evening.

PR #532, opened 2016-02-17
Handle SIGINT and SIGTERM to enable clean shutdown of Homebridge

torarnv:
For now we terminate the process, but in the future we may tell the
server to stop, which may possibly include some teardown logic.

Handling these signals also make it easier to put Homebridge inside
a docker container, as docker uses SIGTERM to tell a container process
to stop, and passes SIGINT when attached to the container and receiving
a Ctrl+C.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -23,5 +23,17 @@ module.exports = function() {
   // Initialize HAP-NodeJS with a custom persist directory
   hap.init(User.persistPath());
   
-  new Server(insecureAccess).run();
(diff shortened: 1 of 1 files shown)


KhaosT:
This is potentially useful for ``plugin-2`` to save updated platform accessories when homebridge shutdown. I'll merge this to master and bring changes to ``plugin-2``.

snowdd1:
@KhaosT very cool, would it be possible to have a events of API the plugins can register to?

PR #764, opened 2016-09-13
New bridge configs

alessiodionisi:
New optional bridge configs:
```json
"bridge": {
    "manufacturer": "Manufacturer here",
    "serialNumber": "Serial number here",
    "model": "Model here"
}
```


KhaosT:
Thanks :)

PR #779, opened 2016-09-20
Replace existing accessory info service with custom

xdissent:
HAP creates a default InformationService for each accessory that's created. If a homebridge plugin returns an InformationService for an accessory, the existing service is updated with the values from the custom service, but the plugin accessory has no way to access the existing service. This PR _replaces_ the existing service with the custom service, defaulting any missing values to the values found in the existing service. It also copies over the "paired identify" event listener that HAP creates on the existing service (and any other listeners). This allows plugin authors to retain access to the information service for later manipulation.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -390,22 +390,38 @@ Server.prototype._createAccessory = function(accessoryInstance, displayName, acc
 
     services.forEach(function(service) {
 
-      // if you returned an AccessoryInformation service, merge its values with ours
+      // if you returned an AccessoryInformation service, merge in default values
       if (service instanceof Service.AccessoryInformation) {
         var existingService = accessory.getService(Service.AccessoryInformation);
 
         // pull out any values you may have defined
+        var name = service.getCharacteristic(Characteristic.Name).value;
         var manufacturer = service.getCharacteristic(Characteristic.Manufacturer).value;
         var model = service.getCharacteristic(Characteristic.Model).value;
         var serialNumber = service.getCharacteristic(Characteristic.SerialNumber).value;
 
-        if (manufacturer) existingService.setCharacteristic(Characteristic.Manufacturer, manufacturer);
-        if (model) existingService.setCharacteristic(Characteristic.Model, model);
-        if (serialNumber) existingService.setCharacteristic(Characteristic.SerialNumber, serialNumber);
+        // pull out existing default values
+        var existingName = existingService.getCharacteristic(Characteristic.Name).value;
+        var existingManufacturer = existingService.getCharacteristic(Characteristic.Manufacturer).value;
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks for the PR but I'm not sure about merging this... Plugin-2.0 API already allows plugin to manage it's own accessory instance and at this stage we're encouraging developers move to the new API. For more details, you can check [API](https://github.com/nfarina/homebridge/blob/master/lib/api.js) or the [example plugin](https://github.com/nfarina/homebridge/blob/master/example-plugins/homebridge-samplePlatform/index.js) for how to use the new API to manage plugin owned accessory instance.

If you think it's absolutely necessary, please let me know and I'll merge it.

xdissent:
Well, it turns out this broke the wink plugin anyway, so never mind for now.

I could be totally missing something, but it looks like the API only gives you access to the plugin accessory *constructors* (`.accessory()`). Perhaps *platforms* have access to their underlying HAP accessories, but I don't see it for plugins that only register accessories.

I also noticed that in the server, plugin accessory constructors aren't being passed the api instance like it is for platforms: 
https://github.com/nfarina/homebridge/blob/master/lib/server.js#L254
https://github.com/nfarina/homebridge/blob/master/lib/server.js#L283
It's still passed to the plugin init function though, so that's not a huge deal.

My problem is demonstrated by this example:
```js
module.exports = (homebridge) => {
  const Service = homebridge.hap.Service
  const Characteristic = homebridge.hap.Characteristic

  class InfoAccessory {
    getServices () {
      // Create custom info service
      const service = new Service.AccessoryInformation()
        .setCharacteristic(Manufacturer, 'Loading...')

      setTimeout(() => {
        // no effect, because service is not added to the HAP accessory instance
        service.setCharacteristic(Manufacturer, 'Info Co')
      }, 10000)

      return [service]
    }
  }

  homebridge.registerAccessory('homebridge-info', 'Info', InfoAccessory)
}
```

Maybe a better solution is to just go ahead and create a HAP accessory instance *before* calling `getServices()` on the plugin accessory instance. Then you could pass either the entire HAP accessory instance, its existing services or just the info service as an argument to `getServices()`. I don't think that'd interfere with the default info merging that is currently happening, so it should be backwards compatible.

xdissent:
Something like this? https://github.com/nfarina/homebridge/compare/master...xdissent:pass-info-service
It does waste a little work if the services are in the old style.

b-straub:
What is the recommended solution for this issue? 

Discard registering a single accessory via registerAccessory and use the platform based solution with full control? 

If so "registerAccessory" API call should be marked as deprecated, otherwise either the proposed fix from @xdissent merged or the accessory services exposed.

PR #803, opened 2016-09-26
Use hap-nodejs 0.4.0

maddox:
This bumps the version of HAP to `0.4.0` so that Homebridge has the new IP camera support.

This is a super naive pull and hasn't  been fully tested. There may be other incompatibilities not taken into consideration, but it looks like `0.4.0` was just additions.

/cc @KhaosT 

I also only bumped Homebridge's version to `0.3.5` since nothing should actually break here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "homebridge",
   "description": "HomeKit support for the impatient",
-  "version": "0.3.4",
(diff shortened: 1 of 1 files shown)


KhaosT:
I'll merge this first and then make changes later when adding proper support for camera.

PR #805, opened 2016-09-26
IP Camera support for homebridge

KhaosT:
Enable plugin to publish IP Camera via homebridge.


KhaosT:
An example plugin can be found under at [here](https://github.com/KhaosT/homebridge-camera-isight)

PR #839, opened 2016-10-09
Fix camera name in collision error message

rxseger:
If a camera address collides, the error message includes a literal "%s":

```javascript
/usr/lib/node_modules/homebridge/lib/server.js:476
      throw new Error("Camera accessory %s experienced an address collision.", accessory.displayName);
      ^

Error: Camera accessory %s experienced an address collision.
```

The `Error` constructor doesn't support format strings: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error 

This PR changes the error message to use string concatenation, correcting the name:

```
Error: Camera accessory Camera1 experienced an address collision.
```


KhaosT:
Thanks :)

PR #917, opened 2016-11-13
Use chalk instead of hardcoded ANSI codes

crzcrz:
Improves the readability of the source code and prevents the raw ANSI codes from appearing in the log, if ran as a systemd service.


KhaosT:
Thanks :)

PR #941, opened 2016-11-21
Allow externally specified config

djMax:
I'm using this module with [confit](https://github.com/krakenjs/confit) which will allow me to split my (relatively complex) config files up, generate some dynamic configs, etc. This simple change allows me to bypass loading config.json cleanly.


KhaosT:
looks good :)

PR #1280, opened 2017-05-06
Add an entry point for Typescript support

corymayer:
This small change would make it possible to create Typescript typings for the API class, which would make writing plugins in Typescript cleaner.


nfarina:
(Months later, sorry!)

PR #1487, opened 2017-09-11
Add support for FirmwareRevision and HardwareRevision characteristics

schmittx:
Adds support for FirmwareRevision (required) and HardwareRevision
(optional) characteristics to both Homebridge itself and accessory platforms/plugins.

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -6,7 +6,9 @@
         "pin": "031-45-154",
         "manufacturer": "@nfarina",
         "model": "Homebridge",
-        "serialNumber": "0.4.20"
+        "serialNumber": "Not Specified",
+        "firmwareRevision": "0.4.27",
+        "hardwareRevision": "<optional>"
     },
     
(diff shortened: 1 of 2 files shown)


nfarina:
Looks good, thanks!

simont77:
Hi, if I understood well, FirmwareRevision has to be manually set in config.json file. This means that one have to update the config file every time homebridge updates. Would it be possible to automatically set the FirmwareRevision characteristic using actual homebridge version? Thanks

nfarina:
Good point - why would Homebridge users ever need to set these manually?

schmittx:
All valid points, hard coding and removing it from the config is easy but I don't know how to link it to the `package.json` (NPM) version. Any suggestions?

simont77:
Same for me, this is why I asked.

nfarina:
You mean like `require('./package.json').version`?

schmittx:
@nfarina Ha, yeah...sorry, I'm still new to Node...I'll update/confirm this weekend and submit a new PR

schmittx:
Check #1554, should be good to go now.

Issue #1514, opened 2017-09-24
Homebridge never adds on iOS 11


AutumnForesty:
I recently set up homebridge again on antergos, and it starts up fine.

It's adding it into the home app that's the issue.

I go and add it, and after some time, the home app says "Home couldn't connect to this accessory."
I'm on the same wifi network as the bridge as well.
What could be the issue here?

I'm running the latest iOS 11, and the latest homebridge.

erpeters3:
Exact same issue.  I was having connectivity issues, random drops.  I removed the bridge and attempted to re-add on iOS11 GM, no joy.

qudarr:
ditto for me. Used to work with IOS 11 Beta, now very unreliable connectivity.

erpeters3:
Now i cannot re-add my Homebridge even back on iOS 10.3.3.  Updated Pi Jessie, NPM, Homebridge, and Homebridge plugins to latest.  No Home apps on any iOS device see them.  Please help before my wife divorces me ;)

apetitjean:
Same issue here! My wife and I have 2 iPhones. I updated mine to iOS 11 and since then I cannot connect to homebridge anymore :-(.
My wife who's still running iOS 10.x has no problem...

erpeters3:
Following this advice from issue #1501 fixed it had to rebuild everything but st least it works

What I did which worked for me was:
	•	Upgrade homebridge to newest version: npm install -g homebridge
	•	Stop homebridge service
	•	rm -rf ~/.homebridge/persist
	•	rm -rf ~/.homebridge/accessories
	•	Under ~/.homebridge/config.json, change your "username": "11:22:33:44:55:66" to something else
	•	Start homebridge service
	•	Delete homebridge accessory from home app
	•	Add homebridge accessory to home app and enter pin etc

AutumnForesty:
@erpeters3 

Did that, no dice.

jcastro:
@erpeters3 @AutumnForesty not working here either :( but thanks!

ttscoff:
The solution from @erpeters3 worked for me. Thanks!

I still haven't figured out how to _delete_ the old items from Home in iOS 11, but one step at a time.

jg506:
FWIW, I got iOS11 issue resolved only after adding following lines to config.json which I previously  omitted. It seems like one or more of these are required by iOS 11 but not required by iOS 10.
        "manufacturer": "@nfarina",
         "model": "Homebridge",
[...]

Issue #1501, opened 2017-09-19
Not working with IOS11


theo-69:
I just updated to IOS 11 and homebridge stopped working.
So I reinstalled with sudo npm install -g --unsafe-perm homebridge, still nothing, homebridge is starting but all Items are not responding and showing with !

sendorm:
I have an iphone on ios 10.2. Homebridge is still working. Also have an ipad pro on ios 11, homebridge is not working on the ipad.

theo-69:
@sendorm same here old iPhone with 10.3 still working

dkrizic:
I have running iOS 11 on all devices and everything working, but... I have homebridge 0.4.25. And I have no problems, either directly nor via Apple TV (also running iOS 11)

theo-69:
just updated homebridge to 0.4.28, still nothing

theo-69:
using the old Version from backup again and surprise after restarting the PI its working again

reallybman:
Same issue. Working on legacy devices but not IOS 11 devices.

B1tMaster:
Upgaded just now both my ipad and iphone to IOS11. My Apple TV is still NOT upgraded to latest TvOS.
Everything works like a charm.  I am running latest version of homebridge. Did not have to change a thing.

VeniceNerd:
I am having the same issue. Updated both my iPad Pro and iPhone 7 Plus to iOS 11 today and Home shows all devices connected through Homebridge with "No Response" and the "!". Updating both my Apple TV's to iOS 11 made no difference either.

martinorob:
Sale issue here.. all devices updates (iPhone iPad appletv) and homebridge not working

martinorob:
Solved:
Deleghe homebridge from home app
Kill homebridge
Remote cachedaccessories
Change Mac address of homebridge with new one 
Remote temp folder
Start homebridge 
Readd homebridge to home app on iOS 11

matevzg:
What I see is homebridge being recognized as a thermostat and not a bridge in iOS 11. It still has all accessories attached to it in Home.app though, but it's status is No Response.

In Matthias Hochgatterer's Home app, homebridge is also recognized as thermostat with Error status.

In any way, after a few minutes of running, homebridge, all accessories become unresponsive. Restarting homebridge helps. All devices running iOS/tvOS 11.

Strange.

DoctorOgg:
tried the procedure that @martinorob has shared, and still nothing...

[...]

Issue #1350, opened 2017-06-14
ios11


jetmcquack:
I know it's definitely too early to talk about ios11 but i just wanted to report my experience so far.
I have noticed that any first accessory in config.json show up as "not Supported" and break the home configuration for every device, even those not on iOS 11.

d0n13:
I have added a new accessory this eve and it seems to be behaving ok under iOS 11. No message in during startup like you are getting.

jetmcquack:
Just for records this happens only on an iPad used as home hub

fedor1210:
I haven't experienced this. Happily. Try to delete the "accessories" folder, may be this will help (warning: it will broke some plugins, for example homebridge-mqtt. You will need to add the accessories to it again).

wild-a:
I’m running iOS 11 on my iPhone and iPad. The issue I ran into is that I’m unable to add homebridge using the Home app and the Eve app. I was able to add it using an iPhone that’s on iOS 10, and then all the accessories were working perfect.

treeiler:
@wild-a there is a manual code button at the bottom of the pairing screen in the home app...worked for me.

EdwinJS:
I have had the issue too, not sure sure of the cause but i have homebridge-cbus and had a light named "Entrance Chandelier" it was not supported when first adding the homebridge, i then stopped homebridge, renamed the light to "Entrance Chandelier 2", started homebridge, it then loaded and supported the light. Next I again stopped and renamed the light name in my config.json correctly and it remained supported and working... Not sure why it does that but it's fixed now

Hope that helped..

wild-a:
@treeiler thanks! I missed that. I was able to add homebridge by manually entering the code

KeyloggedUser:
Hey Everyone,

Just poppin in to say I too have the same issue. But I looked at what apple had to say about its new features to HomeKit and they indeed changed/improved it. Could be the connection either uses a different authentication method and/or the ports are different. iOS11 was never able to find my pi running homebridge.

Hope this is a simple fix for yuh @nfarina . Would be nice to have both awesome groundbreaking features (Homebridge and iOS11) together. Thanks for what you've put out so far! Great Work!
Have a good one!
~KLU

bdruth:
[...]

PR #1533, opened 2017-10-07
Attempt to fix not-responding accessories issue introduced with iOS11

LeoCal:
Hi,

this is an attempt to fix "not-responding" issue with iOS11. 
Basically, after the upgrade of my iPhone to iOS11, I see that sometimes the devices are seen as "not-responding" in the iPhone Home App.

The issue has been reported in many places, e.g.:
nfarina/homebridge#1514
nfarina/homebridge#1414
nfarina/homebridge#1350
nfarina/homebridge#1501

I'm not an expert of Node.js, but could it simply be that the new iOS 11 requires a more tight response time?
For this reason I've reduced the timeout to send the response to 10ms instead of 100ms.
I've tried the fix on my Raspberry PI and for now it seems to be working.
Any input on this?
I'd just like to help you guys fixing this nasty bug.

Thanks in advance

### DIFF
--- a/lib/bridgeSetupSession.js
+++ b/lib/bridgeSetupSession.js
@@ -182,7 +182,7 @@ BridgeSetupSession.prototype.sendResponse = function(response) {
     this.lastResponse = respData;
     setTimeout(function() {
       this.controlChar.setValue(respData);
-    }.bind(this), 100);
+    }.bind(this), 10);
   }
 }
 


nfarina:
I'm not really familiar with `BridgeSetupSession` and the role it plays in "normal" operation of Homebridge - do all plugin responses go through this class? I think @KhaosT will need to weigh in here - if this class isn't somehow handling all HomeKit traffic then I think your change may be a red herring.

LeoCal:
Thanks for your reply, I guess you're right.
I thought my change solved the issue, but after some time it showed up again in an intermittent way. Basically, for a certain amount of time all works fine, i.e. my iPhone6s with iOS11 can see Homebridge correctly. However, after a while it shows as "Not responding". Then back to working. I spawned homebridge in debug mode, but I really see nothing interesting coming out on the console.
I'm wondering if Apple has decided not to poll regularly the devices which are not "officially supported" like Homebridge? Is that possible?
A friend of mine has bought officially supported HomeKit devices and they work just fine apparently.
I'd really like to help troubleshooting this issue and converge to a fix, so if you have a suggestion about how to move forward, I'm more than glad to help.

Cheers,
Leo

NorthernMan54:
Leo,

In a not responding issue, what you may need to look for is something missing from the debug log, rather than an error.

LeoCal:
@NorthernMan54 
True, that's a very good point. Do we have an example of debugs supposed to be seen when everything works fine?
Then we can compare with the failing case and see what is missing - then try to infer why that code path is not getting triggered.

KhaosT:
@nfarina No, that part of the function is only executed when user try to modify homebridge configuration using the configuration handler (which basically no one is doing that right now).

For all iOS 11 related issues, it's mostly caused by user running outdated homebridge/hap-nodejs. In iOS 11, HomeKit changed its sync mechanism, and some objects that previously doesn't have a UUID now have one. However, there is a bug in iOS 11 which HomeKit can generate same UUID for a service and a bridged accessory which resulted in undefined behavior. From user's aspect, it can be accessory not response, accessory not supported, accessory being blocked (Home app shows as not response). For most of situations, the issue can be addressed by update homebridge (and associated hap-nodejs) to latest version. In rare situation where HomeKit stuck in a state that is not recoverable, user may need to remove homebridge, clean up persist folder and pair with homebridge again to fix the issue.

LeoCal:
@KhaosT 
Thanks a lot for you response. Actually, I've already updated to the latest version of homebridge and I see the issue there:
$ homebridge -v
0.4.28

Could it be that the fix you did for iOS11 for HAP-NodeJS has not been ported to homebridge? I don't recall having installed HAP-NodeJS, but just homebridge, which is base on HAP AFAIR.

Thanks

KhaosT:
@LeoCal hap-nodejs is a dependency of homebridge. Please try to run `npm list -g` and see what's the hap-nodejs's version is on your device.

LeoCal:
@KhaosT I have the following version installed:
npm list -g | grep -i hap
│ ├─┬ hap-nodejs@0.4.33

I'll try to upgrade to the latest available version and will let you know if that solves the issue

Thanks,
Leo

KhaosT:
That’s already the latest. In that case I think your situation is different. What kind of plugins do you have on this homebridge instance?

LeoCal:
@KhaosTJust one to get it integrated with my OpenRemote:
homebridge-openremote

Sent from my iPhone

> Il giorno 12 ott 2017, alle ore 21:38, Khaos Tian <notifications@github.com> ha scritto:
> 
> That’s already the latest. In that case I think your situation is different. What kind of plugins do you have on this homebridge instance?
> 
> Sent from my iPhone
>

bobfatherx:
I'm running latest of homebridge using the homebridge-SmartThings plugin and I'm seeing the same behavior. Homebridge works for a little while, then things go "not responding". 

This behavior started happening out of the blue about 3 months ago, and I saw it on iOS 10 and now on iOS 11 also.

KhaosT:
@LeoCal I'm looking at the code for that plugin, maybe the http request didn't come back before some other event happens? Can you provide logs for when it's working normal and when Home app says "no response"?

LeoCal:
@KhaosT 
In the end I managed to get everything working again (hopefully).
I basically removed all the packages (nodejs, npm, node-server, nodered, hap-nodejs, homebridge) and installed everything from scratch.
It's already ~1/2 hour it's up and running and I'm not seeing issues anymore.
I'll update the web pages mentioned above to let everyone having the same issue know about this.
Thanks @KhaosT and @nfarina for your support on this - you guys rock!
I'll be glad to contribute to your projects: first thing I have in mind is to write an accessory for controlling Blinds.. if I manage to get that done, I'll raise a pull request.

nfarina:
Glad you got it working!

poldy79:
I had the same problem for a couple of weeks and also deinstalled and reinstalled everything with npm. Seems to help. Lets see if it persists...

tleo86:
i have same problem
iOS 11.4
homebridge 0.4.43 a
nd my accessories doesn't response =\ 
how i can fix it?

Issue #1039, opened 2017-01-04
homebridge 0.5


KhaosT:
It has been a while since the major changes happened with homebridge. I bet Nick didn't think homebridge community would grow this fast when he first started it (I certainly didn't 😆).

Going through existing issues, it looks like people are having a lot problems with initial setup, unexpected plugin crashes and HomeKit configuration disappear (#874). To address those problems and improve the overall experience, I'm proposing the following changes for the next major release of homebridge. This is a very rough draft of the plan, feel free to comment and add new things to the list. I'm relatively free over the next few months so hopefully I'll be able to implement most of these.

+ Plugin OOP
	- Most of existing plugins will not be affected since API should remains the same
	- Automatically reload when plugin crashes
+ Mandatory accessory caching
	- Optionally opt out?
+ Web Interface
	- Manage cached accessories
		- Remove them
		- Control them?
	- Manage plugins
		- Config plugins
			- Interactive based on existing configuration service?
			- Field based?
		- Install plugins?
+ Dynamically splitting accessories across multiple bridge instances
+ Plugin certification
	- May just be "recommended plugin" in the plugin catalog instead of mandatory requirement for homebridge to load the plugin
	- Code review for plugins
+ New plugin types
	- Observer (Read only)
		- Have homebridge poll endpoints with interval or go with notification based solution?
		- Data logger
	- Manager (Read write)
		- Automation
		- Plugin management

---
Nice to have

+ ES6 rewrite
+ Mac Wrapper
	- A macOS app that can run directly with bundled nodejs env
+ Raspberry Pi setup script
	- For Raspbian

---
PS: Anyone interested in writing a proper web interface?

gismo141:
@KhaosT: see my plugin [homebridge-server](https://github.com/gismo141/homebridge-server).
At the moment it supports only bare-essentials but it works for a lot of different setups already.

gismo141:
@KhaosT: see [homebridge-raspbian-installer](https://github.com/macnow/homebridge-raspbian-installer) for the installation and basic setup of a running homebridge-environment.

At the moment I'm trying to implement additional functionality and flexibility.

KhaosT:
@gismo141 Yeah, I saw ``homebridge-server`` before :) It's really nice you're doing that. I didn't know ``homebridge-raspbian-installer`` exists before, guess that's a done for the last task.

misanthrophy:
[...]

PR #1541, opened 2017-10-11
Check plugin's engines.node version

plasticrake:
I'm currently doing this in [homebridge-hs100](https://github.com/plasticrake/homebridge-hs100/blob/6ffac57083865e55145b3821ec37d6b0023bf734/lib/platform.js#L39) and I got the idea from [ebaauw/homebridge-hue](https://github.com/ebaauw/homebridge-hue/blob/b74732615bdeae08c8e2c729375ffbe80d09eec7/lib/HuePlatform.js#L511). Instead of each plugin checking the node version have Homebridge do it same as it checks `engines.homebridge` already. The check only runs if `engines.node` is specified in the package.json. If node is missing the check won't happen (so it won't break any plugins that do not specify a node version)

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -56,6 +56,13 @@ Plugin.prototype.load = function(options) {
     throw new Error("Plugin " + this.pluginPath + " requires a HomeBridge version of " + versionRequired + " which does not satisfy the current HomeBridge version of " + version + ". You may need to upgrade your installation of HomeBridge.");
   }
 
+  var nodeVersionRequired = pjson.engines.node;
+
+  // make sure the version is satisfied by the currently running version of Node
+  if (nodeVersionRequired && !semver.satisfies(process.version, nodeVersionRequired)) {
+    throw new Error("Plugin " + this.pluginPath + " requires Node version of " + nodeVersionRequired + " which does not satisfy the current Node version of " + process.version + ". You may need to upgrade your installation of Node.");
+  }
+
   // figure out the main module - index.js unless otherwise specified
   var main = pjson.main || "./index.js";
 


nfarina:
LGTM, thanks!

oznu:
This is a breaking change and prevents the `homebridge-hue` plugin from loading on later versions of node.

```
[2017-10-12 22:43:16] ====================
[2017-10-12 22:43:16] ERROR LOADING PLUGIN homebridge-hue:
[2017-10-12 22:43:16] Error: Plugin /homebridge/node_modules/homebridge-hue requires Node version of 6.11.3 which does not satisfy the current Node version of v8.7.0. You may need to upgrade your installation of Node.
```

There could be other existing plugins that have implemented the node engine check incorrectly so I would recommend making this just a warning rather than throwing an error, at least for now.

ebaauw:
I'm having mixed feelings about this change.  I'm happy for homebridge to do the check, but I don't think it should result in an error.  I deliberately issue a warning message, when homebridge-hue runs on a different version of NodeJS than the version I develop and test against.  I spent too much time analysing issues brought by my users, only to find it's yet another breaking change in NodeJS v8.

plasticrake:
@ebaauw I actually got the idea of doing this check from your repo. I implemented it in a similar way (but using semvar instead of hardcoded to a version). My plugin uses syntax that requires node 6, but even with this warning about half of all the issues raised on my plugin's repo were people running node 4 and either ignored or didn't notice that warning message. 

If this is changed to a warning in homebridge then I'm going to leave the check in my plugin and have it fail with an error. (which I am fine with if that is decided)

Alternatively `homebridge-hue` could change its node engine version to  >6 (or whatever the minimum version is, and fail if under) and also have its own check for the recommended version and have a warning if someone is using node 8, etc.

Just some thoughts! I'm not sure how many plugins actually have a node engine version defined in their package.json

ebaauw:
Thanks, @plasticrake,

> Alternatively homebridge-hue could change its node engine version to >6 (or whatever the minimum version is, and fail if under) and also have its own check for the recommended version and have a warning if someone is using node 8, etc.

For now, I've changed my own check in homebridge-hue (and in homebridge-zp), based on `engines.nodejs` instead of `engines.node` in package.json.  And I moved to semvar as well (which had been on my todo list for quite some time...).

> My plugin uses syntax that requires node 6, but even with this warning about half of all the issues raised on my plugin's repo were people running node 4 and either ignored or didn't notice that warning message.

Indeed, too much time is spent on issues with dependencies.  And indeed, all the READMEs, Wikis and warning (or even error) messages are gracefully ignored.  I will be looking at creating a template for issues, hopefully reminding people to look at these first.

> If this is changed to a warning in homebridge then I'm going to leave the check in my plugin and have it fail with an error.

I think this is a difficult one.  There's bound to be conflicting dependencies between plugins, e.g. I develop and test against node v6 LTS, where others use node v8.  When I would raise an error for node v8, other plugins depending on node v8 won't work in the same homebridge instance.  I really think we should find a way to distribute homebridge, a single plugin and their dependencies in a container, or be way more strict in certifying plugins (see #1039).

nfarina:
Hey guys, I didn't realize this was a breaking change - reverting now

PR #1544, opened 2017-10-12
Add Support for Alexa to control compatible devices

NorthernMan54:
This pull request, in conjunction with HAP-NodeJS pull request 512 adds support for Alexa to control compatible devices.  At this time only LightBulb, Outlet, Switch, and Fan devices due to the limitations of the API I'm using from Alexa to Homebridge.

To enable the feature, a minor change must be made to your config.json

add a new setting "ssdp" to the bridge section of your homebridge config.json file. Value must be 1900. i.e

 "bridge": {
    "name": "Howard",
    "username": "CC:22:3D:E3:CE:31",
    "port": 51826,
    "pin": "031-45-154",
    "ssdp": 1900
},

After this is successfully rolled out, I will send a further pull request documenting the config change in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,9 @@ Once you've installed a Plugin or two, you can run Homebridge again:
 
 However, Homebridge won't do anything until you've created a `config.json` file containing your accessories and/or platforms. You can start by copying and modifying the included `config-sample.json` file which includes declarations for some example accessories and platforms. Each Plugin will have its own expected configuration; the documentation for Plugins should give you some real-world examples for that plugin.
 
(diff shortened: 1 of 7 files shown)


nfarina:
This sounds cool but I don't think we want to pin Homebridge's version of HAP-NodeJS to your forked repo - it should be using the latest official version of HAP-NodeJS. I would suggest opening a PR at HAP-NodeJS for feedback there first, or bundling your functionality into a separate plugin.

NorthernMan54:
I was talking back and forth with Tian last night and I’m going look for an approach that  doesn’t require adding non HAP api’s to HAP-Node, and likely just a plugin as you suggested.  I’m going to close this later

PR #1552, opened 2017-10-13
Check plugin's engines.node version and warn if not met

plasticrake:
Same as last time, but only warn so won't cause issues with plugins


KhaosT:
Thanks :)

PR #1554, opened 2017-10-14
Populate accessory information for Homebridge from NPM package

schmittx:
- Populate accessory information from `package.json`
- Remove need for user to manually specify

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -3,12 +3,7 @@
         "name": "Homebridge",
         "username": "CC:22:3D:E3:CE:30",
         "port": 51826,
-        "pin": "031-45-154",
-        "manufacturer": "@nfarina",
-        "model": "Homebridge",
-        "serialNumber": "Not Specified",
-        "firmwareRevision": "0.4.27",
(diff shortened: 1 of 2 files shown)


ebaauw:
I think SerialNumber should be mapped to bridgeConfig.username (the “mac address” of the homebridge server), instead of to bridgeConfig.pin?

schmittx:
I was debating that too, let’s see what @nfarina wants to do and I can update this PR if necessary.

nfarina:
@ebaauw That sounds more correct to me - in the rare care of multiple Homebridge instances running, you would probably pick the same PIN for both, but would have to pick different MAC addresses no matter what.

schmittx:
@nfarina @ebaauw Sounds good to me, I just pushed another commit to update, should be good to go now.

nfarina:
Cool - @KhaosT any reason not to make these values automatic?

KhaosT:
@nfarina Not really, automatically update those values are nice :)

nfarina:
Cool, thanks @schmittx!

PR #1570, opened 2017-10-20
Use proper accessory name in debug log if more than one accessory.

mkloock:
Minor issue but was annoying while debugging a plugin with multiple cameras. Only the last camera name was listed in the the CLI output.


KhaosT:
Thanks, I always meant to fix that but keep forget to do that 😅

mkloock:
Minor - but I was new and it helped me reading the code.

PR #1577, opened 2017-10-27
Adding info on the sample plugin generator

bhagyas:
Adding information on the yeoman based plugin generator.


donavanbecker:
@bhagyas is this something you are still support? if so maybe it would be better to add this into the Wiki under [Writing Plugins](https://github.com/homebridge/homebridge/wiki/Writing-Plugins) Section.

PR #1675, opened 2017-12-23
(Word file) Instructions to Install on Windows 10 64 bit

jvmahon:
Instructions to install on Windows 10 64 bit.
File is in Word 2016.


nicklayb:
A Markdown file would be better, for compatibility and those who don’t have Word.

jvmahon:
I understand that if I convert to Markdown, it loses all the screen shots so it may be less helpful to do that. How about a shorter introduction in Markdown explaining what is in the attached fuller-explanation file, and then to address the Word concern you raised, I add the file but in "rtf" format which is compatible with the Windows built-in WordPad or else as a PDF. It could also be posted on Google Docs or something like that.  Let me know your thoughts (or if there is some  automated way to convert to Markdown but include images, let me know - I'm not that familiar with the Markdown format). But before doing any of this, I was planning to add instructions on how to auto-start Homebridge at boot using Windows Task Scheduler and how to install using OpenSSL version 1.1 or greater (which currently causes a linking error during installs of the ed25519 module used by hap-nodejs) - that may delay this by a week or so.

nicklayb:
Even though some "converters" may work nicely, I recommend writing it manually in Markdown. It's simple and easy to do. When you get used to it, it gets easier that any rich text writer (I even did most of my school work with it). The purpose of md is that any text editor (Atom, VS Code even notepad) can read and you can easily understand it's meaning through the formatting and GitHub will be able to display the markdown formatted in the browser. A `docx` cannot be opened it a text editor (except if you have some plugin), an rtf can but the formatting is a pain.  Since Markdown is the standard for open source coding documentation, I would personally decline the PR until it is md written.

Screenshot can be easily added to a markdown, the only is that your screenshots must be files in your PR. Place them in a folder like `/docs` and name them, for instance `windows_screenshot_1.png`. And then the following format will display it : 
```
![This is how to install on Windows](/docs/windows_screenshot_1.png)
``` 

You can either write it with a text editor with markdown plugin or with a web editor such as [https://dillinger.io/](https://dillinger.io/). Feel free if you have any questions, i'd be glad to help you.

Starbix:
I've had a go at it here: https://hackmd.io/EYRgLAZgDAJjUFoDMB2ArAJgWAnCgHAjgKYwQIBsGSOuSFOMIAxkA===

jvmahon:
Starbix - Thanks. That looks good. I want to polish up Section II  on Setting up auto-start on Windows by pasting in some more screen grabs. I expect to get to that next weekend (I'm currently traveling). Once I finish it, I'll let you know and perhaps you can help by converting it again.

Hackashaq666:
Thank you for the writeup, very helpful!
I have been running 32bit versions of these softwares on 64 bit windows 10 for about a year and was inspired to move to 64 bit versions a couple of months ago after your initial post.
I think the following steps are not needed:
  B: Install Visual Studio Community
  C: Install Python 2.7.14
They are redundant to:
  G: Install Windows Build Tools

I have installed without B and C, after a fresh install of Windows 10 64 bit on my old desktop.
Additionally, he following homebridge addons/accessories are working well: samsungtv2016, TCC, harmonyhub, homeassistant

jvmahon:
Thanks for the feedback.

I’ll re-check on the “B” and “C” steps.

pdlove:
I finally had time to review.
I second what @Hackashaq666 said. Those steps are redundant and cause issues more often than not. the Windows Build Tools is an excellent package that eliminates alot of "I hope this is the right version".
I have two additions...
On Step 7, you might add "Or Command Line (Admin)" to make it clear for slightly older versions of Windows 10.

You might consider adding a final step about where to create the .homebridge folder and put the configuration.

Other than those items, this is a great guide!

KhaosT:
Is there a particular reason that instead of updating [Wiki Article](https://github.com/nfarina/homebridge/wiki/Install-Homebridge-on-Windows), we are having a separate document in the repo for this?

jvmahon:
@pdlove -  I"m updating the guide to cover these points. The windows build tool doesn't install the 64-bit version of Python which is what led me down the path of separately installing it and Visual Studio 2017. However, I've also made a PR to the windows-build-tools repository on the Python issue and hopefully that will be updated soon to  install 64 bit Python on 64 bit windows which makes the separate install unnecessary. I hope to update and re-test the instructions once that happens.

@KhaosT - The Wiki article likely makes more sense as you've suggested. But as background, the guide started out as notes to myself which I kept in Word of the steps I took to make this all work on Windows 10 64-bit. I then augmented with screen shots thinking it was useful to others. Rather than re-edit for the Wiki format, I simply posted what I had. I'm adding further edits to the guide to instruction users on how to set this all up to auto-start on boot and once that is done, it may make sense to transform this to the Wiki format. However, lack of spare time and lack of experience with the Wiki format (particularly with respect to the formatting of pictures of inclusion in that format) made me post as Word as a short-term solution.

PR #1712, opened 2018-01-21
Expose the information service

b-straub:
Expose the information service, this will allow deferred update of characteristics for a standalone accessory that has no platform.

Currently there is no way to deferred update the characteristics of the automatically generated Information Service. Information such as firmware revision and serial number are often retrieved by asynchronous request from an accessory device and should be set when available.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -433,6 +433,10 @@ Server.prototype._createAccessory = function(accessoryInstance, displayName, acc
       }
     });
 
+    // expose the information service, this will allow deferred update of characteristics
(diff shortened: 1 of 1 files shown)


KhaosT:
HomeKit doesn’t expect those to change in the runtime. I believe spec explicitly says those can only change during a firmware update, strange things can happen otherwise.

b-straub:
The change is not intended for runtime changes but deferred setting when the accessory is connected and ready to use, however I understand you concerns what likely means to defer the registering of the accessory until all information characteristics are known.

PR #1716, opened 2018-01-25
use correct var name in sample

chrisjshull:


facastagnini:
This should be merged

PR #1734, opened 2018-02-09
fix: Accessory is name newAccessory

Krillle:
Example crashes with:
/home/pi/homebridge-xxx/index.js:196
    platform.log(accessory.displayName, "Light -> " + value);

ReferenceError: accessory is not defined

----------
Thanks for providing Homebridge!


facastagnini:
dup of #1716

PR #1741, opened 2018-02-13
Update and rename config-sample.json to config.json

Shana611:


justme-1968:
i think the sample file should not have the same name as the regular config file. the risk of overwriting by accident is to high.

dark3nergy:
on Windows 10 i install homebridge fine, however whenever I edit the config file I get errors even after running them a JSON validator

Shana611:
Hi, I need your help. I have Somfy Blinds & curtains connect with TaHoma hub but I can’t work with Siri. I need to add Somfy Blinds & Curtains to Home-bridge. I already have running Home-bridge with Nest Thermostat. Please can you send me full details to connect Somfy to Apple HomeKit. Thanks

PR #1842, opened 2018-05-06
Changes to support @homebridge

NorthernMan54:


NorthernMan54:
No need to merge yet, I'm going to do some testing and validation first.

NorthernMan54:
@KhaosT @nfarina I have completed my testing of this, and it is ready to go, and was looking for a second opinion on the change before committing it.  The changes to plugin.js are to support two major changes needed to support the NPM org Homebridge.  First all plugin names need to have @homebridge added to the front, and second when the plugins are installed they will reside under the directory @homebridge.

iRayanKhan:
Suggestion: Not having to always delete persist and accessories folder when host reboots

AirBorne04:
Hi,

I guess this change is only applying to the official plugins, right? I was thinking if it wouldn't make sense to generally forget about the naming convention (for community packages) and go with the keywords tagging only?

NorthernMan54:
The thought process was that once we have the homebridge org fully setup, we move supporting the moving of supported plugins to the org, provided they met some basic standards.  Eventually this would allow end users to quickly identify plugins that are supported, and maintained.

AirBorne04:
That sounds generally like a good thing, just wondering what maintained is gonna look like, since i guess you can't have all the devices and do debugging, so the control is still with the community authors? Or will you create a package and hand over admin rights of that package to the maintainer? In that case i want to register for a harmony hub package 🥇

NorthernMan54:
As this is a hobby for people, we were looking at having the original developer maintain their plugin once it is in the org.  By having maintained plugins in a org, the governance team would have the ability to manage plugins that are no longer maintained.  Rather than the current situation where they are abandoned, and become a nuisance for people when searching for plugins.

AirBorne04:
sounds like a good plan looking forward getting that live 👍

bluephlame:
Can I suggest including the option for non-homebridge org scoped plugins?   I am having this issue but have modified my own see #2263 .

could the Plugin.installed = function() look in any '@' directory for homebridge plugins?

hassankhan:
Definitely agree with @bluephlame, should not be limiting plugins by scope. A simpler requirement would be for packages to specify `homebridge-plugin` as a keyword (as most already do).

inikolaev:
What's the progress with this? Would really love to see scopes support, so that I could publish plugins under my own scope and Homebridge should still be able to discover them.

Issue #1864, opened 2018-05-21
New option for Homebridge_OPTS


Nastras:
Hello everybody,
i have 24 Homebridge instances running as systemd service, one instance for each plugin.

What has always disturbed me and synonymous has reduced the system speed is that every instance starts all the plugins from global installed / usr / lib / node_modules.
By a note from another user, the solution became aware of the -P option to use. To test I have installed all the plugins local what a little more cumbersome the conventional way and started with the -P option.

Now each instance just launches the plugin it needs.
If I install the plugins globally in / usr / lib / node_modules I have a random access memory consumption when all 24 instances are started from 2.5GB.
With an installation of the plugins locally is that only each instance the plugin loads what it needs at 700 MB. Here is a huge performance difference especially for raspberry pi user.

My own hardware is an Intel nuc, because with the raspberry pi 3 the random access memory was no longer sufficient. On my raspberry pi 3 I could start a maximum of 10 instances. I think that many users use their Homebridge with multiple instances because the advantages are very big compared to the single instance.

I have also tried using the -P option to start the instances when the plugins are installed in / usr / lib / node_modules. Unfortunately all plugins will be loaded in every instance. I also received this message:

`skipping plugin found at '/ usr / lib / node_modules / homebridge-sun-position' since we already loaded the same plugin from '/ usr / lib / node_modules / homebridge-sun-position'.
`

Would not it be useful here to create maybe a new option -X for the (config option) in homebridge which still allows the plugins globally to be installed under / usr / lib / node_modules. This would allow for each instance only the plugin to start what is needed without reinstalling all plugins to local.

For the ever larger Homebridge setups of the users I think that it would be a great new feature.

Greeting Nastra

Nastras:
#1855

Nastras:
#1855

theo-69:
Great Idea

m4d-maNu:
Good idea, please intrigue in HB.

SeydX:
Very nice idea, i have submitted a pull request to make that possible

https://github.com/nfarina/homebridge/pull/1866

Heiliger85:
Great idea. I hope it will come in next version of Homebridge.

NSCHI:
Great idea

oznu:
[...]

PR #1866, opened 2018-05-22
New Option "-X" for multiple instances

seydx:
A new option has been added (**-X**) that allows only certain plugins to be loaded to avoid huge memory usage

**Example** (/etc/default/homebridge-instance)

```
# Defaults / Configuration options for homebridge
# The following settings tells homebridge where to find the config.json file and where to persist the data (i.e. pairing and others)
HOMEBRIDGE_OPTS=-U /var/homebridge-instance -X homebridge-alexa

# If you uncomment the following line, homebridge will log more 
# You can display this via systemd's journalctl: journalctl -f -u homebridge
# DEBUG=*
```
_(i.e. this will only load homebridge-alexa)_

Adding the plugin name after -X allows only to load the given plugin, you can also list several plugins (you just have to separate it with a comma, like **-X homebridge-alexa,homebridge-hue** etc.)

Another advantage is that your homebridge instances will load noticeably faster, feels three times faster than without "-X" AND in my case I have 30% less memory usage than without "-X"

The plugins are loaded "globally", so it will be always possible with "-g" to continue installing or updating plugins

The -X option just insert the added plugin name(s) to the already existing paths, so it works on every system

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -140,9 +140,15 @@ Plugin.addPluginPath = function(pluginPath) {
   Plugin.paths.unshift(path.resolve(process.cwd(), pluginPath));
 }
 
-Plugin.setDefaultPaths = function(customPath) {
-  log.info("Custom path detected! Changing plugin location to " + customPath)
-  Plugin.paths = [customPath];
+Plugin.setPluginPath = function(p) {
+  let plugins = p.split(',')
+  const pluginPaths = [];
+  for(const i in Plugin.paths){
+    for(const j in plugins){
+      pluginPaths.push(Plugin.paths[i]+'/'+plugins[j])
+    }
+  }
+  Plugin.paths = pluginPaths;
 }
 
 // Gets all plugins installed on the local system


m4d-maNu:
Great

Nastras:
Yaehhh man,

Thank you very much, that you've made this great feature so fast. Please add this Option!

Nastras:
#1864

maikburd:
I completely agree! Please add this Feature to Homebridge!

Heiliger85:
Great idea. I hope it will come in next version of Homebridge.

theo-69:
Really Great!

det-box:
Hello,
I think the Ide is great especially for multible instances.

Please take it to the HB

Thank you

Kohle81:
Great idea, please add this to the homebridge! 👍

theo-69:
is there a way to install it already?

theo-69:
@nfarina a merge would be great, waiting for this future to to problems with one plugin.

SeydX:
@KhaosT 

Is the method with passing paths into config.json not "too late" for homebridge? 
The config.json is only loaded after all plugins of homebridge have been loaded or am I wrong?

An argument passed in on launch prevent homebridge to load any other plugins which is listed after -X

theo-69:
@KhaosT I think if someone can handle several instances he can handle -X ;-)

Nastras:
@KhaosT Can you tell us what your next plan is for this pull request? Will he be taken over?
Or should @Sydex rebuild it for the config.json? What about his guess that the config.json loads too late?

Many Thanks!

Greeting Nastra

NSCHI:
like this idea, need it in homebridge.

Razeeer:
I would also be very happy about this feature in the official Homebridge. Please take it.

theo-69:
@KhaosT so why still no pull

KhaosT:
I still think this should live in config file instead of a launch parameter. Regarding the order of operation, I mean we can change that if necessary. I’ll try to write something later today and see how it goes.

SeydX:
that sounds good, but as i said, to get the advantages i listed above, homebridge should check first which plugins we only want to load before loading all installed plugins

ebaauw:
I would prefer a config.json solution as well, over a command-line option.  This would make for simpler system management, as all plugins would still be installed in one place, and only the configuration needs to be changed between instances.

Rather than introducing a new key, would it be possible to use the existing `platform` and `accessory` keys?  Homebridge would derive the plugin directory from config.json, rather than through `registerPlatform()` or `registerAccessory()`.  I’m not sure I oversee the full consequences of this, but at least:
- In config.json, one would specify the plugin suffix rather than the plugin name, e.g. `”platform”: “xxx”` for plugin `homebridge-xxx`;
- The name parameter to `registerPlatform()` and `registerAccessory()` would become obsolete;
- `cachedAccessories` would need to hold the platform directory suffix rather than the platform name.

cubilon:
I agree with @ebaauw. a config.json solution would be a better choice.

theo-69:
Even with -x solution the plugins are still installed in the same place. Its just a edition to the already existing way to load plugins. For now I think it would be the best solution. Maybe for later it should be on the to do list.

r32er:
Great idea! i hope it will be added to the homebridge!

Nastras:
That is really a shame. I thought about the last answer from @KhaosT

>   I'll try to write something later today and see how it goes.

The now speed in the topic comes but unfortunately it does not seem to be the case.

Maybe we have to ask the boss .-)
@nfarina @KhaosT 
What is the current state of affairs on the topic?

PR #1911, opened 2018-06-25
Option to hide QRcode in logs

NebzHB:


KhaosT:
Given this thing only shown once, is it really necessary? The logger one make sense as it affects all logs...

NebzHB:
When homebridge is inserted into a graphical interface, the special chars do not display correctly, something like : [0m[40m  [0m[40m  [0m[47m  [0m[40m
that makes the logs less readable.

KhaosT:
I understood that, but given this only happens at the beginning of the log, does it really matter?

NebzHB:
as the accessory creation log is at the beginning of the log, it is most of case

KhaosT:
Alright ;) In that case, would you mind to consolidate `insecureAccess` and this together as `opts` that get passed to server?
Thanks 🙏

NebzHB:
sure

NebzHB:
done

NebzHB:
the sentence with "Or enter this code ..." is less good now...

NebzHB:
fixed

KhaosT:
Great thank you :)

PR #1912, opened 2018-06-25
bugfix logical

NebzHB:
my bad....


KhaosT:
😅How about we change this to hide QR code and have logic match the allow insecure access? I'm not really a big fan of checking `undefined` like this...

NebzHB:
your command ;)

PR #1958, opened 2018-08-04
Handle both CommonJS and ES6 style module.exports.

jcoleman:
In ES6/Babel 6 doing `export default function() {}` results in that
function being assigned to `module.exports.default` rather than to
`module.exports` directly as it does in CommonJS.

Currently the Homebridge platform plugin setup expects the initializer
function to be available from `require` in CommonJS style, and so it
breaks for projects using ES6. Additionally there is currently no
helpful error when this happens; Homebridge fails to boot with Node
reporting a JS type error that `plugin.initializer` is not a function.

Here I've retained backwards compatibility with CommonJS style, but
check to see if `module.exports` is a function itself or if we need to
access `module.exports.default`. If neither is a function, we throw a
helpful error.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -69,7 +69,14 @@ Plugin.prototype.load = function(options) {
   var mainPath = path.join(this.pluginPath, main);
 
   // try to require() it and grab the exported initialization hook
-  this.initializer = require(mainPath);
+  var pluginModules = require(mainPath);
+  if (typeof pluginModules === "function") {
+    this.initializer = pluginModules;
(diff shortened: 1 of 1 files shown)


jcoleman:
@nfarina Any chance we could get this merged?

ouster:
not sure anyone is home anymore..

nfarina:
Sorry - I am out of town this week and haven’t had time to open the laptop. I will look at this soon. Thank you!

nfarina:
This is great - thank you for your patience! Merging.

pradeepmouli:
yay

PR #2074, opened 2018-11-26
snap package support

kubiko:
upstreaming snap package support to enable builds from upstream


kubiko:
Forgot to add stats about current install base:
https://snapcraft.io/homebridge
When installed as snap, HOMEbridge auto updates without any user interaction. And it's one liner to install it

NorthernMan54:
After watching all the challenges the big players have with updates, are we confident this is a great idea?  i.e. Recent Microsoft updates?

kubiko:
it only adds another option to deploy HOMEbridge does not affect any existing deployments.
I'm personally running HOMEbridge as snap for past year or so. What snap give you is not only unattended automatic updates. When new version is installed, previous two versions are kept around for rollback. So in case new version fails to start its service, system will automatically rollback to previous revision which has been kept around for this very purpose. This very system is used for industrial gateways (e.g. Dell edge gateways) so pretty robust.
As I'm running my house on it, I like the idea HOMEbridge running in own sandbox protected from rest of the system.
From stats about installed versions, when new version is released I can see  ~90% of active users being updated within a day. 
In my experience with Wekan, once snap option was enabled, most of the installs are done as snap, at least on linux systems. It's one line, as all dependencies are bundled in...

mike4president:
I have one question: when I install Homebridge via snap. Is the config UI installed as well?

PR #2107, opened 2019-01-09
Update ISSUE_TEMPLATE.md

NorthernMan54:
@nfarina @ebaauw 

Any thoughts on this wording change?  Am thinking to try and funnel people with issues to the Troubleshooting page on the wiki first, then an issue afterwards.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,9 @@
 <!-- If you have an issue with a plugin consider creating an issue in the plugin's github repo instead. -->
+### Pre Steps
+
+Before opening an issue, please review the Troubleshooting Page on the Wiki to ensure that this is a new issue, and alternatively search the closed issues for similar problems.
+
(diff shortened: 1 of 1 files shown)


ebaauw:
Wording looks fine to me.  You're going to see  a lot of issues with this text, if you don't enclose it in a comment.

I created an issue template for homebridge-hue, but I find hardly any-one uses it.  I find people hardly read the README, nor the WiKi, let alone search closed issues.  Although sometimes I get an additional question to a closed issue.  I fantasised doing an automatic reply to new issues, asking for the info requested from the template if it's not there.  I feel there should be different forums for end-user support vs (plugin) developer issues, but I don't have a clue how to implement that (Slack vs GitHub?), let alone how to enforce that.

NorthernMan54:
@ebaauw Thanks for the feedback on using the comment block. ;-)

I agree with you on directing end-user support to a different forum, I had thought about using reddit for that, as it is public and fully searchable.  With the added bonus of not requiring any effort to run.

With Slack, I find it is very good for the problem of the moment, but not very good for searches or history.  And with Github, I wasn't aware that they had anything besides issues for managing these type of inquiries.

PR #2164, opened 2019-02-25
Resolve custom config path so plugins extracting it can work with it

afharo:
If running `homebridge -U ./`, the plugin `homebridge-config-ui-x` doesn't work because it cannot find the config.json file.

Running it with an absolute path does work instead. So I suggest converting the relative path into an absolute one so when plugins request the config path, they'll get the absolute version of it.

### DIFF
--- a/lib/user.js
+++ b/lib/user.js
@@ -43,5 +43,5 @@ User.cachedAccessoryPath = function() {
 }
 
 User.setStoragePath = function(path) {
-  customStoragePath = path;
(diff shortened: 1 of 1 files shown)


afharo:
@nfarina Can this PR be merged?
Or is there any reason why it shouldn't be?

nfarina:
So sorry it took me forever to get to this! I don't see any reason why this shouldn't be merged.

Issue #697, opened 2016-07-04
Merging of Service.AccessoryInformation breaks later updates to it's characteristics


FrodoKenny:
In Server.prototype._createAccessory a potentially supplied AccessoryInformation from the getServices of an Accessory is merged with the default one from the constructor. Because the reference from the constructor is kept instead of the one from getServices, it is not possible to update the Characteristic.Model etc. with a setValue after a callback, using the reference from getServices.

Can it be changed to keep the reference from getServices and merge the default fields into that one?

PR #2169, opened 2019-03-03
Remember listeners on AccessoryInformation if returned by getServices

mrose17:
if `getServices` returns a Service.AccessoryInformation instance, the  existing code copies the `value` but not the `listeners`. In some cases (e.g., in https://github.com/homespun/homebridge-accessory-apcupsd/blob/master/index.js#L344), it isn't "convenient" to retrieve model and serial number information synchronously (cf., https://github.com/nfarina/homebridge/issues/697) .

The proposed code copies any listeners in the returned Service.AccessoryInformation instance so as to defer binding of the value until "convenient"

Enjoy!

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -437,26 +437,39 @@ Server.prototype._createAccessory = function(accessoryInstance, displayName, acc
       if (service instanceof Service.AccessoryInformation) {
         var existingService = accessory.getService(Service.AccessoryInformation);
 
-        // pull out any values you may have defined
-        var manufacturer = service.getCharacteristic(Characteristic.Manufacturer).value;
-        var model = service.getCharacteristic(Characteristic.Model).value;
-        var serialNumber = service.getCharacteristic(Characteristic.SerialNumber).value;
-        var firmwareRevision = service.getCharacteristic(Characteristic.FirmwareRevision).value;
-        var hardwareRevision = service.getCharacteristic(Characteristic.HardwareRevision).value;
-
-        if (manufacturer) existingService.setCharacteristic(Characteristic.Manufacturer, manufacturer);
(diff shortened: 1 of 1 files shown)


tobekas:
Name, Model, Manufacurer and SerialNumber must be persistent during the lifetime of the accessory. (by spec)
This may be the reason why these values are copied... But since Firmware/Hardware Revision may chance, this issue still has its raison d’être. See also: #2207

lib/server.js
```diff
@@ -437,28 +437,41 @@ Server.prototype._createAccessory = function(accessoryInstance, displayName, acc
...
 
     return accessory;
   }
 }
 
+Server.prototype._augmentService = function(srcService, dstService, characteristicUUID) {
+  var srcCharacteristic = srcService.getCharacteristic(characteristicUUID);
+  var dstCharacteristic = dstService.getCharacteristic(characteristicUUID);
+
+  if (!srcCharacteristic) return;
```
Supereg, line 460: This will always return false, as "getCharacteristic" will create the characteristic if it is not present.

Supereg:
This PR was pretty much replaced by https://github.com/homebridge/homebridge/pull/2472

But we could adopt the way listeners get replaced. But maybe we should add the code doing the replacements to hap-nodejs? So we can better track any possible changes made to event types.

afharo:
> This PR was pretty much replaced by #2472

I don't think that's entirely true. We might need to update this PR based on the changes in #2472 to match the new generic way of looping through all the services but the UUID one. 
But I don't think it's entirely replaced by it since the listeners are not passed through, which is the original intention of this PR.

donavanbecker:
closing in favor of #2476

PR #2186, opened 2019-03-27
Added an item to the list of common issues

hepcat72:
This is based on: https://alwayswithyouwhatcannotbedone.wordpress.com/2019/03/10/temporary-fix-for-an-iphone-that-cannot-access-local-addresses/


NorthernMan54:
Tks very much

Issue #2120, opened 2019-01-20
Possibility to clean/flush/rebuild cachedAccessories to remove old entries after plugin deletion


phenomeus:
@nfarina is there a clean way to get old entires from HomeKit removed after a plugin is uninstalled? (i.e.: switches from uninstalled homebridge-harmonyhub are still visible in homekit)

afaik I understood it, the cachedAccessories file keeps the entries there, but the modification by hand using nano is pain in the a and also there is a good chance that after saving the HomeKit service is not starting anymore. so the only solution I had, is to kill the file cachedAccessories but than I lose all my room organisation and naming and I have to start over with all the lights, sensors and switches. 

what I would like to use is maybe a possibility to --rebuild or --rescan command after a plugin uninstall. so that all unused and old entries are getting cleaned or flushed. 

what do you think?

ebaauw:
While refactoring my (platform) plugins to use the dynamic accessory model, I found that if you (your plugin) removes the accessory, it gets deleted from cachedAccessories as well.  I'm doing so when the plugin is loaded, but not specified in config.json.  This does require special user attention when uninstalling a plugin:
- Remove the entry from config.json;
- Restart homebridge with the plugin still installed to clean cachedAccessories;
- Uninstall the plugin;
- Restart homebridge again.

I think a better way could be not to restore the accessory in
https://github.com/nfarina/homebridge/blob/422582ad34a05f7d4510fbea7549a0461b65f063/lib/server.js#L357
when no `platformInstance` is found (i.e. when the plugin has been uninstalled).  If I understand correctly, this should remove the stale accessories, next time `cachedAccessories` is written.

phenomeus:
:/ right now that's not the case with all plugins right? so every dev has to implement the clean way. I can't do anything right now, correct?

liambolling:
Yeah this leaves it up to the dev. There should be a global way to flush the cache.

phenomeus:
can we bump this? a flush command or rebuild caches command with only the installed plugins + right config would be very helpful

ebaauw:
Happy to prepare a PR for this if we can agree on the specs: a command line option to delete cached accessories for which no plugin was loaded.  I.e. plugins that are no longer installed or, when `plugins` is specified in config.json, plugins that are no longer whitelisted.

phenomeus:
@ebaauw that would be awesome, exactly how you proposed

lboue:
Starting homebridge with -R option was fine for me 👍 
[...]

PR #2190, opened 2019-04-05
Clean orphaned cached accessories / handle uncaught exceptions

ebaauw:
See #2120:
- Add `-R` or `--remove-orphans` command line option;
- Clean orphaned cached accessores, for which no plugin was loaded, when `-R` was specified on the command line.

Handle uncaught exceptions to do a clean shutdown of homebridge instead of crashing NodeJS.

@nfarina, @KhaosT, please review.  Some technical notes:
- This change will unconditionally remove cached "stuff" that is not a `PlatformAccessory`.  I could add the check to do this only when `-R` was specified, but I see no reason why you would want to keep this "stuff" in the cache;
- I daren't remove elements from the `_cachedPlatformAccessories` while looping over them, hence the build up of `verifiedAccessories` to replace `_cachedPlatformAccessories` after the loop.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -12,6 +12,7 @@ module.exports = function() {
   var cleanCachedAccessories = false
   var insecureAccess = false;
   var hideQRCode = false;
+  var shuttingDown = false;
 
   program
     .version(version)
@@ -33,6 +34,11 @@ module.exports = function() {
   var signals = { 'SIGINT': 2, 'SIGTERM': 15 };
   Object.keys(signals).forEach(function (signal) {
     process.on(signal, function () {
+      if (shuttingDown) {
+        return
+      }
+      shuttingDown = true;
+
       log.info("Got %s, shutting down Homebridge...", signal);
 
       server._teardown();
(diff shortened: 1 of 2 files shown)


lib/cli.js
```diff
@@ -42,5 +49,12 @@ module.exports = function() {
     });
   });
 
+  process.on('uncaughtException', function(error) {
```
KhaosT: Is this related to the clean cached accessories?

ebaauw: Not really, sorry.  When I commit a new change, GitHub adds this to the open PR.

It’s actually more related to the check for the number of bridged accessories in HAP-NodeJS, since that error is unchecked by homebridge.  I’ve had this logic in my homebridge-zp plugin for months, to make sure the subscriptions with the Sonos speakers are canceled before homebridge exits.  Only then, the error is logged by the plugin logger, making it seem like my plugin caused the exception instead of only handling it.  See https://github.com/ebaauw/homebridge-zp/issues/41 and https://github.com/ebaauw/homebridge-zp/issues/49.

KhaosT:
LGTM Thanks! 🙏

lboue:
Thanks a lot, worked fine for me with homebridge 0.4.48. I noticed that in log file:
```
Failed to find plugin to handle accessory Det. Fumee etage
Removing orphaned accessory Det. Fumee etage
```

+1

PR #2205, opened 2019-04-24
Node.js 12 support

KhaosT:
To support Node.js 12, the underlying library HAP-NodeJS swapped library used for curve25519 and ed25519 to `libsodium.js`. Due to the introduction of async/await, I bumped node dependency to 8.0.0 or later.

Any concerns? @nfarina @ebaauw

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,13 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=4.3.2"
   },
   "preferGlobal": true,
   "dependencies": {
     "chalk": "^1.1.1",
     "commander": "2.8.1",
-    "hap-nodejs": "0.4.50",
+    "hap-nodejs": "0.4.51",
     "semver": "5.0.3",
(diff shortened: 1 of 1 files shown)


KhaosT:
Actually nvm. I managed to switch to `TweetNaCl` instead.

oznu:
Tested this, works fine.

It's great that we no longer have any post install compiling to do - this will make major upgrades to the users Node.js version much less troublesome.

nfarina:
I don't know what those libs are but I'm happy to not have to upgrade Node on my Pi 😂

ebaauw:
I'm also very happy to see the dependencies on C-modules disappear.

I develop and test my plugins against the latest LTS version of NodeJS (currently 10.15.3), so raising the dependency to v8.0.0 is no issue for me.

Just patched the `package.json` of my homebridge v0.4.48 installations on my development macBook and my test RaspBerry Pi to use HAP-NodeJS v0.4.51, deleted `node_modules` and ran `npm i`, and restarted homebridge.  No issues.  A bit late, perhaps, but I think this PR is good to go.

PR #2271, opened 2019-07-27
Update server.js

Frank-Friemel:
Hi, while logging/tracing a Plug-In I've noticed that I couldn't distinguish accessories which share the same name. So I've appended the uuid_base to the logging prefix. Thanks


Supereg:
Could you proved an example output of the old and the new style?

PR #2302, opened 2019-09-11
Website optimisations and improved development process

James-Bovis:
These commits implement the following features to the Homebridge website:
- Compress the compiled CSS saving 2KB in file size
- Remove unused dependencies
- Add a `browser-sync` gulp task which adds a web server for local development of the website and hot reloading every time a `.scss` file is changed.


nfarina:
Ah darn, I tried to run this locally but it looks like Gulp 3 doesn't work with node v12. I tried upgrading to Gulp 4 but the format has changed (I don't use Gulp myself).

PR #2355, opened 2019-11-18
It works, but not sure if it’s where you are headed

plainsane:
I just really need the functionality of hap 0.5.3 for my lock


KhaosT:
Can you share which specific feature you need hap-nodejs 0.5.x for the lock? I’m not aware anything new for lock in there…

plainsane:
i need to set the permissions on my lockdesiredstate for admins only.  i really just need the Accessory class.

oznu:
Closing in favour of #2391

PR #2380, opened 2020-01-03
Changes to support local debugging HAP-NodeJS

joconor:
I'm trying to test some potential changes to HAP-NodeJS within Homebridge. Two requires in homebridge make this difficult, as homebridge reaches into the HAP-NodeJS package expecting modules in specific paths. HAP-NodeJS has a different path structure as a published npm package vs a locally built git repo (where the built files are in a dist folder). This PR changes the two requires that 'reach in' to the HAP-NodeJS package.

In lib/server.js, require the `once` function exported from HAP-NodeJS. It is not necessary to reach into the package for this function.

In lib/api.js, handle the exception if requiring hap legacy types fails to find the module in the path expected in the published npm package. The catch clause retries the require, this time looking in the dist folder. 

These two changes allow debugging HAP-NodeJS when using a locally cloned git repo and npm link.


Supereg:
This is already addressed in the current beta version. See https://github.com/homebridge/homebridge/pull/2391

PR #2366, opened 2019-12-16
Bump hap-nodejs to use 0.5.0

jedis00:
Bump hap-nodejs to use 0.5.0.

PR #2391, opened 2020-01-16
Update hap-nodejs to v0.5.5

Supereg:
This is a follow up to #2366

@nfarina you mentioned that you maybe wanted to combine the typescript migration with the hap-nodejs update.
However I think this introduces some more unnecessary delay to the process. So the following PR should be seen as just a suggestion to speed this process up.

In this PR I updated hap-nodejs to the currently latest version and adjusted some of the direct access imports, which caused some problems with the new project setup of hap-nodejs.

I tested it with a simple switch plugin and couldn't find any more problems. I followed development of hap-nodejs closely since v0.5.0 and thus know that we haven't introduced any API changes which could cause further problems. However I'm happy about anyone proving me wrong regarding this and help fixing it.

~ Andi


nfarina:
Ok fair point! I’m happy to merge this but can’t publish at the moment as I’m traveling. @KhaosT do you mind merging and publishing?

Supereg:
@KhaosT sounded like he found some more severe API issues in the other PR, am I missing something?

KhaosT:
Here we’ll also need to bump the engine version. And to safe, we probably should ask a few plugin developers to try out their plugins with this version ^^

KhaosT:
@nfarina yeah I can take care of that.

Supereg:
Adjusted the way [once is imported](https://github.com/nfarina/homebridge/compare/e84ae119f9be25900d7561f8f9bdec0587e7e7f2..f73c75c7a3e4089a0d01dbeb43ec5473d408c7a7) since hap-nodejs exports this directly in the index.
Should be more robust that way.

KhaosT:
Additionally for those people really pushing to update to v0.5, we’ll need to make changes to PlatformAccessory so it properly exposes the multi-camera setup behind a bridge. The current one never expected that use case and restoring cached accessory with camera source doesn’t really work.

Supereg:
> Here we’ll also need to bump the engine version. And to safe, we probably should ask a few plugin developers to try out their plugins with this version ^^

Yeah thought of this as well, however the change of hap-nodejs to bump the required node version was not introduced with v0.5.5 so I decided to delay this change until homebridge updates to an hap-nodejs version which has this change.

Only thing I could do is to test this at my live instance at home and with my own plugins. But they aren't that complex I think.
As long as no plugin does the same direct path import everything should be fine.

Supereg:
> Additionally for those people really pushing to update to v0.5, we’ll need to make changes to PlatformAccessory so it properly exposes the multi-camera setup behind a bridge. The current one never expected that use case and restoring cached accessory with camera source doesn’t really work.

Hm okay, I'm pretty unfamiliar with the homebridge codebase, so not quite sure what you mean or rather what things need to be handled.

Is there any work need to add support for the newly added Apple TV remote controllers? I know that the current Remote API ideally needs a reference to the Accessory to properly instantiate links between services. AFAIK normal homebridge plugins do not have that reference, so currently only platform accessories would be supported(?).

Supereg:
@KhaosT So I browsed through some of the homebridge code base. The problematic part you were talking about is [restoring cached PlatformAccessories](https://github.com/nfarina/homebridge/blob/2a44b10cabee7adba66866efbc4872af80e136f8/lib/platformAccessory.js#L209), am I right?

Could you shortly clarify for what exactly the PlatformAccessory object is used. Following my current understanding it's the homebridge way for plugins to create Accessories without the need for devs to access the underlying hap accessory object (or expose any hap accessory internals). 
A PlatformAccessory can be registered in two ways, as an external accessory (camera accessory) or by registering as a platform accessory (loading it dynamically; the problematic way). Using the later way you would set it up once, homebridge would cache it on disk and you would need to add your event handlers on a new startup. (If I may ask, what the hell was the reason to add this caching feature 😅)

First of all I think this serialization and deserialization of services and characteristics is pretty doomed. Since every time we make a change to characteristic and service properties in hap-nodejs, we would need to try to not forget adding those changes here too (we did that also with v0.5.5. adminAccess was added to some special Apple TV remote characteristics). What about the idea, that we move this serializing stuff into hap-nodejs, so everything is at the same place (while of course respecting the current homebridge defined json structure to keep backwards compatibility)?

Secondly we want to add support for those cached accessories to support cameraSource (and probably also Apple TV remotes)? My current understanding is that cached camera accessories are indeed supported(?). Of course we cannot serialize the cameraSource/cameraHandler but as for set/get handlers the plugin would just set the cameraSource in the `configureAccessory` function and everything else would be handled internally by the hap-nodejs bridge? (We need to trust the plugins devs though that they correctly hook up the cameraSource in `configureAccessory`).
The only problematic thing would be RemoteControllers, as I did not know that homebridge does not expose a reference to the internal Accessory to the plugin developer rendering [my api](https://github.com/KhaosT/HAP-NodeJS/blob/a62a1c0ebe9ac34cb5b387bda9120a1062541004/src/lib/HomeKitRemoteController.ts#L474) pretty unusable. Also deserialization needs to be handled for remotes 🤔

I would derive three todos before merge:
* Move serialization/deserialization to hap-nodejs/or just support serialization/deserialization of the new `accessRestrictedToAdmins` property (depending for what I get the o.k.; imo the first one would be desired as it gives us more safety to change stuff and we do not need to play around with hap-nodejs internals)
* maybe handle the possibility that cached camera accessories which previously had an cameraSource attached, but haven't anymore after a restart and after calling `configureAccessory` are handled 'correctly' (don't know how we would handle that 🤔)
* add support for the RemoteController API

Feedback appreciated and information im currently missing @KhaosT @nfarina

Supereg:
Also it seems hidden and primary services are unsupported or rather not reconstructed/saved by the accessory cache 🤔

Supereg:
And maybe we could export the old legacy types.js from the accessories folder from hap-nodejs, publish all this in a new v0.5.6 update. Then we could adjust our import in homebridge to use the exported one and remove the direct file access, so we have one thing less to care about in the future.

KhaosT:
@Supereg Yeah the cache thing was problematic as we start doing more complicated stuff. The whole cache idea was introduced since a lot of plugin developers will do async discovery of accessories, and if they didn't re-register those accessories by the time we publish the HAP accessory, homed will assume those accessories were removed and remove them from user's home database. Having homebridge manage that resolved that issue.

I guess moving serialization to hap-nodejs is a good idea as that ensures the data can be properly save and restored when things changed.

Restoring cached camera accessory will not work currently since `configureCameraSource` will trigger homebridge [adding services to platform accessory](https://github.com/nfarina/homebridge/blob/master/lib/platformAccessory.js#L137), and for a cached accessory, it will trigger conflicts. I guess we can change `addService` to just override the underlying service when a duplicated service is detected... That way we can always be sure all the hooks are correctly wired up for the `cameraSource`.

For `RemoteController`, I guess we should expose a new API that takes in `HomeKitRemoteController` and manage the addition internally? And utilize the same service override logic as above.

Hope this make sense?

Supereg:
@KhaosT thanks for the clarification. Sounds pretty smart and works well for all accessories other than cameras 😅 

Your idea to just override the services seems pretty simple and straight forward (and is probably the fastest to roll out), however I think we could run into problems with that. We would need to trust the devs that they do not change the subtype after first creation. Since we currently identify our services by UUID and subtype, a change of the subtype on a startup would lead to, us not being able to find the service which needs to be replaced. This will result in having one service (or characteristic, the same applies here) twice and only one of them being hooked up correclty. For an individual installing such plugins this could get pretty hard to fix.

What I could imagine is that we integrate the controllers into the serialization and deserialization process (as proposed by [HAP-NodeJS/#780](https://github.com/KhaosT/HAP-NodeJS/pull/780)). This would probably take more work, take longer and maybe also complicate the configureAccessory function (devs would need to distinguish between setting up a cameraSource/controller and configuring/restoring one) but I think we could get some very good benefits from it.
First of all we solved the little problem explained above since we do not give devs the opportunity to change the subTypes (or anything else). Secondly we could solve the persistent storage problem for the [RemoteController](https://github.com/KhaosT/HAP-NodeJS/blob/7a612bcd5bfc866964b7f800e394275184dc94e7/src/lib/HomeKitRemoteController.ts#L310) easily. This becomes even more relevant for the upcoming addition of secure video support as it REQUIRES that the state of the selected configuration is stored across reboots.
Also additions of new controllers could just be added without the need to change any mechanisms in Homebridge on how those get added.

How could this look like? We would probably create a controller abstraction (some interface or abstract subclass). This could be then implemented by the RemoteController or the CameraSource. The controller abstraction would expose function to serialize and deserialize custom data as well to provide an initializer to supply previously created services.
When creating a new accessory a developer would create a new controller instance (maybe override it if needed) and call something like 'addController' on the accessory (homebridge would need to shadow this in PlatformAccessory object).
In the 'configureAccessory' function the plugin would just pass the constructor to the controller to the accessory (and maybe use bind to supply custom constructor arguments). The accessory would then initialize the controller by looking at the saved serialized json content.

So what's the drawback with this approach. While this would perfectly work for the RemoteController without breaking any existing API, I think this would not be so easy for the CameraSource. The problem here is that every plugin creates his custom CameraSource object/class. The hap-nodejs supplied [Camera](https://github.com/KhaosT/HAP-NodeJS/blob/7a612bcd5bfc866964b7f800e394275184dc94e7/src/lib/Camera.ts#L33) controller doesn't get used at all (and cannot really be used, as you cannot change any of the options if you subclass it). I would probably suggest creating a new CameraController class which implements the things explained above and deprecate the old CameraSource style implementation.
Would that hurt anyone? No! Current implementations could just continue to exist, only plugins devs who would like to create dynamic cameras (and probably people who want to support secure video) would need to adapt immediately. This would just be a hell lot more work and way more documentation to write if we want to enable any skill level to developer these plugins.

Additionally we should remember that we provide also an easy out of the box way for people only using hap-nodejs to enable them to persistently save the controller states too.


Would be nice if you could drop me some feedback on this. I hope I could explain it as brilliant and comprehensive as it is currently in my head 😅😂 and the reason why imo this is the far better solution.

JerGitHub:
This is more of a question since I have been observing this thread and had a question about the current state of homebridge because of this discussion.

Is the reason @Adrian.Cable is able to make all Nest products (Cameras, Alarm, Locks, Thermostats, Protects, etc.) work with his Starling (https://www.starlinghome.io) product because he forked homebridge and added hap-nodejs v.0.5.x awhile ago? 

And since the current homebridge has been behind for months without this updated version of hap-nodejs, we cannot use all of the Nest integrations he has working with Starling because of this?

Sorry if this is off-topi,c but I was generally trying to understand how Starling has all of the Nest integration working perfectly, yet no one can do the same with the current version of homebridge.

Supereg:
One short question. Can somebody explain to me the reason for the existence of the PlatformAccessory object? I don't get it 😅 Except for the [getServiceByUUIDAndSubType](https://github.com/nfarina/homebridge/blob/2a44b10cabee7adba66866efbc4872af80e136f8/lib/platformAccessory.js#L113) method pretty much everything is the exact same as in the hap-nodejs provided Accessory class (most code is even duplicated; some code already changed in hap-nodejs). I mean somebody issuing a PR in hap-nodejs to change certain things simply won't expect the same code lying here in homebridge again 😅 
Was it to isolate some methods? If that's the case, what things do you wanted to be isolated?
Anyone having more experience with Homebridge code base would be welcome, @KhaosT you seem to be the code owner though.

KhaosT:
I'm traveling this week so probably don't expect fast reply from me 😅

----
@JerGitHub I think he saw the need from people wanting a plug-and-work solution and created a dedicated hub hardware to fulfill the market. It's pretty cool actually 👍 With that in mind, I assume he developed the solution directly using hap-nodejs as the HAP library, since that's a lot easier and straightforward as most of homebridge plugins can be easily converted to a standalone HAP accessory with minimal changes.

Regarding homebridge didn't aggressively update to hap-nodejs v0.5.x, the full situation is a little more complicated than just bumping a version of the library. Since v0.5.x is basically a rewrote of the whole project in typescript, there are things that might accidentally get dropped during the rewrite. If you check the history (https://github.com/KhaosT/HAP-NodeJS/commits/master), you can see there are a lot fixes after v0.5.0 to address behavior mismatch or improve implementation accuracy. Since homebridge has a larger audience, I don't think it's responsible to just bump the version and hope it works if I can't allocate time to test it 😅 Especially when the new "camera behind bridge" feature can be workaround with existing APIs in homebridge.

Hope this answers your question :)
----

@Supereg I don't think we should worry too much about people changing subtypes after the first creation, since that's an issue we already have today for other services and the majority of use cases it shouldn't be the case.

The problem with serializing controller is that, typically those controllers has hooks can only be constructed by the code, upon restoring the serialized object, developer still need to reconfigure it to get it to work... A delegate style API might make things easier but it does impose the shape of implementation we are expecting, which I'm not sure is a good idea since plugins may need to behave differently in order to work with the accessory they tried to bridge?

For `PlatformAccessory`: There was at one point I'm thinking about moving plugins to a more managed solution and move them out of homebridge's process so a misbehaved plugin won't cause the entire homebridge to exit (https://github.com/nfarina/homebridge/issues/1039). In order to do that, there need to have something to be able to decouple HAP and handles XPC for individual "accessory". `PlatformAccessory` was an abstraction that hope to help us get there eventually but unfortunately due to a series of events in my personal life during that period, this wasn't a priority so it kinda got put on hold.

Supereg:
@KhaosT thanks for the reply despite of you holiday. I'm checking some options and maybe create some drafts, which illustrate my standpoint better. But first of all enjoy your holiday, everything else can wait.

JerGitHub:
@KhaosT thanks for the reply. That does make sense. And I agree - starling is cool for people that don't like to tinker.I guess one followup question would be - once hap-nodjs is updated to 0.5.x, would homebridge be on par with starling and allow enabling of all the same Nest products starling has enabled? I do like to tinker and am hoping to get all the Nest stuff working just like starling but on homebridge once hap-nodejs is on the 0.5.x train.

KhaosT:
@JerGitHub I mean... he can already implement everything with what we have today... He choose to use the camera behind the bridge thing added in recent iOS release, and that's the only reason it won't work with homebridge right now ^^ If you like tinker, you can totally rewrite that part to use the normal API available right now ;)

netRunner0:
Have a nice day, everyone!

Reading the good news above, I want to offer you such an idea (well, apart from making the plugins separately):

Since hb has the best implementation of HAP, can a constructor of devices for homekits be based on hb?
For example, I want to get a thermostat from a bluetooth sensor and a relay connected via mqtt
Or a humidifier, again from the relay and sensor

Yes, there is nodered, but it is extremely raw and complicated for such simple things 

How do you like this idea? It seems to me that this will push hb to further development, and the advent of automation in the future

JerGitHub:
@KhaosT I didn't realize that! I thought I was reading in other threads the new hap-nodejs was needed to get things going. I mean, I guess it is given the way he implemented camera support behind the bridge, but I didn't realize that it could be done today with the current API.Thanks!

oznu:
@KhaosT @Supereg we could publish a beta/test tag to npm that includes this change so willing testers could update and report any issues they encounter?

We would need to change [this line](https://github.com/nfarina/homebridge/blob/master/lib/plugin.js#L55-L57) to allow test versions from:

```js
  if (!semver.satisfies(version, versionRequired)) {
```

To:

```js
  if (!semver.satisfies(version, versionRequired, { includePrerelease: true })) {
```

Then change the version in the `package.json` to `0.5.0-beta.1`.

Then publish to npm under the `@beta` tag:

```
npm publish --tag=beta
```

Then users can install/update the beta version using:

```
npm install -g homebridge@beta
```

If a user encounters errors they can rollback using:

```
npm install -g homebridge@latest
```

I could also potentially add a feature to [homebridge-config-ui-x](https://github.com/oznu/homebridge-config-ui-x) that asks users if they would like to help test the beta version (with an option to roll back) which would really boost the number of testers.

Supereg:
I try to get back to working on this the next few days in order to maybe finally push some beta version or so

NorthernMan54:
Tks

Let us know when we should start testing our plugins

donavanbecker:
@Supereg Any update on this?

donavanbecker:
We have merged this to a beta branch that will allow us to test this until everything is figured out with hap-nodejs.

donavanbecker:
We will have a beta tag that will be released soon that plugin developers can test with.

Issue #2401, opened 2020-01-27
Node version requirements outdated? / Weird bug [doc issue?]


apparentorder:
### Expected Behavior
Pairing with Home app Just Works™ with any Node version >4.something (as documented), or at least unsupported versions throw an error.

### Current Behavior
- Pairing with Node 8.x: Doesn't work ("wrong" password)
- Pairing with Node 13.x: Works fine

So, I've spent several hours trying to figure out why pairing didn't work for me. The debug log kept complaining about the password, so did the Home app (iOS 13.3), with a minimal configuration / default password.

I have tried pretty much everything, changing port / username / password regularly, and even the "forget Wifi trick" (#1794) and similar stunts like Airplane Mode. But no luck. But also no error messages whatsoever, everything suggested there really is a password mismatch.

Finally I noticed that I had installed node8 instead of node "latest". After changing to node13, _the very same setup / configuration worked immediately_.

This seems extremely odd to me, as I would have expected errors, one way or another, when the Node version isn't properly supported or cryptographic functions are used "differently".

I assume debugging this would be extremely time-consuming and somewhat pointless -- but simply bumping the required Node version to 13 doesn't make sense either. This could even be FreeBSD- and/or ARM-specific behavior. Maybe a FAQ entry?

### Steps to Reproduce (for bugs)
I've been following this article: https://blog.zs64.net/2017/12/hooking-up-harmony-hub-and-generic-zigbee-lights-to-homekit/
Basically the steps are:
- pkg install npm-node8 avahi-libdns
- npm install mdns, homebridge (global, unsafe)
- run with minimalist/default config.json

### Versions

- **Homebridge**: 0.4.50
- **Node**:       13.3.0 [https://www.freshports.org/www/node]
- Node8: 8.16.2 [https://www.freshports.org/www/node8]
- **OS**:         FreeBSD 12.1-RELEASE-p1 on armv7 (RPi2)

### Homebridge Log / Command Output
Debug log gives:
```
Error while checking pincode: client did not use the same password
```

jvmahon:
hap-nodejs, the underlying library, recently updated their minimum node requirement to ">=10.17.0" so I think it would make sense if homebridge used the same minimum.

apparentorder:
fwiw, that makes good sense to me :-) I'd be happy to re-test and verify that it still works with 10.17.0 >= x < 11, if that would help.

apparentorder:
[...]

PR #2407, opened 2020-01-31
Create NodeJs Workflow

donavanbecker:


.github/workflow/nodejs.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install, build and test
+        run: |
+          npm ci
+          npm run build --if-present
+          npm test
```
alexindigo: There is no `test` or `ci` commands in this project – https://github.com/nfarina/homebridge/blob/master/package.json#L6

.github/workflow/nodejs.yml
```diff
@@ -0,0 +1,27 @@
+name: Node-CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 13.x]
```
alexindigo: This doesn't match announced version support: https://github.com/nfarina/homebridge/blob/master/package.json#L24

.github/workflow/nodejs.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install, build and test
+        run: |
+          npm ci
+          npm run build --if-present
```
alexindigo: Why wouldn't `build` command be first?

donavanbecker: This is just like [hap-nodejs](https://github.com/KhaosT/HAP-NodeJS/raw/master/.github/workflows/nodejs.yml)

.github/workflow/nodejs.yml
donavanbecker: This is just like [hap-nodejs](https://github.com/KhaosT/HAP-NodeJS/raw/master/.github/workflows/nodejs.yml)

.github/workflow/nodejs.yml
donavanbecker: This is just like [hap-nodejs](https://github.com/KhaosT/HAP-NodeJS/raw/master/.github/workflows/nodejs.yml)

Should 4.x be added?

.github/workflow/nodejs.yml
alexindigo: `hap-nodejs` has more elaborate scripts section https://github.com/KhaosT/HAP-NodeJS/blob/master/package.json#L31-L35

You can try to run `npm test` locally and see that it would error out.

This projects needs tests before it could be automated to run tests in CI. :)

donavanbecker: Do you wanna change it?

.github/workflow/nodejs.yml
alexindigo: Looks like we better wait for this one to be decided on – https://github.com/nfarina/homebridge/issues/2401

Meanwhile, keeping those three should be fine, probably.

.github/workflow/nodejs.yml
alexindigo: I mean, it will fail CI, because test script isn't defined. If you add some rudimentary test script, it would a good start and it will make that CI runs meaningful, otherwise, all we'd get is failed status marks on each PR, not really useful.

PR #2411, opened 2020-02-02
Fixed typo in `peerDependencies` check

alexindigo:


nfarina:
Oops 🤦‍♂ Thanks!

fkistner:
Hi,

Unfortunately, it seems this is causing some problems for me.
This typo fix is a breaking change in "published api" that prevents homebridge from starting correctly with some old plugins:
```
Error: Plugin /homebridge/node_modules/homebridge-milight does not contain the 'homebridge' package in 'engines'.
    at Plugin.load (/usr/local/lib/node_modules/homebridge/lib/plugin.js:49:11)
    at Server.<anonymous> (/usr/local/lib/node_modules/homebridge/lib/server.js:153:14)
    at Array.forEach (<anonymous>)
    at Server._loadPlugins (/usr/local/lib/node_modules/homebridge/lib/server.js:145:22)
    at new Server (/usr/local/lib/node_modules/homebridge/lib/server.js:57:24)
    at module.exports (/usr/local/lib/node_modules/homebridge/lib/cli.js:32:16)
    at Object.<anonymous> (/usr/local/lib/node_modules/homebridge/bin/homebridge:17:22)
    at Module._compile (internal/modules/cjs/loader.js:1158:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
```
(compare with [homebridge-milight#package.json](https://github.com/dotsam/homebridge-milight/blob/7f6e50a/package.json))

AFAIU, this is a workaround solely for legacy plugins and new plugins are discouraged from using it.
Would you consider reverting this commit to retain compatibility with old plugins?

donavanbecker:
@fkistner instead of reverting this commit, why not put a pull request into that plugin to get it updated so that we can advance homebridge?

alexindigo:
@nfarina I think we need to stop requiring “engines” field in package.json, since package.json is node thing and homebridge is abstraction on top of that. It meant to describe what can execute your package and it’s always node, homebridge doesn’t parse/evaluate plugins code. And npm/yarn actually complains about unknown engine there.

What could be better is to have standalone “homebridge” branch in package.json, that can have versions and much more, and will be aligned with how package.json meant to be used, and how other meta projects do that.

Something like:

```
“name”: “homebridge-plugin”,
“version”: “1.0.0”,
“dependencies”: {},
“homebridge”: {
  “version”: “0.4.51”
  “other”: “things”,
  “itWants”: “to ask for”
}
```

fkistner:
> instead of reverting this commit, why not put a pull request into that plugin to get it updated so that we can advance homebridge?

I agree it makes sense to try to get this fixed upstream, but to me that is a separate discussion.
This is a compatibility shim. The change unnecessarily broke (potentially many other) plugins in order to fix a typo.

alexindigo:
@fkistner There shouldn't be many plugins like that, since it's a typo in one of the npm specific fields and it prevents packages from utilizing that functionality.

PR #2430, opened 2020-02-29
Make manufacturer & model update-able in the bridge config

donavanbecker:
Proposing that we can change the manufacturer and model. This would make it be update-able in the bridge config.

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -2,7 +2,7 @@
   "bridge": {
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
-    "manufacturer": "Homebridge.io",
+    "manufacturer": "homebridge.io",
     "model": "homebridge",
     "port": 51826,
     "pin": "031-45-154"
--- a/config-sample.json
+++ b/config-sample.json
@@ -3,7 +3,6 @@
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
     "manufacturer": "Homebridge.io",
-    "firmware": "0.4.51",
     "model": "homebridge",
     "port": 51826,
     "pin": "031-45-154"
--- a/config-sample.json
+++ b/config-sample.json
@@ -3,8 +3,8 @@
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
     "manufacturer": "Homebridge.io",
-    "firmware": "4.11.0",
-    "model": "config-ui-x",
+    "firmware": "0.4.51",
+    "model": "homebridge",
     "port": 51826,
     "pin": "031-45-154"
   },
--- a/config-sample.json
+++ b/config-sample.json
@@ -2,6 +2,9 @@
   "bridge": {
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
+    "manufacturer": "Homebridge.io",
+    "firmware": "4.11.0",
+    "model": "config-ui-x",
     "port": 51826,
     "pin": "031-45-154"
   },
--- a/lib/server.js
+++ b/lib/server.js
@@ -108,10 +108,10 @@ Server.prototype._publish = function() {
   var bridgeConfig = this._config.bridge || {};
 
   var info = this._bridge.getService(Service.AccessoryInformation);
-  info.setCharacteristic(Characteristic.Manufacturer, `${toTitleCase(require('../package.json').author.name)}`);
-  info.setCharacteristic(Characteristic.Model, `${toTitleCase(require('../package.json').name)}`);
+  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer);
+  info.setCharacteristic(Characteristic.Model, bridgeConfig.model);
   info.setCharacteristic(Characteristic.SerialNumber, bridgeConfig.username);
-  info.setCharacteristic(Characteristic.FirmwareRevision, require('../package.json').version);
+  info.setCharacteristic(Characteristic.FirmwareRevision, bridgeConfig.firmware);
 
   this._bridge.on('listening', function(port) {
     log.info("Homebridge is running on port %s.", port);
--- a/lib/server.js
+++ b/lib/server.js
@@ -108,10 +108,10 @@ Server.prototype._publish = function() {
   var bridgeConfig = this._config.bridge || {};
 
   var info = this._bridge.getService(Service.AccessoryInformation);
-  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer || `${toTitleCase(require('../package.json').author.name)}`);
-  info.setCharacteristic(Characteristic.Model, bridgeConfig.model || `${toTitleCase(require('../package.json').name)}`);
+  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer || `${toLowerCase(require('../package.json').author.name)}`);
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "homebridge",
   "description": "HomeKit support for the impatient",
-  "version": "0.4.50",
+  "version": "0.4.51",
   "scripts": {
     "dev": "DEBUG=* ./bin/homebridge -D -P example-plugins/ || true"
   },
   "author": {
-    "name": "Nick Farina"
+    "name": "Homebridge"
```
alexindigo: I think it's better to keep @nfarina as an author, since it's metadata for npm and used for number of purposes,
and just decouple it from Manufacturer field in the Homebridge itself. Either hardcode it or read it from some config file.

donavanbecker:
@alexindigo so it is something we could possibly do?

package.json
donavanbecker: @alexindigo i will make it look at the config.

ebaauw:
Please continue to use the values from package.json, if nothing is specified in config.json.

ebaauw:
That's not what I meant: leave the old values in the code when the config.json doesn't contain these values.

donavanbecker:
> That's not what I meant: leave the old values in the code when the config.json doesn't contain these values.

I was just changing that for the same not from your comment. if someone wants to add that to this pull request on my fork - patch 1 branch then we would be good.

donavanbecker:
@ebaauw @alexindigo I have added what is needed so that if the config isn't set in the config.json then it will pull from the package.json.

donavanbecker:
@nfarina @KhaosT Would be nice if we could set these values.

alexindigo:
@donavanbecker Thanks for going extra mile. Looks great, although keeping firmware version in sync will be a chore, do you see any benefits in ability to override real firmware version (package.json) with a custom one from the config?
Thanks for the suggestion @ebaauw 

@nfarina this one looks like definite improvement. Please take a look.

donavanbecker:
@alexindigo only added firmware for customization. But I think keeping it synced with the homebridge version would be a good idea.  I have removed that in my last commit. So once @nfarina or @KhaosT look at this. We should be good to go.

nfarina:
Seems fine to me! OK to merge?

config-sample.json
```diff
@@ -2,6 +2,9 @@
   "bridge": {
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
+    "manufacturer": "Homebridge.io",
+    "firmware": "0.4.51",
```
nfarina: The `firmware` value doesn't seem to be used anywhere?

nfarina: Ah I see you removed it. So it should probably be removed from this sample config as well?

donavanbecker: Just removed it.

donavanbecker:
@nfarina let me know if you think anything else should be changed.

nfarina:
Wait ... ha ... I never realized the manufacturer is "Nick Farina" by default - Looks like I merged a PR years ago that just took some values from `package.json`. This seems wrong to me. I don't know why these values would come from `package.json` (or why it would default to my name).

It seems like these values should just be hardcoded to what you have in the example config - "Homebridge.io" and "homebridge". Any good reason not to do that?

donavanbecker:
@nfarina sure we can definitely do that. I just added what I added so that it would be customizeable to users if they please.

nfarina:
Would you mind updating this PR to change it to "homebridge.io" and "homebridge" (all lowercase) instead of reading from `package.json`?

donavanbecker:
do you want to leave it customized or have it hardcoded?

nfarina:
What was the original rationale for having it customizable? (Besides removing my name which I totally agree with)

donavanbecker:
@nfarina was really to allow people to customize the name to their liking. Especially if they are running multiple instances of HomeBridge, then they can name them each whatever they would like. But I totally understand if you wanna leave it and set it as homebridge.io

donavanbecker:
@nfarina tell me what You would prefer and I will change it in this PR and then we will be good!

lib/server.js
```diff
@@ -108,10 +108,10 @@ Server.prototype._publish = function() {
   var bridgeConfig = this._config.bridge || {};
 
   var info = this._bridge.getService(Service.AccessoryInformation);
-  info.setCharacteristic(Characteristic.Manufacturer, `${toTitleCase(require('../package.json').author.name)}`);
-  info.setCharacteristic(Characteristic.Model, `${toTitleCase(require('../package.json').name)}`);
+  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer || `${toTitleCase(require('../package.json').author.name)}`);
+  info.setCharacteristic(Characteristic.Model, bridgeConfig.model || `${toTitleCase(require('../package.json').name)}`);
   info.setCharacteristic(Characteristic.SerialNumber, bridgeConfig.username);
-  info.setCharacteristic(Characteristic.FirmwareRevision, require('../package.json').version);
+  info.setCharacteristic(Characteristic.FirmwareRevision, `${toTitleCase(require('../package.json').version)}`);
```
nfarina: Why toTitleCase here?

config-sample.json
```diff
@@ -2,6 +2,8 @@
   "bridge": {
     "name": "Homebridge",
     "username": "CC:22:3D:E3:CE:30",
+    "manufacturer": "Homebridge.io",
```
nfarina: Could this be `homebridge.io` all lowercase?

nfarina:
> We could also set the manufacturer to HomeBridge.io and then the @oznu suggested using the bridge name in the config.json? So we could us that for the model?

I don't quite follow what you mean here?

lib/server.js
donavanbecker: I was just copy and pasting from the top and changing it to firmware. I will revert it back.

config-sample.json
donavanbecker: Done

lib/server.js
```diff
@@ -108,8 +108,8 @@ Server.prototype._publish = function() {
   var bridgeConfig = this._config.bridge || {};
 
   var info = this._bridge.getService(Service.AccessoryInformation);
-  info.setCharacteristic(Characteristic.Manufacturer, `${toTitleCase(require('../package.json').author.name)}`);
-  info.setCharacteristic(Characteristic.Model, `${toTitleCase(require('../package.json').name)}`);
+  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer || `${toLowerCase(require('../package.json').author.name)}`);
```
oznu: Did you want to make the default `homebridge.io`?

```js
  info.setCharacteristic(Characteristic.Manufacturer, bridgeConfig.manufacturer || 'homebridge.io'`);
```

donavanbecker: I updated it for this!

donavanbecker:
> > We could also set the manufacturer to HomeBridge.io and then the @oznu suggested using the bridge name in the config.json? So we could us that for the model?
> 
> I don't quite follow what you mean here?

Ignore that.

@nfarina i have completed changing it to homebridge.io for manufacturer and homebridge for model

nfarina:
Looks great. Thanks all

donavanbecker:
Thanks for merging!

alexindigo:
Thanks @donavanbecker for seeing it all the way through.

donavanbecker:
@alexindigo no problem.

PR #2446, opened 2020-03-14
Update Readme and Package.json for homebridge organization changes.

donavanbecker:
Don't merge right away

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,10 @@ If you're having an issue with a particular plugin, open an issue in that plugin
 
 This is the offical Homebridge community on Reddit.
 
-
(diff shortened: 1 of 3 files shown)


oznu:
Suggested changes here: https://github.com/homebridge/homebridge/tree/donavanbecker/master

I don't think we want to go overboard with lots of badges, and need to keep the intro short otherwise people will just skip over it.

I swapped out the logo for a variant that did not include the "Homebridge" text embedded in the image so we could keep "Homebridge" text in the markdown as text. We can swap this out for the circle version of the logo if you don't like the one I picked.

oznu:
Merging into "readme-update" branch for further work before merging into master.

PR #2447, opened 2020-03-16
Add support for @scoped/homebridge- plugins

oznu:
This pull request adds support for scoped npm packages as Homebridge plugins.

Plugins in the following format will be matched:

```
@scope/homebridge-dummy
```

Plugins are registered in Homebridge under their scope:

```
Loaded plugin: @scope/homebridge-dummy
```

Tested and working with globally installed plugins, as well as when Homebridge is pointing to a specific folder for a plugin using the `-P` flag.

### DIFF
--- a/lib/plugin.js
+++ b/lib/plugin.js
@@ -17,13 +17,14 @@ module.exports = {
  * Allows for discovering and loading installed Homebridge plugins.
  */
 
-function Plugin(pluginPath) {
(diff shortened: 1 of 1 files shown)


donavanbecker:
after updating to the latest homebridge-config-ui-x@test (v4.12.1-test.18) Everything looks good! Ran like a charm!

PR #2455, opened 2020-03-19
Update Engine - Node

donavanbecker:
This PR is to update the node engine to be in line with ```hap-nodejs``` and their node engine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=10.17.0
+    "node": ">=10.17.0"
   },
   "files": [
     "README.md",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=4.3.2"
+    "node": ">=10.17.0
   },
   "files": [
     "README.md",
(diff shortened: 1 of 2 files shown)


donavanbecker:
> This is good, but wondering if we should set the minimum to be the current nodeJS LTS version?
> 
would we want to do ```latest``` or ```stable```

donavanbecker:
@NorthernMan54 I have updated it to be ```Node.js Version | v12.16.1```

package.json
```diff
@@ -21,7 +21,7 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=10.17.0"
+    "node": ">=12.16.1"
```
donavanbecker: @NorthernMan54 I have updated it to be ```Node.js Version | v12.16.1```

oznu:
Set the minimum in engines to v8.15.1. Everything still works on this version.

Update the readme to say v10+ as that's a "Current" LTS still.

oznu:
Forcing people to update Node.js more often than they have to will cause a spike in issues.

donavanbecker:
> Set the minimum in engines to v8.15.1. Everything still works on this version.
> 
> Update the readme to say v10+ as that's a "Current" LTS still.

@oznu I have updated the readme to say v10 and also changed the engine to v18.15.1.

PR #2468, opened 2020-03-31
Labeler for Pull Requests

donavanbecker:
This will auto assign labels to pull requests based off of what file/folder/subfolder is being changed.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
-# Add 'website' to any changes within 'website' folder/subfolders and index.html
+# Add 'website' to any changes to website files
 website:
   - website/**/*
   - ./index.html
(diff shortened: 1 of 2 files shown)


AppleTechy:
Not sure it is entirely necessary we have a low number of PRs that manually adding a specific label isn't time consuming.  Plus with PRS devs are pretty good about giving a detailed explanation as to what it is doing.

donavanbecker:
> Not sure it is entirely necessary we have a low number of PRs that manually adding a specific label isn't time consuming. Plus with PRS devs are pretty good about giving a detailed explanation as to what it is doing.
> Not sure I have an opinion on this, but let's see how this works out. I think most users (incl. myself) will check release notes of a new version to see what changed before upgrading.

Sounds good! I will close for now.

PR #2476, opened 2020-04-04
Migrating project to Typescript

Supereg:
Okay, so. I think that's the third pull request regarding a typescript migration on this project (https://github.com/homebridge/homebridge/pull/2266, https://github.com/homebridge/homebridge/pull/2393). Both of them a currently not able to get merged and both of them don't seem to work (at least that is what I've been told).
[Edit: had a look at them, and there are some not so got parts about them]

This PR aims to finally migrate this project. And I think with the ongoing beta phase we have a higher confidence to release it to the public (without backtracking on prior testing and code review!).

Currently this PR only contains configuration changes for the new project. The idea was to split configuration changes and actual code changes into (at least) two commits, to improve code review. As I don't know what exact project configuration is preferred, I wanted to use this as a starting point to propose some configurations and let them be discussed here. Any requested changes can currently be incorporated into the now ongoing process of migration.
Ideally this PR should suggest to other contributors to maybe delay some of their PRs, if they plan to make bigger changes in the Code, as otherwise this will result in a big ass merge conflict 🙃

The plan is to squash all commits when the PR is ready, as I currently do not guarantee a working project on every commit

-----
#### Project Configuration
* Migrated to Typescript
    * created tsconfig.json
    * created `src/` and moved contents of `lib/` into it (`lib/` will be the output dir so backwards compatibility for project structure is provided)
* Added ESLint
* Added jest for future unit tests
* Adjusted GitHub Actions "build" jobs
-----

Current PR Status: **READY TO MERGE**


Supereg:
Okay, so to further improve the process of code review I try to create a commit for every migrated file. Don't know how well this is gonna be for the bigger ones, but works for all the util files.

Supereg:
> I don't have any experience with TypeScript yet. I'm a bit worried about the increased footprint, but I understand run-time, you'd only use the JavaScript files generated by the TypeScript compiler. I'm happy to test my plugins against a beta, but I'm afraid I won't be of much value in a code review.

I'm happy to help if you got any questions. Nonetheless it could be helpful if you try to check changes made, to maybe spot any errors made in **logic** (not necessarily in language).
Typescript transpiles into JavaScript, exactly. This happens before we publish the package. So we still deploy JavaScript (and additional type files generated by the Typescript compiler used by IDEs). As a result any normal Javascript homebridge plugin can still be used.

But you could also run Typescript directly using `ts-node` and basically compile it on the fly, which results in a reduced execution/startup time though.

ebaauw:
What IDE do you use or recommend?

Supereg:
> What IDE do you use or recommend? I currently use Atom with the `standard` plugin for some code checking, but not nearly enough.

I'm currently using WebStorm. Has some pretty nice  auto completion and some decent default linting, but also adopts dynamically if project specifies an eslint configuration for example. I know that VSCode does the same AFAIK.

hassankhan:
Just wanted to thank you @Supereg 👍

Supereg:
BridgeSetupManager was hell to type 🙃

Supereg:
Last thing to migrate is server.ts

src/plugin.ts
```diff
@@ -1,236 +1,247 @@
...
-        // throws an Error if this isn't a homebridge plugin
-        pjson = Plugin.loadPackageJSON(pluginPath);
-      }
-      catch (err) {
-        // is this "trying" to be a homebridge plugin? if so let you know what went wrong.
-        if (!name || name.indexOf('homebridge-') === 0 || (name.charAt(0) === '@' && name.split(path.sep).length > 1 && name.split(path.sep)[1].indexOf('homebridge-') === 0)) {
-          log.warn(err.message);
+    private static loadPlugin(absolutePath: string): Plugin {
+        const packageJson: PackageJSON = PluginManager.loadPackageJSON(absolutePath);
+        const name = PluginManager.extractPluginName(packageJson.name);
```
oznu: Does this register the plugin in homebridge without the `@scope/`? If so we should be keeping the `@scope` so two plugins with the same name, but different scopes can be identified easily.

Supereg: Yeah it does actually. I tried to scrap the scope as soon as possible to try to make the code more understandable and more clear what is meant when the code refers to "pluginName".

Should this actually be a feature? I suspected a plugin with the same name but a different scope to just be the same plugin published in an organization(?). Also current hombridge code is not able to handle that either: https://github.com/homebridge/homebridge/blob/e67630a3978aa126786711f8ec06610db517ca74/lib/server.js#L166 (would register the plugin correctly, would just be overwritten in the dic maintained by the server [which is never used though, to be fair]).

So to get the change clear, two homebridge plugin with the same name but different scope should be treated as two different plugins? Will change that later.

oznu: > So to get the change clear, two homebridge plugin with the same name but different scope should be treated as two different plugins? 

Yes, they should be treated as seperate plugins. This is so people can publish their own versions of existing plugins to test / fork, or just publish under their own scope if they feel like it. By keeping the scope in place the logs will also show exactly which plugin got loaded which will make things easier to troubleshoot when users send us their logs.

>  Also current hombridge code is not able to handle that either:

`plugin.name()` will return the `@oznu/homebridge-example` in the current codebase:

https://github.com/homebridge/homebridge/blob/e67630a3978aa126786711f8ec06610db517ca74/lib/plugin.js#L26-L28

I think just using exactly what is in the `name` attribute of the `package.json` is probably reasonable and simplest way to go. Thoughts?

Supereg: Lol I did read it the other way round 😅😄

Supereg: > I think just using exactly what is in the name attribute of the package.json is probably reasonable and simplest way to go. Thoughts?

Yeah probably. I just add a try catch in the server when we call the initializer. Should there be a conflict we want to fail gracefully.

oznu: > Should there be a conflict we want to fail gracefully.

Sounds good. That's what the existing code does if it finds the same plugin in two different paths.

Supereg:
Some quick question, If it's okay I would move some of the stuff inside server.ts to places they suite better. For example some of the plugin loading stuff inside server.ts would get move into plugin.ts. Don't know what's better for review. Otherwise I would just do a migration while trying to keep most of the current structure and do all the refactoring in another PR or in a follow up commit in this PR. @oznu

src/plugin.ts
Supereg: Behavior was restored in latest force push.

oznu:
It might be easier to keep the structure similar to the current code for now. Maybe we can do an initial beta rerelease with these changes (once complete) then a follow up beta with the restructure - all before 1.0.0?

Supereg:
Okay great. Thought readability could definitely be improved just my moving some config loading and plugin loading into the respective plugin.ts and user.ts (which is basically a config.ts).

Supereg:
Okay migration should be complete. I tested it on a basic installation with homebridge-dummy enabled. Would want to run/test it on my local instance before merging it. You all are welcome to do your own tests though.

oznu:
Nice work! Thanks @Supereg.

I'll test this out tonight.

oznu:
Should we include https://github.com/evanw/node-source-map-support ?

This will make stack traces alot easier to follow.

oznu:
https://github.com/homebridge/homebridge/commit/f6f3d87ea6f827b506896a8949d4cb5584b5e8ed needs to go in to prevent these errors:

```
[4/8/2020, 5:39:36 PM] TypeError: Cannot read property 'updateValue' of undefined
    at /usr/local/lib/node_modules/homebridge/src/server.ts:523:82
```

oznu:
scoped module paths are not resolved correctly.

Example: https://www.npmjs.com/package/@kacepe/homebridge-shelly

Homebridge tried to load the plugin from `/homebridge/node_modules/homebridge-shelly` instead of `/homebridge/node_modules/@kacepe/homebridge-shelly/`

```
[4/8/2020, 5:43:33 PM] Plugin /homebridge/node_modules/homebridge-shelly does not contain a package.json.
```

.eslintrc
```diff
@@ -0,0 +1,31 @@
...
+    "ecmaVersion": 2018,
+    "sourceType": "module"
+  },
+  "ignorePatterns": [
+    "bin/",
+    "lib/"
+  ],
+  "rules": {
+    "quotes": ["error", "double"],
+    "indent": ["error", 4, { "SwitchCase": 1 }],
```
oznu, line 19: HAPNodeJS uses two, not sure if we want to follow in style.

Supereg: That is exactly one of the things I wanted to be discussed with my initial commit.

HAP-NodeJS is basically using both 😅 It's not quite consistent.
I personally grew up using four so that's where the default values is coming from. But idk, also long we can be consistent on one.

Supereg:
> [f6f3d87](https://github.com/homebridge/homebridge/commit/f6f3d87ea6f827b506896a8949d4cb5584b5e8ed) needs to go in to prevent these errors:
> 
> ```
> [4/8/2020, 5:39:36 PM] TypeError: Cannot read property 'updateValue' of undefined
>     at /usr/local/lib/node_modules/homebridge/src/server.ts:523:82
> ```

Oh that's where this part of the code is coming from I was surprisingly unfamiliar with. You may made the error go away, but HAP-NodeJSs getCharacteristic is a bit odd in the way that it will forcefully add the characteristic anyway (not just if it is contained in the optionalCharacteristics). I rather think my original PR was passing the wrong type. Your fix makes the error go away but also prevents the characteristics ever get copied.
This PR fixed it my passing the `typeof characteristic`. Works and I do not have any issues with it. Or if the issue persist could you just state which characteristics are exposed in your AccessoryInformationService?

Supereg:
Added two commits to add source-map support and fixing scoped plugins. @oznu

Supereg:
> [f6f3d87](https://github.com/homebridge/homebridge/commit/f6f3d87ea6f827b506896a8949d4cb5584b5e8ed) needs to go in to prevent these errors:
> 
> ```
> [4/8/2020, 5:39:36 PM] TypeError: Cannot read property 'updateValue' of undefined
>     at /usr/local/lib/node_modules/homebridge/src/server.ts:523:82
> ```

Okay this is actually somewhat of a bigger problem. The `getCharacteristic` of HAP-NodeJS is simply not designed to do that. I created a method over at https://github.com/KhaosT/HAP-NodeJS/pull/807 to create a HAP-NodeJS supported way of replacing a services characteristics with characteristics from another service. This also deals with get and set listeners thus fully replacing https://github.com/homebridge/homebridge/pull/2169.

Supereg:
Okay, so the current state of the PR is now feature complete. It runs already the latest HAP-NodeJS beta (with the revamped Camera controllers and stuff, but support for this will be moved to another PR). Currently runs smooth for me on my personal instance (with some cameras and some basic plugins).
I'm currently a bit unhappy about the way plugin devs will interface with hap-nodejs exported classes. Currently everything is exported from hap-nodejs through homebridge. But IDEs (at least for my experience) have a hard time auto completing those imports which come from hap-nodejs (as they seem to not recursively dig through all modules). Anyone an idea how to improve on that? (@oznu). I will still do some experimenting and create some example plugins (for myself), to check the usage of the API.

Besides that I'm just waiting for everyone doing a code review to finish with that. Testers are welcome too. If I get the okay, I will restructure this commit as above explained to only consist of two commits. Only after that I would suggest to merge into the beta branch.

oznu:
Thanks @Supereg, the new update are working well.

I'm happy for this to be merged into the `beta` branch when you are.

Supereg:
Okay, squashed all changes into two commits (which should be preserved when merging).
Reformatted code to enforce indent of 2.

I'm ready to merge (into beta!). I would address any further changes to the exposure of the types for plugin developers in a follow up PR if that's okay.
@oznu

PR #2479, opened 2020-04-08
Update node workflows

donavanbecker:
updating node workflows so that they don't require the package-lock.json


donavanbecker:
> `npm ci` is the correct command here. It ensures that the `package-lock.json` is up to date, which it always should be. If `npm ci` fails, then the action should fail so the committer can update `package-lock.json`

but this doesn't update the `package-lock.json` in the repo though? maybe we can leave `npm ci` but add something that would update the `package-lock.json`?

dgreif:
`package-lock.json` is updated automatically whenever a homebridge developer does `npm i` locally.  If you are working in a branch and change/update dependencies, your `package-lock.json` will update accordingly.  `npm ci` ensures that your `package-lock.json` is valid based on the dependencies listed in `package.json`, and ensures that the exact same dependency tree is installed in ci as what you installed in your local environment.

dgreif:
@donavanbecker was this change requested by one of the homebridge developers?

donavanbecker:
> @donavanbecker was this change requested by one of the homebridge developers?

@oznu and I have discussed this.

donavanbecker:
> `package-lock.json` is updated automatically whenever a homebridge developer does `npm i` locally. If you are working in a branch and change/update dependencies, your `package-lock.json` will update accordingly. `npm ci` ensures that your `package-lock.json` is valid based on the dependencies listed in `package.json`, and ensures that the exact same dependency tree is installed in ci as what you installed in your local environment.

so should we do the same workflow as HAP-nodeJS then @homebridge/collaborators?

Supereg:
I think the workflow is good as is. Triggering on GitHub release creation is the thing we want for homebridge and config-ui? Maybe we want to run on multiple node version not only on the latest LTS (though about 10, 12 and maybe the current 13). We should definitely add node v10 as this is currently our requirement.

donavanbecker:
@dgreif and @Supereg I have added the `package-lock.json` back, I have added the node versions. only thing that we maybe want to add is the [nodejs.yml](https://github.com/donavanbecker/homebridge-honeywell-home/blob/master/.github/workflows/nodejs.yml) that runs on push and commits?

.github/workflows/npmpublish-beta.yml
```diff
@@ -18,14 +18,18 @@ jobs:
...
+
   publish-npm:
     needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: ${{ matrix.node-version }}
```
Supereg, line 32: You defined the different node version for the publish job. This will result in the publish job running multiple time (3 time exactly for every version one). And the build job continues to run on nodev12 (only). We would want to run the build job on the node version matrix and if the build jobs succeeds for every node version the publish job should run once (with some predefined node version like 10 or so). Same applies for the npmpublish.yml

.github/workflows/npmpublish.yml
```diff
@@ -13,13 +13,17 @@ jobs:
...
+        node-version: [10.x, 12.x, 13.x]
+
   publish-npm:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: ${{ matrix.node-version }}
```
Supereg, line 26: publish still runs multiple times here

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Cache Node Modules
+        uses: actions/cache@v1
+        id: cache
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+
+      - name: Install NPM
+        if: steps.cache.outputs.cache-hit != 'true'
```
Supereg, line 24: It seems to me that build and test won't happen if node modules are still cached? Or am I wrong?

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 13.x]
```
Supereg, line 11: and actions/setup-node seems to be never called with those version in the below steps

.github/workflows/npmpublish-beta.yml
dgreif: Totally agree @Supereg.  Only `nodejs.yml` needs `matrix` since it's responsible for running the tests on all the different versions.  Ideally the publish wouldn't go through unless all of those tests pass

.github/workflows/nodejs.yml
dgreif: Agreed.  There needs to be another, separate step after `Install NPM` that always runs `build` and `test`

donavanbecker: I have added these extra steps

.github/workflows/npmpublish-beta.yml
Supereg: > Ideally the publish wouldn't go through unless all of those tests pass

This automatically covered by the `needs: build`

donavanbecker:
@dgreif and @Supereg does it look good now?

.github/workflows/npmpublish-beta.yml
donavanbecker: okay I have fixed this.

donavanbecker: okay I have fixed this.

.github/workflows/npmpublish-beta.yml
```diff
@@ -10,12 +10,17 @@ on:
 
 jobs:
   build:
+    
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 13.x]
```
dgreif, line 16: There is no value in having `matrix` in the publish files since we only want to use one node version

donavanbecker: okay I changed them to be as before but for node 10.

dgreif: Why node 10?  Node 12 is latest LTS.  That only impacts what version of node the github action is run with, not what version of node is supported on the users end

Supereg:
Can we please squash when merging 🌚🙃

PR #2480, opened 2020-04-09
Enabling support to configure Controllers (CameraController, RemoteController)

Supereg:
This PR enables API access to new configureController method provided by HAP-NodeJS.
This should enable access to the redesigned CameraController and automatically provide support for bridge cameras.

### DIFF
--- a/src/platformAccessory.ts
+++ b/src/platformAccessory.ts
@@ -1,8 +1,8 @@
 import { EventEmitter } from "events";
 import {
   Accessory,
-  AccessoryEventTypes,
-  Categories,
+  AccessoryEventTypes, CameraController,
+  Categories, Controller, ControllerConstructor,
   LegacyCameraSource,
   SerializedAccessory,
   Service,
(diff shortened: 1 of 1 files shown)


dgreif:
@Supereg thanks for getting this added!!  Is there any documentation for a migration path for existing plugins?  I can give this a try in the `homebridge-ring` plugin, but want to make sure I don't break everything for my users

Supereg:
Documentation is currently a bit lacking. But will definitely work on providing some when this is getting released.
In order to use bridged cameras you would need to move away from the current way of exposing a external accessory (external camera accessory) and just expose a the PlatformAccessory using the regular style of registering PlatformAccessory behind the bridge (meaning you would need to adopt the regular style of dynamic Platform plugins). And then you could use configureCameraSource as usual (this will break in previous homebridge releases though). How an old style CameraSource is still handled is explained [here](https://github.com/homebridge/HAP-NodeJS/blob/df6e36e48b3913655bfb36195c209f16d650c896/src/lib/Accessory.ts#L519-L541) in more detail.
If you want to go one step further you can leave the legacy CameraSource implementation behind and interface with the CameraController and "configureController" directly. It has a relatively similar API with some refinements to the callbacks and types, as well as removing the need to create StreamControllers by yourself. An example of such an implementation can be found [here](https://github.com/homebridge/HAP-NodeJS/blob/beta/src/accessories/Camera_accessory.ts). This example is written to interface with hap-nodejs directly, so adjust where needed (Accessory creating and publishing).

Hope this helps to get you started. If you need any more or detailed information, just ask, happy to help, and would also give some insight when creating the documentation.

PR #2494, opened 2020-04-14
Automatically pass the current plugin version to the FirmwareRevision characteristic

Supereg:
As the title already says, the current plugin version is now automatically set as the `FirmwareRevision` value for external accessories, platform accessories and normal accessory plugins. If a developer wishes to expose another versioning system they are advised to fall back to the `SoftwareRevision ` or `HardwareRevision` characteristic.

Additionally the plugin class was optimized to never store the content of the package.json forever and hap-nodejs was updated to the latest beta.

No changes where made to the platformAccessory (except of one type correction). I only fixed the indents there to be 2 spaces. ESLint seems to be somewhat weird or not as aggressive for classes lol.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2254,9 +2254,9 @@
       "dev": true
     },
     "fast-srp-hap": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/fast-srp-hap/-/fast-srp-hap-1.0.1.tgz",
-      "integrity": "sha1-N3Ek0Za8alFXquWze/X6NbtK0tk="
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/fast-srp-hap/-/fast-srp-hap-2.0.1.tgz",
+      "integrity": "sha512-dClwnyfRd3BZxu3KAdhvAayozq7fLazXGlDc4HAHJV1M+olqGKAT52pygXQu5UiDSHxz/WB3KRvsojqQ5zmXOA=="
     },
     "fb-watchman": {
       "version": "2.0.1",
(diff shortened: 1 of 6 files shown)


oznu:
Can plugin developers overwrite FirmwareRevision if they need to?

Supereg:
Currently not. That is currently also the intention. Don't know what's more important, consistency or in some way, let's call it, "backwards compatibility".

Supereg:
Optionally we could also introduce so step up our game with default values. "Manufacturer" could be set to the plugin name and "model" to the accessory name/ platform name (except for external accessories, there we have to have some weird default "model". But for those I would suggest to give the opportunity to overwrite it.

Supereg:
But would also favor to force those values. Could be helpful in the future for some sort of Managing App on iOS or something to have a nice way to get plugin names etc.

oznu:
The value set in firmware value does not effect the "Firmware Update Available" service does it?

> Could be helpful in the future for some sort of Managing App on iOS or something to have a nice way to get plugin names etc.

It's probably a long way off, but if we had our own app then we could use custom characteristics to store this sort of info.

Supereg:
> The value set in firmware value does not effect the "Firmware Update Available" service does it?

What service are you talking about?

oznu:
Like these sort of issues where Apple thinks an update is available for an accessory when it's just the Firmware version in the plugin not matching. Granted this would be a pretty rare scenario as the device (or similar models of the device) would need native HomeKit support already and the developer would need to have set the manufacturer and model numbers correctly.

https://www.reddit.com/r/homebridge/comments/7wp4k9/lifx_z_firmware_update/?utm_source=amp&utm_medium=&utm_content=post_body

https://www.reddit.com/r/homebridge/comments/fu3605/homebridge_device_showing_update_available/

Supereg:
Hm. Don't know how much of an issue this is. Never experienced it. But we can make this change also overridable (It definitely is, just with some workarounds, timers and stuff).
I don't really know how the Home App update notification procedure works exactly, only that it has nothing to do with the accessory and it is some on device magic.

NorthernMan54:
Sounds great, I have been coding mine to do this for a while

Supereg:
Do we now want to "force" FirmwareRevision to be the plugin version or should we respect values passed by the dev? @oznu

oznu:
I'm happy to try this out in beta and see if anyone raises issues with it.

ebaauw:
I prefer to respect the values passed by the plugin.  Happy to default values if not set by the plugin, but plugin should have the final say.

I need to check the firmware revision, but I try and expose accessories similar to the way the Hue bridge exposes them.  So they take the model, manufacturer, serial number, and hardware, software, and firmware revisions from the device instead of from the plugin.

PR #2496, opened 2020-04-14
Fix Windows @scoped plugin resolution

oznu:


src/pluginManager.ts
```diff
@@ -80,7 +80,7 @@ export class PluginManager {
   }
 
   public static isQualifiedPluginIdentifier(identifier: string): boolean {
-    return PluginManager.PLUGIN_IDENTIFIER_PATTERN.test(identifier);
+    return PluginManager.PLUGIN_IDENTIFIER_PATTERN.test(identifier.replace(path.sep, "/"));
```
oznu, line 83: Windows will use `\` in paths while the npm package names will always use `/` as the scope seperator. Package names cannot contain `\` so this won't conflict with anything else.

src/pluginManager.ts
```diff
@@ -285,6 +285,7 @@ export class PluginManager {
 
             const absolutePath = path.join(searchPath, scopeDirectory);
             fs.readdirSync(absolutePath)
+              .filter(name => PluginManager.isQualifiedPluginIdentifier(name))
```
oznu, line 288: Scoped packages can only be one level deep, this saves having to stat every directory in the @scope that cannot possibly be homebridge plugins.

src/pluginManager.ts
Supereg: I don't like the fact that we put that in to the isQualifiedPluginIdentifier function. And the backslash part is only really used in the context of plugin loading. Maybe we could add the replace just at the places this function is called in context of dealing with paths?

Supereg:
Also the windows path style breaks the active plugin checks some lines below:
``` js
     && (!this.activePlugins || this.activePlugins.includes(pluginIdentifier)); // check if activePlugins is restricted and if so if the plugin is contained
```

src/pluginManager.ts
```diff
@@ -285,8 +285,9 @@ export class PluginManager {
...
               .filter(name => fs.statSync(path.resolve(absolutePath, name)).isDirectory())
-              .forEach(name => relativePluginPaths.push(path.join(scopeDirectory, name)));
+              .forEach(name => relativePluginPaths.push(scopeDirectory + "/" + name));
```
Supereg, line 290: By not using `path.join` and assembling the plugin by ourself and forcing a `/` we remove the need to have the replace thing in the PluginManager.isQualifiedPluginIdentifier function and also restore functionality for our activePlugins.includes check.

Supereg:
Let me now what your opinion is on my proposed change. @oznu

PR #2501, opened 2020-04-17
Minor API adjustments

Supereg:
* Added some documentation for the basic API types
* Fixed the return type of the PlatformPluginConstructor
* Renamed 'LegacyPlatformPlugin' to 'StaticPlatformPlugin'. This type is still widely used for plugins who's accessories do not change over the lifespan of the plugin. I think a name which contains 'legacy' is not appropriate here, as it turns out to be still valid API.
* Added 'IndependentPlatformPlugin' for use by platforms which do not intend to add accessories to the bridge
* Additionally the api is now also passed to an accessory plugin constructor
* Some naming inside `plugin.ts` was adjusted to better reflect the behavior

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -22,42 +22,109 @@ export enum PluginType {
   PLATFORM = "platform",
 }
 
+/**
+ * The {PluginInitializer} is a method which must be the default export for every homebridge plugin.
+ * It is called once the plugin is loaded from disk.
+ */
 export interface PluginInitializer {
 
+  /**
+   * When the initializer is called the plugin must use the provided api instance and call the appropriate
+   * register methods - {@link API.registerAccessory} or {@link API.registerPlatform} - in order to
+   * correctly register for the following startup sequence.
+   *
+   * @param {API} api
+   */
   (api: API): void;
 
 }
 
 export interface AccessoryPluginConstructor {
-  new(logger: Logging, config: AccessoryConfig): AccessoryPlugin;
+  new(logger: Logging, config: AccessoryConfig, api: API): AccessoryPlugin;
 }
 
 export interface AccessoryPlugin {
 
+  /**
+   * Optional method which will be called if a 'identify' of a Accessory is requested by HomeKit.
+   */
   identify?(): void;
 
+  /**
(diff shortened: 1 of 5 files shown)


src/api.ts
```diff
@@ -152,8 +219,11 @@ export class HomebridgeAPI extends EventEmitter implements API {
     super();
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public static isLegacyPlatformPlugin(platformPlugin: any): platformPlugin is LegacyPlatformPlugin {
+  public static isDynamicPlatformPlugin(platformPlugin: PlatformPlugin): platformPlugin is DynamicPlatformPlugin {
+    return "configureAccessory" in platformPlugin;
+  }
+
+  public static isLegacyPlatformPlugin(platformPlugin: PlatformPlugin): platformPlugin is StaticPlatformPlugin {
```
oznu, line 226: should this be renamed to `isStaticPlatformPlugin` as well?

Supereg: oh yes, missed that, thanks

Supereg:
rename `isLegacyPlatformPlugin` to `isStaticPlatformPlugin`

src/api.ts
```diff
@@ -22,42 +22,109 @@ export enum PluginType {
...
 
 }
 
-export interface LegacyPlatformPlugin {
+/**
+ * Platform that exposes all available characteristics at the start of the plugin.
+ * The set of accessories can not change at runtime.
+ * The bridge waits for all callbacks to return before it is published and accessible by HomeKit controllers.
+ */
+export interface StaticPlatformPlugin extends PlatformPlugin {
```
jonavila, line 93: 🙌🏻

Supereg: This is the interface you were searching for? This PR also fixes the return type of the PlatformPluginConstructor type, which previously only allowed to return a dynamic platform 😅

jonavila: Yup, this should mix the type error that I get now. I'll keep an eye on NPM for the new beta with these changes.

Supereg: beta 45 should be released

PR #2502, opened 2020-04-17
Provide the possibility to override the FirmwareRevision again

Supereg:
I think I changed my standpoint on this. I think it's helpful for developers that we set the default to their plugin version, but for those who wanna mirror the firmware of the device they are bridging, it's the better choice.
@ebaauw @oznu

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -276,10 +276,12 @@ export class Server {
           return false; // filter it from the list
         }
       } else {
-        // we always use the last registered
-        platformPlugins[0].configureAccessory(accessory);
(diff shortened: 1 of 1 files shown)


Supereg:
We currently update the FirmwareRevision every time we load cachedAccessories from disk (so a possible plugin version increase is correctly displayed). So in order to still provide your own FirmwareRevision you will need to override it every time in the `configureAccessory` method.

ebaauw:
Not happy.  I don't mind changes, but I do mind surprises.  Can we please document all changes in .54 in one place?

PR #2522, opened 2020-04-27
Do not prevent Homebridge loading plugins based on Homebridge engines req.

oznu:
This is a temporary solution to the issue many people are facing and unfortuately wasn't caught in the beta due to our version numbers. While this might not be an ideal solution long-term it will reduce the amount of immediate issues our users face and allow us to focus on other issues.

Users will still get a RED error message:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "homebridge",
-  "version": "0.4.53",
+  "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


Supereg:
Could we please incorporate the actual version into the commit message? meaning a classic npm version commit?

oznu:
Done.

PR #2527, opened 2020-04-27
Improve cached accessory resolution 

Supereg:
This PR tries to improve the plugin resolution of cached accessories by trying to find a plugin just by the associated platform name.

This change was done in order to mitigate issues arising when dynamic platform plugins decided to change the plugin identifier. When the associated plugin for a given cached accessory is not found, homebridge will retry by searching for plugins which have registered the associated platform name. If the result is unique the ownership of the accessory is transferred.

This basically mimics the way how platform configurations are resolved (with the addition of updating the associatedPlugin property)

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -126,8 +126,10 @@ export class Plugin {
     platforms.unshift(platformPlugin);
   }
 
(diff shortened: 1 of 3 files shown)


nicoduj:
@Supereg thank you very much for this change that will mitigate the issue (and my mistake) on the pluginName on harmony plugin ... And great work by the way !

Issue #2520, opened 2020-04-27
Can’t add hombridge to Apple Home on 1.0.0


thevisualist:
# 💡 please read this: https://github.com/homebridge/homebridge/issues/2520#issuecomment-619919265

---

Can‘t add Homebridge to Apple Home after upgrading to 1.0.0

Home app discovers Homebridge but then promts
That it can‘t establish a connection to the device.

Node 12.16.1
NPM 6.14.3

glowfishDE:
Same here. After upgrading to 1.0.0 all connections to any device are broken (device not responding)

GreatCookie:
Me Too - Running on Synology Docker. All devices come back on Homekit (iOS Devices) as "No Response" (although Cameras are working). Everything working fine via Homebridge UI. 
Will probably downgrade first to 0.4.53 again first.

armandoi75:
Same problem, 
homebridge with fresh install, I can add the HUB but accessories not responding
node -v
v12.16.2
npm -v
6.14.4

oznu:
Hi Everyone.

If you're facing this issue, can you please help us out by posting debug logs here?

Start Homebridge with the environment variable `DEBUG=EventedHTTPServer,HAPServer,Accessory`, and post the full log from a  Homebridge restart to you attempting to open the iOS Home app and control the accessories.

### hb-service users (and up-to-date Docker users)

If you're using `hb-service` to manage Homebridge (or a recent Docker image) you can set the `DEBUG` variable in the Homebridge UI by clicking the drop down menu in the upper-right-corner, selecting *Homebridge Settings* and enter `EventedHTTPServer,HAPServer,Accessory`. Then restart Homebridge.

![image](https://user-images.githubusercontent.com/3979615/80366645-f0a0d700-88cc-11ea-9311-eb47b60b92f3.png)

### systemd users

If you are using `systemd` without `hb-service` open the `/etc/default/homebridge` file and edit or create the `DEBUG` line to match this (make sure you uncomment it). Once done, restart Homebridge.

`DEBUG=EventedHTTPServer,HAPServer,Accessory`

nin9s:
`Enter this code with your HomeKit app on your iOS device to pair with Homebridge:`
                       
    ┌────────────┐     
    │ 308-32-619 │     
    └────────────┘     
                       
```
2020-04-27T11:55:23.587Z EventedHTTPServer Server listening on port 51826
[...]

Issue #2518, opened 2020-04-27
Plugins not responding after upgrading to v1.0.0


codyc1515:
# 💡 please read this: https://github.com/homebridge/homebridge/issues/2518#issuecomment-619919562

---

**Describe The Bug:**
I have a single accessory which shows air quality installed that I wrote myself.

**To Reproduce:**
 Upgrading to v1.0.0 causes the plugin to go to "No response" state.

**Expected behavior:**
 Downgrading back to v0.4.53 brings back the accessory again.

**Logs:**
I cannot see any obvious reason why from the logs. https://gist.github.com/codyc1515/d3605c77afb729bae5418ef076a329a6

**Environment:**

* **Node.js Version**: v12.16.1n
* **NPM Version**: 6.13.4
* **Homebridge Version**: v1.0.0
* **Operating System**: Raspbian
* **Process Supervisor**: hb-service

netRunner0:
I support
My list does NOT work with:
homebridge-hue
homebridge-camera-ffmpeg v0.1.17
homebridge-cmdswitch2 v0.2.10
homebridge-http-switch v0.5.29
homebridge-mi-hygrothermograph v3.0.0
homebridge-mqtt-fan-tasmota v0.0.5
homebridge-mqtt-power-consumption-log-tasmota v0.9.0
homebridge-mqttthing v1.1.10
homebridge-webos-tv v1.7.1
homebridge-xbox-tv v0.1.54
homebridge-xiaomi-roborock-vacuum v0.7.6
homebridge-yamaha-avr v1.0.3
homebridge-yamaha-home v0.1.9

glowfishDE:
Same here:

Got the error that said that some plugin that is registered (in config.json) is not available. However it´s installed. I had to remove it from the config.json so homebridge would start up.
 
My plugin that prevented the homebridge startup because it was suddenly not working anymore was:

Homebridge Fritz (https://github.com/andig/homebridge-fritz#readme)

GreatCookie:
Me Too - Running on Synology Docker. All devices come back on Homekit (iOS Devices) as "No Response" (although Cameras are working). Everything working fine via Homebridge UI.
Will probably downgrade first to 0.4.53 again first.

Could #2520 be related as well?

glowfishDE:
Thanks ! Yes, i also have it on a synology docker container (OZNU Image).

I´ve downgraded to 0.4.53 and everything works again as before. Thank you :)

[...]

PR #2536, opened 2020-04-28
Preparing release of 1.0.3

Supereg:
The version contains the following changes.

* Added check that we don't publish an empty accessory
* Bumped hap-nodejs to v0.6.8
  * Bonjour address resolution was improved
  * Introduce a fix for #2518 and #2520: Added a fix where in some circumstances (when a plugin supplied empty values for mandatory AccessoryInformation service characteristics) the instance would become unresponsive.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2502,9 +2502,9 @@
       "optional": true
     },
     "hap-nodejs": {
-      "version": "0.6.6",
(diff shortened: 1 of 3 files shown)


Supereg:
maybe we will wait with publishing. Got a new setup and this seem to have a different issue

PR #2549, opened 2020-04-30
Release of 1.0.4

Supereg:
* Fixed a crash that could occur if a plugin called `updateReachability` before the accessory was added to the bridge (https://github.com/devbobo/homebridge-arlo/issues/40#issuecomment-620928214)
* Fixed a crash that could occur while pairing when running plugins (like homebridge-nest) which register a AccessoryInformation service that already has added a Identify listener of HAP-NodeJS (https://github.com/homebridge/homebridge/issues/2548)
* Fixed mdns advertising to include all (and only) reachable public addresses for the given machine

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "homebridge",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


NorthernMan54:
Looks good

PR #2565, opened 2020-05-08
Adding plugin version to the log output when a plugin was loaded

Supereg:
Prints now something like the following on startup
```
Loaded plugin: homebridge-example@2.1.4
```


NorthernMan54:
Two thumbs up

Supereg:
one of my bigger PRs lul 
6+ hours went into this

PR #2570, opened 2020-05-10
Adding support for plugins to expose their own (external) bridge accessories

Supereg:
* Raises api version to `2.6`
* Added `platformBridge` field to the api class
* Added `PlatformBridge` class (inheriting from `PlatformAccessory`)

A `PlatformBridge` can only be published via `publishExternalAccessories` (and you would only call that for the main bridge accessory).

A `PlatformBridge` has the method `addBridgedAccessory` which can be used to add `PlatformAccessory` instances to the bridge (before and after being published).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file. This projec
 
 ### Notable Changes
 
+* Bumped the api version to `2.6` with the following changes:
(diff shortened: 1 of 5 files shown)


oznu:
Nice work.

Does this handle cached accessories?

Supereg:
Nope. As cached accessories are only for accessories behind the main bridge and those will and can never be behind the main bridge obviously.

PR #2646, opened 2020-08-12
Fix @scoped plugin registration to avoid warnings

oznu:
This fixes warnings such as:

```
A platform configure a new accessory under the plugin name '@oznu/homebridge-plugin-name'. However no loaded plugin could be found for the name!
```

and:

```
When searching for the associated plugin of the accessory 'Wifi fan' it seems like the plugin name changed from '@milo526/homebridge-tuya-web' to '@milo526/homebridge-tuya-web'. Plugin association is now being transformed!
```

### DIFF
--- a/src/pluginManager.ts
+++ b/src/pluginManager.ts
@@ -340,7 +340,7 @@ export class PluginManager {
     }
 
     const plugin = new Plugin(name, absolutePath, packageJson, scope);
-    this.plugins.set(name, plugin);
+    this.plugins.set(packageJson.name, plugin);
     return plugin;
   }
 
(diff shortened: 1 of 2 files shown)


src/pluginManager.ts
```diff
@@ -340,7 +340,7 @@ export class PluginManager {
     }
 
     const plugin = new Plugin(name, absolutePath, packageJson, scope);
-    this.plugins.set(name, plugin);
+    this.plugins.set(packageJson.name, plugin);
```
Supereg, line 343: may use the `identifier` variable above? Not really important just a very minor aesthetic thing (as we also use above `const alreadyInstalled = this.plugins.get(*identifier*);`

oznu: Good point. Done.

PR #2664, opened 2020-09-01
Add abitity to type an accessory context using generics

mgcrea:
For platform with many accessories leveraging an uniform context, it a bit cumbersome to always cast the proper typing:

eg. `const {deviceId, endpointId} = context as TydomAccessoryContext;`

This PR adds an optional type generic (without breaking changes) so that you can do something like below where you don't have to re-cast the types.

```ts
export const setupContactSensor = (accessory: PlatformAccessory<TydomAccessoryContext>, controller: TydomController): void => {
  const {context} = accessory;
  const {deviceId, endpointId} = context;
```

### DIFF
--- a/src/platformAccessory.ts
+++ b/src/platformAccessory.ts
@@ -13,13 +13,13 @@ import {
 } from "hap-nodejs";
 import { PlatformName, PluginIdentifier, PluginName } from "./api";
 
-export interface SerializedPlatformAccessory extends SerializedAccessory {
+export type UnknownContext = Record<string, unknown>;
+
+export interface SerializedPlatformAccessory<T extends UnknownContext = UnknownContext> extends SerializedAccessory {
 
   plugin: PluginName;
   platform: PlatformName;
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


Supereg:
Does changing from type any to type unknown for the value of the context object break anything?

mgcrea:
I don't think there is, `unknown` is a better `any` regarding type-safety but works the same way. Only breaking change is that you can't assign an unknown typed value to another type (without casting with `as`), while with `any` there was no type-safety at all. Since the Context type wasn't even exported before I don't think anyone would notice the change.

I you want to err on the cautious side, I can update the PR and switch back to any.

Supereg:
> I you want to err on the cautious side, I can update the PR and switch back to any.

Nah, I would prefer the change anyways. Just checking if we need to put some breaking change information into the changelog for the devs. Doesn’t impact any running systems anyways. Only compile time issues could potentially arise.

Supereg (approved):

Supereg:
scheduled changed. This feature will be pulled into v1.2.0
Including it in the current beta now.

Supereg:
now released with v1.2.0

PR #2684, opened 2020-09-16
Filter out broken symlinks in node_modules folder

### LINKED ISSUE #2678: ENOENT no such file or directory
Hi, the installation works fine, but the homebridge is not up because this error.
What should i do?

```
[14.9.2020, 15:52:37] Error: ENOENT: no such file or directory, stat '/usr/lib/node_modules/JStillery_Server'
    at Object.statSync (fs.js:1042:3)
    at /usr/lib/node_modules/homebridge/src/pluginManager.ts:291:38
    at Array.filter (<anonymous>)
    at /usr/lib/node_modules/homebridge/src/pluginManager.ts:291:12
    at Set.forEach (<anonymous>)
    at PluginManager.loadInstalledPlugins (/usr/lib/node_modules/homebridge/src/pluginManager.ts:277:22)
    at PluginManager.initializeInstalledPlugins (/usr/lib/node_modules/homebridge/src/pluginManager.ts:116:10)
    at Server.start (/usr/lib/node_modules/homebridge/src/server.ts:150:24)
    at cli (/usr/lib/node_modules/homebridge/src/cli.ts:80:10)
    at Object.<anonymous> (/usr/lib/node_modules/homebridge/bin/homebridge:17:22)
```

oznu:
This fixes #2678.

I did think about doing various checks to see if the item is a symlink, then making sure then symlink target exists, but it's probably faster, and will cover more edge cases, to just `try...catch` the `fs.stat` call.

### DIFF
--- a/src/pluginManager.ts
+++ b/src/pluginManager.ts
@@ -15,7 +15,6 @@ import fs from "fs";
 import { Plugin } from "./plugin";
 import { Logger } from "./logger";
 import { execSync } from "child_process";
-import { dir } from "console";
 
 const log = Logger.internal;
 
--- a/src/pluginManager.ts
+++ b/src/pluginManager.ts
@@ -15,6 +15,7 @@ import fs from "fs";
 import { Plugin } from "./plugin";
 import { Logger } from "./logger";
 import { execSync } from "child_process";
+import { dir } from "console";
 
 const log = Logger.internal;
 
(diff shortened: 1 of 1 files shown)


Supereg:
Maybe at least print some debug warnings about finding broken symlinks (maybe including error.message)? Not a fan of silenced catch clauses.

oznu:
Done - added debug logs. I don't think we need to do it as a warning level here, as I don't think it's going to be caused by Homebridge plugins.

Supereg:
👏🏽

Issue #2698, opened 2020-10-03
Debug logging while debugging is disabled


milo526:
In my [homebridge-tuya-web](https://github.com/milo526/homebridge-tuya-web) plugin I created some [helper functions to aid in debugging](https://github.com/milo526/homebridge-tuya-web/blob/master/src/accessories/BaseAccessory.ts#L264).

These helper functions make use of the `platform.log.log(level, message, args)` call.
It turns out that calling `platform.log.debug(msg)` [properly disables itself](https://github.com/homebridge/homebridge/blob/0849c7711eb8dd7acc421f65e633bf98813d2aec/src/logger.ts#L143) when debugging is disabled, [the same however does not apply](https://github.com/homebridge/homebridge/blob/0849c7711eb8dd7acc421f65e633bf98813d2aec/src/logger.ts#L161) to `platform.log.log('debug', message, args)`

```javascript
platform.log.debug("This does not get printed with debug disabled");
platform.log.log("debug", "This does get printed with debug disabled.")
```

If this is not by design, I would like to create a PR to fix this.  
If this is by design I will alter my plugin to use the debug call directly.

Supereg:
Thanks for pointing this out. Will be fixed with the next release.

Supereg:
Fix was added to the beta channel and released as `v1.3.0-beta.6`. Thanks for reporting.

PR #2732, opened 2020-11-29
[BUG] or [FEATURE] Logger.log(LogLevel.DEBUG, msg) ignores Logger.debugEnabled?

johannrichard:
There is an apparent inconsistency in the `Logger.ts`. I'm not sure whether this is a bug (I'd think so) or a feature:

- when calling`Logger.debug(message)` directly, Logger will check whether `Logger.debugEnabled` is set or not; however
- when calling `Logger.log(LogLevel.DEBUG, message)`, no check is made

Obviously, under certain circumstances, this can lead to a situation where we will see debug messages in the log despite `Logger.debugEnable` not being true. I wonder whether this is intentional or not. 

If it's indeed a bug, then I'm happy to propose a fix in this PR. If it's a feature then that's OK as well. I haven't found anyone else asking for a fix / change for that so I hope its not something that's been discussed at length before.

The relevant pieces of code: 

https://github.com/homebridge/homebridge/blob/3c9fe900e7f538b68f2737a12ae5bb91146217c4/src/logger.ts#L142-L176


Supereg:
Duplicates https://github.com/homebridge/homebridge/issues/2698
This fix is part of the current beta.
https://github.com/homebridge/homebridge/blob/4cfb22733cd20163f2b9342e8fe21d3a69d22c21/src/logger.ts#L142-L150

johannrichard:
> Duplicates #2698
> This fix is part of the current beta.

Thanks, and apologies for the noise. Didn't have the `beta` on my radar yet.

PR #2772, opened 2021-01-10
Create Dependabot config file

naveensrinivasan:
Found a  library that has security vulnerabilities. Dependabot specifically looks for those and sends a PR to the repo.

### Dependabot checks for updates
Dependabot pulls down your dependency files and looks for any outdated or insecure requirements.

### Dependabot opens pull requests
If any of your dependencies are out-of-date, Dependabot opens individual pull requests to update each one.

### Dependabot setting has to be enabled like this
image

Here is an example PR https://github.com/fun-hacking-code/homebridge/pull/4 that has issues that was opened by dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10


Supereg:
We have Dependabot enabled for homebridge repositories. Why is there a need for a dedicated config file? Also, are these defaults values you have chosen? Or is there any specific intention behind those? Otherwise I think, having dependedabot enabled should be enough right?

naveensrinivasan:
These were standard config. The goal would be as you are using dependabot you would need things more like this https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#about-the-dependabotyml-file which can be configured using the config file.

donavanbecker:
Closing to go with #2829

PR #2792, opened 2021-02-03
Load plugin paths only when required

oznu:
`loadDefaultPaths()` includes an exec call to run `npm -g prefix` which is not required and slows things down when launching child bridges.

This PR moves `loadDefaultPaths()` so it's called immediately before it is required in `loadInstalledPlugins()`.


Supereg (approved):
Looks good, we could also just remove the `searchPath` property from the plugin manager? Or is there a need to store those plugin paths "forever"?

oznu:
I'll keep it there for now as we add the custom plugin path to the set in the constructor.

Issue #2753, opened 2020-12-18
Devices revert to "default room" and automations break on Homebridge ui update


JeremyLaurenson:
**Describe The Bug:**
Frequently when I apply an update, homebridge "moves" devices back tot he default room, and breaks my automations. I believe theyre being removed and re-added, especially based not he log.

**To Reproduce:**
Update UI plugin

**Expected behavior:**
Dont break my stuff.

[**Logs:**]
<details>
<summary>Logs (Click to expand)</summary>

```shell
[12/18/2020, 10:32:37] [Ring] Far Right Detected Motion. Loading snapshot before sending event to HomeKit
[12/18/2020, 10:32:39] [Ring] Front yard Detected Motion. Loading snapshot before sending event to HomeKit
[12/18/2020, 10:33:22] [Ring] Jeremy Detected Motion
[12/18/2020, 10:39:18] [Config] Running Command: sudo -E -n npm install homebridge-config-ui-x@latest
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:circle'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:circle' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:triangle'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:triangle' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:square'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:square' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:minus'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:minus' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:cross'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:cross' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'b:fob:circle'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'b:fob:circle' is 'true'
[12/18/2020, 10:39:32] [HttpWebHooks] Getting current state for 'j:fob:plus'...
[12/18/2020, 10:39:32] [HttpWebHooks] State for 'j:fob:plus' is 'true'
[...]

Issue #2697, opened 2020-09-30
Custom Plugin: Accessories Resetting


Vortec4800:
**Describe Your Problem:**
I've written a custom plugin and it seems to work well. However, once every few days the custom accessory exposed by my plugin resets the room/favorite/status switches back to the defaults and it gets removed from any automations it was in. It's like HomeKit thinks its a new accessory.

I haven't figured out a pattern as to when this happens, and I haven't seen anything in the logs when it does either. This is really a plugin development issue but I don't know where the best place is to get support for that. Can you point me in the right direction?

**Environment:**

* **Node.js Version**: v12.18.2
* **NPM Version**: v6.14.7
* **Homebridge Version**: 1.2.3
* **Operating System**: Official Pi Image

NorthernMan54:
Pls share a link to your github repo with the source code.

Typically this is caused by accessories being deleted and added again in the plugin.

Vortec4800:
Sure, here it is:

https://github.com/Vortec4800/homebridge-ecobee-away

NorthernMan54:
That is about 2 months old....

Vortec4800:
I don't understand. I wrote it 2 months ago?

NorthernMan54:
Is that your latest code?

Vortec4800:
Yes, aside from this reset bug everything works fine and I haven't had to make any changes.

NorthernMan54:
My initial thought was device discovery removing and adding the accessory again or the service constructor creating new devices occasionally, but your code looks good.

Is your .persist or .accessories directories or files being removed or ???

Vortec4800:
I also thought maybe it was a bug in my caching code, I used it basically unchanged from the plugin template and I thought maybe I had to add something else to handle an edge case or something, but I wasn't sure.

Not that I'm aware of, I'm using the Homebridge official Pi image and I don't have anything else installed on that Pi, it's dedicated to Homebridge only. I really only manage it from the Config-X interface and it doesn't lose power or anything unusual.

I'll see what I can do about monitoring the accessory more closely for the next few days and see if I can't find out when it happens, and try to grab logs from it. I'll report back if I can do that, otherwise let me know if you think of anything else to check or try.

[...]

PR #2796, opened 2021-02-05
Prevent unintentional cache resets

oznu:
This PR addresses an issue that could cause the Homebridge accessory cache to be deleted or overwritten unintentionally in some circumstances.

### Background:

Homebridge saves its accessory cache to disk every time a new platform accessory is registered, removed or updated.

We also save the accessory cache just before shutdown, after it receives a SIGTERM or SIGINT signal (except on Windows where shutdown signals are not a thing).

There is some code that deleted the cached accessories file if it had not yet been loaded, I'm not sure why this was needed historically - but it did mean the file could be deleted if Homebridge received a SIGTERM/SIGINT signal before it had loaded the cached accessories file.

### Solution:

This PR prevents the cached accessories file from being saved to disk if we have not yet attempted to load it. It also removes the unnecessary code that could delete the file.

### Issues:

#2697
#2753

### DIFF
--- a/src/bridgeService.ts
+++ b/src/bridgeService.ts
@@ -109,7 +109,7 @@ export class BridgeService {
   private readonly allowInsecureAccess: boolean;
 
   private cachedPlatformAccessories: PlatformAccessory[] = [];
-  private cachedAccessoriesFileCreated = false;
+  private cachedAccessoriesFileLoaded = false;
   private readonly publishedExternalAccessories: Map<MacAddress, PlatformAccessory> = new Map();
 
   constructor(
@@ -176,7 +176,7 @@ export class BridgeService {
     info.setCharacteristic(Characteristic.FirmwareRevision, getVersion());
 
     this.bridge.on(AccessoryEventTypes.LISTENING, (port: number) => {
-      log.info("Homebridge v%s is running on port %s.", getVersion(), port);
+      log.info("Homebridge v%s (%s) is running on port %s.", getVersion(), bridgeConfig.name, port);
     });
 
     // noinspection JSDeprecatedSymbols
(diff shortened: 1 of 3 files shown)


src/bridgeService.ts
```diff
@@ -270,14 +270,11 @@ export class BridgeService {
...
-        this.cachedAccessoriesFileCreated = true;
-
+      // only save the cache file back to disk if we have already attempted to load it
+      // this should prevent the cache being deleted should homebridge be shutdown before it has finished launching
+      if (this.cachedAccessoriesFileLoaded) {
         const serializedAccessories = this.cachedPlatformAccessories.map(accessory => PlatformAccessory.serialize(accessory));
         this.storageService.setItemSync(this.bridgeOptions.cachedAccessoriesItemName, serializedAccessories);
-      } else if (this.cachedAccessoriesFileCreated) {
-        this.cachedAccessoriesFileCreated = false;
-        this.storageService.removeItemSync(this.bridgeOptions.cachedAccessoriesItemName);
```
oznu, line 280: This is the main offending code. If `cachedAccessoriesFileCreated` was not true (as we had not created or loaded the cache yet) the file would be deleted on SIGTERM/SIGINT.

Supereg (approved):
The `cachedAccessoriesFileCreated` property was there to track if there was a file created on disk.
If we would encounter saving a cachedAccessories array of length zero, we would delete the file.
So it was really only there to do some cleanup, instead of storing an empty json array.

But I'm fine with removing that logic.

> This PR prevents the cached accessories file from being saved to disk if we have not yet attempted to load it.

👍

Supereg:
Would it maybe make sense to back port this fix to 1.1/1.2 line?

JeremyLaurenson:
Hey folks, happy to test this fix;
1... My config is in a state where I have new plugin updates to install which is whewre I potentially will run into the bug.
2... How do I know when its in the "official" beta accessible from the Homebridge UI so I can try update tot he beta first.
3... How can I "safely" shut down my Homebridge instance in order to avoide this and back up my variables first?

oznu:
Thanks @JeremyLaurenson 

> How do I know when its in the "official" beta accessible from the Homebridge UI so I can try update to the beta first, before plugin update?

It's merged so it's in the beta builds (this happens automatically).

> How can I "safely" shut down my Homebridge instance in order to avoide this and back up my variables first?

Do a backup first, this backs up cached accessories too: https://github.com/homebridge/homebridge/wiki/Backup-and-Restore

JeremyLaurenson:
Did the update; updated plugins with no negative side-effects. Looking good so far.

PR #2803, opened 2021-02-13
Replace mdns.legacyAdvertiser with bridge.advertiser enum

oznu:
As discussed with @Supereg.

Example:

```json
{
  "bridge": {
     "advertiser": "ciao"
  }
}
```

```json
{
  "bridge": {
     "advertiser": "bonjour-hap"
  }
}
```

Defaults to `bonjour-hap` when no value is provided, or when an invalid value is provided (displays warning in log if invalid value is provided).

Warns existing beta users that have `mdns.legacyAdvertiser` set to false this option is no longer valid.

---

If approved I'll update the UI in this weeks release cycle.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3090,9 +3090,9 @@
       }
     },
     "hap-nodejs": {
-      "version": "0.9.0-beta.123",
-      "resolved": "https://registry.npmjs.org/hap-nodejs/-/hap-nodejs-0.9.0-beta.123.tgz",
-      "integrity": "sha512-bZj9+5DaIIrxu5rWNvBPZqPjFC6dU/2z3Hcq/tJ7MHWzF8KNC0qOE1qWQCna2NTaGrWNOrMdqeag9RHw3LK+5Q==",
+      "version": "0.9.0-beta.126",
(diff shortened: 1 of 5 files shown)


Supereg:
Maybe it would make more sense to incorporate the enum configuration directly into the HAPNodeJS publish options. I'll prepare the change over there.

Supereg:
Introduced the `MDNSAdvertiser` enum and the `advertiser` `PublishInfo` with https://github.com/homebridge/HAP-NodeJS/commit/28ac11d860a532fc4ea4db518223cc3a9f8f3067.
Live with 0.9.0-beta.126

Supereg:
We may want to pull that check into homebridge to link to the homebridge specific config options (as you currently do in the PR)
https://github.com/homebridge/HAP-NodeJS/blob/28ac11d860a532fc4ea4db518223cc3a9f8f3067/src/lib/Accessory.ts#L1103-L1104

oznu:
Changes made. 

Users will never see the same warning in HAP as the useLegacyAdvertiser is not passed through when using Homebridge now.

PR #2915, opened 2021-04-28
Add support for ESM modules and async plugin initializers.

ShogunPanda:
# Name of the PR

Add support for ESM modules and async plugin initializers.

Today, if you try to load a package which `type=module` in package.json it throws an error as it tries to use `require()` on a ESM module.

## :bulb: Proposed solution

This PR introduces detection of ESM module and if detect it uses `await import()`.

### Implications

By using `await import`, by definition the entire plugin loading phase has become asynchronous. All affected code places have been refactored to be `async function` and to `await` for plugin loading and initialization.

Almost for free, HomeBridge plugins can now export a `async function` as default and this will be awaited. This means that now plugin can do async initialization.

### Reviewer Nudging

Plugin and PluginManager code.

### DIFF
--- a/src/pluginManager.ts
+++ b/src/pluginManager.ts
@@ -24,6 +24,7 @@ export interface PackageJSON { // incomplete type for package.json (just stuff w
   keywords?: string[];
 
   main?: string;
+  type?: "module" | "commonjs";
 
   engines?: Record<string, string>;
   dependencies?: Record<string, string>;
@@ -116,14 +117,14 @@ export class PluginManager {
     return identifier.split(".")[0];
   }
 
-  public initializeInstalledPlugins(): void {
+  public async initializeInstalledPlugins(): Promise<void> {
     log.info("---");
 
     this.loadInstalledPlugins();
 
-    this.plugins.forEach((plugin: Plugin, identifier: PluginIdentifier) => {
+    for(const [identifier, plugin] of this.plugins) {
       try {
-        plugin.load();
+        await plugin.load();
       } catch (error) {
         log.error("====================");
         log.error(`ERROR LOADING PLUGIN ${identifier}:`);
@@ -144,18 +145,18 @@ export class PluginManager {
         log.info(`Loaded plugin: ${identifier}@${plugin.version}`);
       }
 
-      this.initializePlugin(plugin, identifier);
+      await this.initializePlugin(plugin, identifier);
 
       log.info("---");
-    });
+    }
 
     this.currentInitializingPlugin = undefined;
   }
 
-  public initializePlugin(plugin: Plugin, identifier: string): void {
+  public async initializePlugin(plugin: Plugin, identifier: string): Promise<void> {
     try {
       this.currentInitializingPlugin = plugin;
-      plugin.initialize(this.api); // call the plugin's initializer and pass it our API instance
+      await plugin.initialize(this.api); // call the plugin's initializer and pass it our API instance
     } catch (error) {
       log.error("====================");
       log.error(`ERROR INITIALIZING PLUGIN ${identifier}:`);
--- a/src/childBridgeFork.ts
+++ b/src/childBridgeFork.ts
@@ -60,7 +60,7 @@ export class ChildBridgeFork {
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/pluginManager.ts
```diff
@@ -144,18 +145,18 @@ export class PluginManager {
...
 
     this.currentInitializingPlugin = undefined;
   }
 
-  public initializePlugin(plugin: Plugin, identifier: string): void {
+  public async initializePlugin(plugin: Plugin, identifier: string): Promise<void> {
     try {
       this.currentInitializingPlugin = plugin;
-      plugin.initialize(this.api); // call the plugin's initializer and pass it our API instance
+      await plugin.initialize(this.api); // call the plugin's initializer and pass it our API instance
```
Supereg, line 159: The await here has currently no effect(?), what was the intend here?

You said a homebridge plugin could now also export a async function.
Though looking into the `Plugin.initialize(api: API): void` (which is currently not a async function), the line `this.pluginInitializer(api);` isn't waited for(?), so seemingly async functions exported by the plugin aren't waited for, or am I missing something?

Supereg (commented):
Thanks for your PR.
I'm not sure about the support for async functions exported by a homebridge plugin, everything else looks good 👍 
Have you tested this with a ESM module?

src/pluginManager.ts
```diff
@@ -24,6 +24,7 @@ export interface PackageJSON { // incomplete type for package.json (just stuff w
   keywords?: string[];
 
   main?: string;
+  type?: "module" | "commonjs";
```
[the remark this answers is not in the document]
ShogunPanda: Sure, I'll do it soon!

src/pluginManager.ts
ShogunPanda: Good catch. I missed that function. 
Rather than awaiting, I think it's better if we do `return ...` in `https://github.com/homebridge/homebridge/blob/master/src/plugin.ts#L203` and thus the new signature should be `Plugin.initialize(api: API): Promise<unknown>`.

ShogunPanda:
Yup, I tested with a custom module and it works awesomely. The only thing I haven't tested is child bridges.

src/pluginManager.ts
ShogunPanda: Updated. PTAL.

src/pluginManager.ts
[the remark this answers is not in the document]
ShogunPanda: Done!

ShogunPanda:
Hello! Any chance for this to be merged?

Fish2:
any news of pulling this in ?

Supereg:
We have most likely an Issue related to this PR. Could one please investigate: https://github.com/homebridge/homebridge/issues/3013#issuecomment-974328394?

seydx:
Hey guys, does anyone else have the following problem on Windows?

```
Error: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:371:5)
    at defaultResolve (node:internal/modules/esm/resolve:1016:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:422:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:222:40)
    at ESMLoader.import (node:internal/modules/esm/loader:276:22)
    at importModuleDynamically (node:internal/modules/cjs/loader:1041:29)
    at importModuleDynamicallyWrapper (node:internal/vm/module:437:21)
    at importModuleDynamically (node:vm:381:46)
    at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14)
    at eval (eval at <anonymous> (C:\Users\Long\AppData\Roaming\npm\node_modules\homebridge\src\plugin.ts:24:24), <anonymous>:3:1)
```

ShogunPanda:
Most likely it will happen to everybody on Windows.
I don't have a Windows PC and only developed on MacOS or Linux.

seydx:
> Most likely it will happen to everybody on Windows.
> I don't have a Windows PC and only developed on MacOS or Linux.

Will post a PR later with a fix

PR #3016, opened 2021-11-22
Add support for package.json exports field.

ShogunPanda:

New packages in NPM (especially ESM one) might use the `exports` field in `package.json` rather than `main`. At the moment homebridge does not use the `exports` field.

## :bulb: Proposed solution

This PR added parsing of the `exports` field according to the documentation found [here](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_package_entry_points).

## :gear: Release Notes

User should not see any difference and actually should see less failures in the future.


src/plugin.ts
```diff
@@ -57,8 +57,32 @@ export class Plugin {
...
+
+    // figure out the main module
+    // exports is available - https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_package_entry_points
+    if (packageJSON.exports) {
+      // main entrypoint - https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_main_entry_point_export
+      if (typeof packageJSON.exports === "string") {
+        this.main = packageJSON.exports;  
+      } else { // subpath export - https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_subpath_exports
+        // conditional exports - https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_conditional_exports
+        const exports = packageJSON.exports.import || packageJSON.exports.require || packageJSON.exports.node || packageJSON.exports.default || packageJSON.exports["."];
```
Supereg, line 71: Also applying to the line below which handles condition exports.

I'm not quite sure if this simple handling for conditional exports will work?
The `exports.import` is used to supply the entry point for ESM modules, right? And `exports.require` is used to supply the "regular" require location.
Meaning, depending on the value we use, we would need to adjust the `isESM` property, do we?

While I'm not sure what the right decision here is, we should definitely create some documentation around `exports` field on https://developers.homebridge.io/ (and in the according template repos) when writing homebridge plugins.

E.g. https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_package_entry_points writes:
`
Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change.
`
This is required for homebridge plugins.
Can we detect such an error, when `package.json` wasn't included in the `exports` declaration, such that we can log an error explaining the situation?

src/pluginManager.ts
```diff
@@ -24,6 +24,7 @@ export interface PackageJSON { // incomplete type for package.json (just stuff w
   keywords?: string[];
 
   main?: string;
+  exports?: string | Record<string, string | Record<string, string>>
```
Supereg, line 27: Would be great if you could include the link to the docs https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_package_entry_points into the docs comment for the field.

bauer-andreas (commented):
Thanks for you contribution. 🚀 

PR looks good so far. Just had a little question in regard to condition exports where I'm not sure if I understand how node handles such situations.

src/plugin.ts
ShogunPanda: I'm not sure either on how to treat the package when both `exports.import` and `exports.require` are defined. I assumed that if a package uses both, then it works in both ESM and CJS. Since we support both, loading one of them does not make any difference and thus I stopped at first found.

About the warning about being not able to require "private" files, I don't think it specifically applies to homebridge. First of all, we load `package.json` via `fs.readFileSync`, for which this restriction does not apply (that one applies only to `require`), then we only load the main entrypoint of a package.

For package developers that try to include private files will receive an error from node, but I don't think we should have a doc for all possible error coming from using Node, isn't it? :)

src/pluginManager.ts
ShogunPanda: Done!

ShogunPanda:
Hello! Any update on when this is going to be merged?

ShogunPanda:
I've updated the branch. Any update for this to get merged?

ShogunPanda:
I've rebased the branch against the latest master. Any chance for this to be merged?

ShogunPanda:
@Supereg Any chance for this to land?

ShogunPanda:
@Supereg Up again (sorry, this is preventing me to update some packages) - Can we please merge this and release a new version?

src/plugin.ts
ShogunPanda: @Supereg @oznu Are we good on this? Can we land this PR?

Supereg: I'm not sure if my point was addressed yet?

> The exports.import is used to supply the entry point for ESM modules, right? And exports.require is used to supply the "regular" require location.
Meaning, depending on the value we use, we would need to adjust the isESM property, do we?


> "I'm not sure either on how to treat the package when both exports.import and exports.require are defined."

Not sure if you may misunderstood my point, I was actually not talking about the situation where both are defined, but e.g. when only `exports.import` is defined (implying a ESM module). Would this require us to set the `isESM` property (which currently is only set depending on the package.json `type` field)? Or is the `type` property set to `module` anyways?

Apart from that point, where I'm unsure myself on how to handle that situation (haven't yet encountered any packages using the `exports` property), I'm happy to land this PR.
Maybe @oznu has some expertise on this?

ShogunPanda: Oh, I see your point now. I think I misunderstood it.
No, `package.exports` is also available for CommonJS. Only `type` explicitly sets the type to be ESM.

Supereg: > No, package.exports is also available for CommonJS.

Yes that's true. However `exports.import` will point to a ESM module file. As of right now, we require some workaround to import ESM modules (see #2915 and https://github.com/microsoft/TypeScript/issues/43329).

To my understanding this needs to be handled. I may be wrong though.
Could you verify that this does indeed work? Or have you already tested it thoroughly?

ShogunPanda: I forgot to test te `exports.import` case. If it is set now I force `isESM` otherwise it will not work below.
The `await import` of the TS file works fine instead.

RobinMac:
Why are you still fucking with my head asshole

src/plugin.ts
ShogunPanda: @Supereg All clear now? Can I rebase so we finally land this?

bauer-andreas (approved):
LGTM 🚀 

Just created the 1.5.0 beta line and rebased the PR onto it.

src/plugin.ts
```diff
@@ -58,8 +58,37 @@ export class Plugin {
...
+
+        // check if conditional export is nested
+        if (typeof exports !== "string") {
+          if(exports.import) {
+            this.main = exports.import;
+            this.isESM = true;
```
aduh95, line 78: That's not at all how Node.js does it, `isESM` should be equals to `this.main.endsWith('.mjs') || (this.main.endsWith('.js') && packageJSON.type === "module")`.

ShogunPanda: I already do the packageJSON type check in line 61.
I could probably move the isESM check after the if opened on line 66. But are you sure that `exports.import` does not automatically assume ESM?

aduh95: I'm sure
```sh
mkdir repro
cd repro
echo '{"name":"myPackage","exports":{"import":"./entry.cjs"}}' > package.json
echo 'console.log(this === undefined ? "ESM" : "CJS")' > entry.cjs
node --input-type=module -e 'import "myPackage"'
cd ..
rm -r repro
```

PR #2973, opened 2021-09-11
Support HomeKit Secure Video

california444:
# Support HomeKit Secure Video

## :recycle: Current situation
Support for HomeKit Secure Video will be added to HAP NodeJS with [PR904](https://github.com/homebridge/HAP-NodeJS/pull/904).

- [ ]  **WORK IN PROGRESS until [PR904](https://github.com/homebridge/HAP-NodeJS/pull/904) gets merged **

## :bulb: Proposed solution
This PR exposes added types and classes for Secure Video from HAP to be used in homebridge plugins

### Problem that is solved
Provide a description and link issues that are solved

### Implications
*Describe the implications, e.g. refactoring*

## :heavy_plus_sign: Additional Information
*Provide some additional information if possible*

### Testing
*tested locally*

### Reviewer Nudging
*Where should the reviewer start, where is a good entry point?*

PR #3056, opened 2022-01-06
HomeKit Secure Video

bauer-andreas:

--

## :bulb: Proposed solution

This PR adds support for HomeKit Secure video by updating the beta to HAP-NodeJS `0.10.0-beta.7` and exporting the newly added types such that they are available to plugin developers. Refer to https://github.com/homebridge/HAP-NodeJS/pull/920.

## :gear: Release Notes

* Added support for HomeKit Secure Video

This PR replaces #2973 

### Testing

--

### Reviewer Nudging

--


gh0sti:
would ring doorbell work with this?

mitch7391:
I believe it won't ever, based on what I have read on Reddit (I think it is also detailed on the GitHub page somewhere too). It has something to do with how the footage is fetched from Ring servers and would be difficult to implement or slow in response. Also the developer of the Ring plugin does not want to implement it due to how it would affect the subscription service to Ring; which could make them change their mind on how supportive they have been of the plugin.

gh0sti:
Ring was supposed to come out with homekit integration which stinks that we have to get a plugin to do this implementation already. Thanks for the info.

PR #3070, opened 2022-01-20
Fix ES Module loading with abolute path

seydx:

Loading ES modules with absolute paths fails on Windows because of the url scheme.

The dynamic import needs a url-sheme with `file://..`, but on Windows it starts with `c:`, which breaks the whole loading.

```
Error: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:371:5)
    at defaultResolve (node:internal/modules/esm/resolve:1016:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:422:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:222:40)
    at ESMLoader.import (node:internal/modules/esm/loader:276:22)
    at importModuleDynamically (node:internal/modules/cjs/loader:1041:29)
    at importModuleDynamicallyWrapper (node:internal/vm/module:437:21)
    at importModuleDynamically (node:vm:381:46)
    at importModuleDynamicallyCallback (node:internal/process/esm_loader:35:14)
    at eval (eval at <anonymous> (C:\Users\Long\AppData\Roaming\npm\node_modules\homebridge\src\plugin.ts:24:24), <anonymous>:3:1)
```

## :bulb: Proposed solution

The PR will transform the absolute path into a "file url"

### Testing

Tested on windows and linux without any issues

### Reviewer Nudging

Plugin code


oznu (approved):

oznu:
@Supereg @seydx - should this target the 1.4 beta branch?

Supereg:
> @Supereg @seydx - should this target the 1.4 beta branch?

Would have included it in 1.4.1. 1.4.0 is frozen as release is imminent.

Supereg:
But if it’s super low risk I’m also open to include it last minute, if we merge it right away.

seydx:
> But if it’s super low risk I’m also open to include it last minute, if we merge it right away.

I do not see what speaks against it :)

longzheng:
I ran into this issue with `homebridge-camera-ui` on my Windows machine and @seydx figred out the solution. I can confirm the fix works.

@Supereg I would suggest this is a good fix to get out as soon as possible because otherwise plugins using ES Modules are essentially not usable on Windows.

Supereg (approved):

Issue #2771, opened 2021-01-06
AccessoryPlugin interface is missing name property


Patrick-Remy:
**Describe The Bug:**
TypeScript interface [AccessoryPlugin](https://github.com/homebridge/homebridge/blob/master/src/api.ts#L51-L81) is missing the required `name` property. This property is accessed on [platform creation](https://github.com/homebridge/homebridge/blob/master/src/server.ts#L426) and suppressed with `@ts-ignore`. In the examples it is marked as [private here](https://github.com/homebridge/homebridge-examples/blob/master/accessory-example-typescript/src/accessory.ts#L49), it must be public there either. Another example just places a comment that this property [must exist](https://github.com/homebridge/homebridge-examples/blob/master/static-platform-example-typescript/src/switch-accessory.ts#L18-L19).

I tried to develop a `StaticPlatformPlugin` but failed with the exception `Accessories must be created with a non-empty displayName` and  `Initializing platform accessory 'undefined'...`. After adding the property, everything worked well.

Same for the optional property [`uuid_base`](https://github.com/homebridge/homebridge/blob/master/src/server.ts#L429).

**Expected behavior:**
TypeScript hints for the missing required property `name`.

**Logs:**
```
Jan 06 22:08:13 pi-server homebridge[21128]: [2021-1-6 22:08:13] Initializing platform accessory 'undefined'...
Jan 06 22:08:13 pi-server homebridge[21128]: [2021-1-6 22:08:13] [Artnet Universe] Create lightbulb service
Jan 06 22:08:13 pi-server homebridge[21128]: [2021-1-6 22:08:13] Failed to find plugin to handle accessory Spot 1
Jan 06 22:08:13 pi-server homebridge[21128]: [2021-1-6 22:08:13] Removing orphaned accessory Spot 1
Jan 06 22:08:13 pi-server homebridge[21128]: [2021-1-6 22:08:13] Error: Accessories must be created with a non-empty displayName.
Jan 06 22:08:13 pi-server homebridge[21128]:     at new Accessory (/usr/lib/node_modules/homebridge/node_modules/hap-nodejs/src/lib/Accessory.ts:232:29)
Jan 06 22:08:13 pi-server homebridge[21128]:     at Server.createHAPAccessory (/usr/lib/node_modules/homebridge/src/server.ts:463:25)
Jan 06 22:08:13 pi-server homebridge[21128]:     at /usr/lib/node_modules/homebridge/src/server.ts:427:34
[...]

PR #3072, opened 2022-01-23
Code cleanup and improved typings

ericanderson:
This is a code cleanup.


src/api.ts
```diff
@@ -56,6 +56,12 @@ export interface AccessoryPlugin {
    */
   identify?(): void;
 
+  /**
+   * @deprecated
+   * @internal Version needed for internal use to maintain js lecel backcompat
```
ericanderson, line 61: For potential consumers, ts won't emit .d.ts for `@internal`. So it is only usable in this repo.

src/api.ts
```diff
@@ -120,7 +126,7 @@ export interface StaticPlatformPlugin extends PlatformPlugin {
    *
    * @param {(foundAccessories: AccessoryPlugin[]) => void} callback
    */
-  accessories(callback: (foundAccessories: AccessoryPlugin[]) => void): void;
+  accessories(callback: (foundAccessories: Array<AccessoryPlugin & AccessoryConfig>) => void): void;
```
ericanderson, line 129: The code assumes things like `name` and potentially `uuid_base` exist for a static platform, so declare it as such.

src/bridgeService.ts
```diff
@@ -481,13 +480,11 @@ export class BridgeService {
...
 
-      // listen for the identify event if the accessory instance has defined an identify() method
       if (accessoryInstance.identify) {
+        // listen for the identify event if the accessory instance has defined an identify() method
         accessory.on(AccessoryEventTypes.IDENTIFY, (paired: boolean, callback: VoidCallback) => {
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
           // eslint-disable-next-line @typescript-eslint/no-empty-function
-          accessoryInstance.identify!(() => { }); // empty callback for backwards compatibility
+          accessoryInstance.identify?.(() => { }); // empty callback for backwards compatibility
```
ericanderson, line 487: Technically, `accessoryInstance.identify` can be set to `null | undefined` after initialization and therefor this will throw as it is not a function. Safeguarding.

src/bridgeService.ts
```diff
@@ -531,22 +528,13 @@ export class BridgeService {
         log.warn(getLogPrefix(plugin.getPluginIdentifier()), "This plugin is taking long time to load and preventing Homebridge from starting. See https://git.io/JtMGR for more info.");
       }, 20000);
 
-      platformInstance.accessories(once((accessories: AccessoryPlugin[]) => {
+      platformInstance.accessories(once(accessories => {
```
ericanderson, line 531: Blocks on https://github.com/homebridge/HAP-NodeJS/pull/929

src/logger.ts
```diff
@@ -76,26 +62,22 @@ export class Logger {
...
-      log.prefix = logger.prefix;
-
+      const logging = Object.assign(logger.info.bind(logger), {
+        info: logger.info,
+        warn: logger.warn,
+        error: logger.error,
+        debug: logger.debug,
+        log: logger.log,
+        prefix: prefix,
+      });
```
ericanderson, line 79: Object assign returns `typeof arg1 & typeof arg2`, giving proper typing (when using `prefix` instead of `logger.prefix` as `prefix` on `Logger` is optional but we know it is defined for this function). This avoids the dangerous `@ts-ginore` assignment below.

src/platformAccessory.ts
```diff
@@ -28,10 +28,25 @@ export const enum PlatformAccessoryEvent {
...
-export declare interface PlatformAccessory {
+export interface PlatformAccessory {
 
   on(event: "identify", listener: () => void): this;
+  on(event: "identify", listener: (paired: boolean) => void): this;
 
+  /**
+   * @deprecated This version with the noopCallback will be removed in a future version
+   */
+  on(event: "identify", listener: (paired: boolean, noopCallback: () => void) => void): this;
```
ericanderson, line 39: This is actually the reality of what is called.

src/platformAccessory.ts
```diff
@@ -28,10 +28,25 @@ export const enum PlatformAccessoryEvent {
...
+
+  /**
+   * @deprecated Kept for back compat, do not rely on paired or callback.
+   */
+  emit(event: "identify", paired: boolean, callback: () => void): boolean;
```
ericanderson, line 44: This is also the current reality.

src/platformAccessory.ts
```diff
@@ -28,10 +28,25 @@ export const enum PlatformAccessoryEvent {
...
+
+  /**
+   * @alpha This is the future supported version of emit("identify"), invalid until
+   *        the next breaking change.
+   */
   emit(event: "identify"): boolean;
```
ericanderson, line 50: This is the desired reality but if someone were to emit like this today, it would cause bugs.

src/api.ts
Supereg, line 129: I think this typing doesn't really represent reality (and might even break current implementations?).

The weird thing is, that we really only require the `name: string` property (and optionally check for the `uuid_base?: string` property). Using `AccessoryConfig` though will add the `accessory: AccessoryName | AccessoryIdentifier` requirement and the optional `_bridge?: BridgeConfiguration` property which both are never used.

  name: string;
   uuid_base?: string;

src/bridgeService.ts
```diff
@@ -531,22 +528,13 @@ export class BridgeService {
...
         // loop through accessories adding them to the list and registering them
         accessories.forEach((accessoryInstance, index) => {
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
-          const accessoryName = accessoryInstance.name; // assume this property was set
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
-          const uuidBase: string | undefined = accessoryInstance.uuid_base; // optional base uuid
-
-          log.info("Initializing platform accessory '%s'...", accessoryName);
```
Supereg, line 547: Please don't remove this log statement.

src/logger.ts
Supereg: Thanks for resolving this 🚀

src/platformAccessory.ts
```diff
@@ -28,10 +28,25 @@ export const enum PlatformAccessoryEvent {
...
   on(event: "identify", listener: () => void): this;
+  on(event: "identify", listener: (paired: boolean) => void): this;
 
+  /**
+   * @deprecated This version with the noopCallback will be removed in a future version
+   */
+  on(event: "identify", listener: (paired: boolean, noopCallback: () => void) => void): this;
+
+  /**
+   * @deprecated Kept for back compat, do not rely on paired or callback.
```
Supereg, line 42:
```suggestion
   * @deprecated Kept for backward compatibility. Do not rely on `paired` or `callback`.
```

bauer-andreas (commented):
Thanks improving some of the typings of the project 🚀 

I had two minor points. Apart from that, this PR looks good to me 👍

src/api.ts
Supereg: Resolving this would make #2815 obsolete and resolve #2771

src/api.ts
```diff
@@ -120,7 +126,7 @@ export interface StaticPlatformPlugin extends PlatformPlugin {
    *
    * @param {(foundAccessories: AccessoryPlugin[]) => void} callback
    */
-  accessories(callback: (foundAccessories: AccessoryPlugin[]) => void): void;
```
storeonetera, line 123: ericanderson:cleanup-ts-ignore

ghost (changes requested):

bwp91:
Hi all, just popping into get a status update on this!

@ericanderson and @Supereg any status update on this:?

cc @donavanbecker for context

Cheers, Ben!

bwp91:
@Supereg sorry for tag 😃 how can this PR be edited or resolved for merge?

I like the fact that you commented earlier:

> Resolving this would make https://github.com/homebridge/homebridge/pull/2815 obsolete and resolve https://github.com/homebridge/homebridge/issues/2771

donavanbecker:
@ericanderson, please reopen direct towards `beta-1.7.0` branch.

PR #3117, opened 2022-04-22
Strict plugin resolution proposal

oznu:

Currently Homebridge will attempt to load plugins from all `require` paths, default global node module paths, and the custom plugin path if set.

## :bulb: Proposed solution

The optional `--strict-plugin-resolution` flag will ensure Homebridge only loads plugins from the desired location.

If the  `--strict-plugin-resolution`  **and** `-P --plugin-path [path]` flags are set together, Homebridge will only load plugins from the location set from ` --plugin-path`, ignoring any globally installed modules or default system paths.

If the  `--strict-plugin-resolution` is set, **without** the `-P` flag, Homebridge will only load plugins from the current context's global node module install location as resolved by `$(npm -g prefix)/lib/node_modules`. It will ignore all other default system paths.

## :gear: Release Notes

* Added the optional `--strict-plugin-resolution` flag which will ensure Homebridge only loads plugins from the desired location if set.

This is to facilitate isolated, self contained Homebridge installs that do not interact with global installs of node, npm or npm modules.

### Reviewer Nudging

@Supereg

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -23,6 +23,7 @@ export = function cli(): void {
   let hideQRCode = false;
   let keepOrphans = false;
   let customPluginPath: string | undefined = undefined;
+  let strictPluginResolution = false;
   let noLogTimestamps = false;
   let debugModeEnabled = false;
   let forceColourLogging = false;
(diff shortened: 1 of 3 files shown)


bauer-andreas (approved):
Looks great. Thanks 👍

PR #3118, opened 2022-04-26
Replace `git.io` link within the error message

SukkaW:

All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

## :bulb: Proposed solution

Replace all `git.io` links with their actual URLs.


bauer-andreas (approved):
Thanks for the notice. Will be scheduled for release this week.

PR #3235, opened 2022-09-24
add option to override firmware on bridge

donavanbecker:

Currently you can't override the firmware of the bridge for customization

## :bulb: Proposed solution

add config.json option to be able to override firmware.

## :gear: Release Notes

Added option to be able to override bridge firmware version.

### Reviewer Nudging

@Supereg

### DIFF
--- a/config-sample.json
+++ b/config-sample.json
@@ -4,6 +4,7 @@
     "username": "CC:22:3D:E3:CE:30",
     "manufacturer": "homebridge.io",
     "model": "homebridge",
+    "firmware": "1.5.0",
     "port": 51826,
     "pin": "031-45-154"
   },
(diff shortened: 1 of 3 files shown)


Supereg:
What is the rationale making this configurable?

The `FirmwareRevision` characteristic is actually an important way to determine the current version of homebridge running. I see no real use in being able to modify this. And at the same time it might make it much more difficult for us to determine running homebrige version.

donavanbecker:
Just to be able to customize it.

Supereg:
> Just to be able to customize it.

I don’t think it adds much value for the user. But it will possibly increase confusion on our side when we try to debug issues.

PR #3260, opened 2022-11-11
Support resolved mDNS advertiser

elyscape:
I'm gonna be honest, this is a one-line change that should be self-explanatory and I don't feel like filling out the big huge PR template. Context is in https://github.com/homebridge/homebridge/issues/3224#issuecomment-1312064433.

cc: @Supereg


Supereg (approved):
Thanks for addressing this so quickly 🚀

Issue #3332, opened 2023-03-13
Enable ability to store secrets separately from config.json


burkenyo:
### Current Situation

This request revisits a discussion from #1951.

Many accessory and platform plugins require logging in to a third-party API, which often means storing credentials inside config.json. This means that config.json must be treated with extra caution as it now contains secrets. For example, this precludes checking the config file into a source control system. Further, some secrets such as JWTs are very long, which can make the config.json file look unwieldy for hand-editing.

The official documentation even recommends pasting the config.json file into jsonlint.com to help validate the JSON structure and help people less familiar with the format troubleshoot syntax errors. While well-meaning, _please don’t encourage users to paste secrets into an untrusted service_!

### Proposed Change

Instead of placing secrets in the config file directly, they would be stored in a .env file or environment variables. At config load time, instance of variable names prefixed with a “${” and ending in “}” would be replaced with the value. This style of replacement is pretty common, for example in docker compose files. The config file should still be syntactically valid JSON prior to replacement, so replacement variables will need to still be inside valid string values. To prevent breaking existing config.json files that may have what looks like embedded replacement variables, the replacement behavior will be _opt-in_ via a boolean value at the root of the config file. Perhaps this setting can default to true for all new files.

Example:
(before replacement)
```json
{
  "replaceVariables": true,
  "bridge": { ... },
  "platforms": [
    {
      "token": "${PET_SAFE_API_TOKEN}",
      "platform": "PetSafeSmartFeed"
    }
  ]
}
```

(after replacement)
```json
{
  "replaceVariables": true,
  "bridge": { ... },
  "platforms": [
    {
      "token": "eyJ ... (some long JWT)",
      "platform": "PetSafeSmartFeed"
    }
  ]
}
```

Not in scope for this initial work would be to add a UI for setting these secrets and/or previewing the config with replacements made. However, a future endeavor could be to add such a UI to the excellent [oznu/homebridge-config-ui-x](https://github.com/oznu/homebridge-config-ui-x) plugin.

### Additional Context

I’m prepared to take a swing at implementing this, but wanted to give the core maintainers a chance to opine before I drop a pull request out of the blue, as well as hopefully mention any possible gotchas.

burkenyo:
I’ve already sketched out the replacement code:

```ts
type Json = string | number | boolean | null | Json[] | { [key: string]: Json };

function replaceVars(json: Json): void {
  if (!json || typeof json != "object") {
    return;
  }

[...]

PR #3370, opened 2023-05-20
Enable ability to store secrets separately from config.json

burkenyo:

#3332

Many accessory and platform plugins require logging in to a third-party API, which often means storing credentials inside config.json. This means that config.json must be treated with extra caution as it now contains secrets. For example, this precludes checking the config file into a source control system. Further, some secrets such as JWTs are very long, which can make the config.json file look unwieldy for hand-editing.

## :bulb: Proposed solution

A new, optional property named “variableReplacementProvider” is added to config.json. Valid values correspond to the key of a `replacer.ValueProvider` function in the `replacer.variableReplacementProviders` object (`replacer.ValueProvider` is the type of a function which takes the name of a replacement variable and returns the value or undefined if the variable is not defined/known).

When a valid variableReplacementProvider is given, the server calls `replacer.replaceVars` to replace instances of variables names found in config strings with their values. Replacement happens immediately after the config JSON text is parsed. To be replaced, variable names must be delimited by `${...}`. This PR introduces providers which use environment variables or a secrets file.

## :gear: Release Notes

This feature is opt-in. To enable, the user should add “variableReplacementProvider” to their `config.json`. Then, they can replace any sensitive information in the config with delimited variables.

To use the “Environment” variableReplacementProvider, set the appropriate environment variables before starting the server. This can be done in combination with the docker image or using the facilities provided by the OS shell when running directly.

To use the “SecretsFile” variableReplacementProvider, create a file named `config.secrets.json` alongside the config file. It should be simple JSON object: the keys are the replacement variables names.

The use of a `replacer.ValueProvider` function type allows adding other variableReplacementProviders in the future. For example, providers that access a vault service could be added.

### Testing

The `replacer.spec.ts` contains tests for `replacer.replaceVars` and `variableReplacementProviders`. The `server.spec.ts` file is amended to include tests that ensure variables are actually replaced.

### Reviewer Nudging

A good place to start would be in `server.ts` in the `loadConfig()` function.


donavanbecker (dismissed):

NorthernMan54:
@Supereg What's your thoughts ?  The concept of making the secrets external to the main config file makes sense.   The PR has tests included so that's a good thing.

Obliviously this will need developer documentation and complementary adjustments to config-ui-x to support the feature, and we would need to take those on.  But what do you think?

And the failed tests need investigation.

We were thinking to merge the feature in beta, then while it is in beta, work thru the config-ui-x changes, then when both are complete do the release.

donavanbecker (dismissed):

burkenyo:
Hi @NorthernMan54, thanks for looking at this! Is there something you need me to do?

NorthernMan54:
@burkenyo My goal is to have this published, with a couple of plugins leveraging the feature prior to the end of December ( fingers crossed )

After looking at the implementation, our initial thoughts were that creating and maintaining the separate secrets are going to be a steep curve for some users, so having it setup automatically within config-ui-x will be required.  So the changes with config-ui-x will need to be worked out.

Maybe have something that looks for common credential key names, and does the insertion into config.json automatically.  If a plugin has a key called password, insert the substitution with PLUGIN-NAME-KEY, and store the value within secrets.json.  It would also need some clean up process when a plugin is removed from the config, hence having the plugin name as part of it.

bwp91:
Hi @burkenyo @NorthernMan54 

Just looking at this myself. Totally on board with the concept!

I think there are some situations in which we need to at least consider

#### 1. Plugins which write directly to the config file

If such plugins exist (?). A user might look at their config json, and replace a value with the secret syntax, only for it to then be overridden by the plugin again. Then we would have an 'old' secret value in the proposed file as well as the (potentially) different value in the hb config.

#### 2. Plugin UIs which write to the config file

With the custom UI functionality in config-ui-x, same as (1) basically.

---

Which then leads to further questions:

1. would a plugin which is running be able to detect if the user has this option enabled, and be able to decypher between writing to the secrets file rather than the config itself (related to point 1 above)

2. same for a plugin UI on the basis of 1 (related to point 2 above)

---

considerations for the HB UI:

- would this file be included in backups?
- if the user is saving a raw json config, should it validate that any newly entered secret exists? (and existing?)

---

i'm sure i'll have more thoughts on next reflection!

bwp91:
Upon an as promised reflection!

I realise my thoughts above are more about the UI package and also general feature visibility to devs.

And i’m sure that if a value in the main json file were to be overridden by a plugin then the keen user would correctly kick up a fuss!

src/util/replacer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return;
+  }
+
+  let didReplace = false;
+
+  // look for variable names surrounded by “${” and ”}”
+  // match variable names starting with a letter or underscore and
+  // consisting entirely of letters, numbers, and underscores
+  const newValue = value.replace(/\$\{([A-Za-z_]\w*)\}/g, (fullMatch, envVarName) => {
```
bwp91, line 95: I just wonder if this could/should be restricted somehow to ensure the `${` and final `}` are on the same line, just to make sure that if a user incorrectly manages to or a plugin miswrites to the config.json then there is less chance of any mishaps.

NorthernMan54:
> Hi @burkenyo @NorthernMan54
> 
> Just looking at this myself. Totally on board with the concept!
> 
> I think there are some situations in which we need to at least consider
> 
> #### 1. Plugins which write directly to the config file
> If such plugins exist (?). A user might look at their config json, and replace a value with the secret syntax, only for it to then be overridden by the plugin again. Then we would have an 'old' secret value in the proposed file as well as the (potentially) different value in the hb config.

In homebridge-config-ui-x, we would need to identify and block these plugins for using or supporting the feature automatically.  ( Start there and adjust this as needed )

> 
> #### 2. Plugin UIs which write to the config file
> With the custom UI functionality in config-ui-x, same as (1) basically.
> 
> Which then leads to further questions:
> 
> 1. would a plugin which is running be able to detect if the user has this option enabled, and be able to decypher between writing to the secrets file rather than the config itself (related to point 1 above)
> 2. same for a plugin UI on the basis of 1 (related to point 2 above)
> 
> considerations for the HB UI:
> 
> * would this file be included in backups?

This is a must, or else restores would break


> * if the user is saving a raw json config, should it validate that any newly entered secret exists? (and existing?)


A lot of complexity with this, am thinking no

> 
> i'm sure i'll have more thoughts on next reflection!

src/util/replacer.ts
burkenyo: Hi @bwp91, this regex will only match tokens starting with `${`, containing only letters, number, and underscores, and ending with `}`, so new-lines or any other white-space should be matched.

bwp91:
Not sure if this has been called out already

> Maybe have something that looks for common credential key names, and does the insertion into config.json automatically. If a plugin has a key called password, insert the substitution with PLUGIN-NAME-KEY, and store the value within secrets.json. It would also need some clean up process when a plugin is removed from the config, hence having the plugin name as part of it.

What happens if a user is running two instances of a platform plugin?

hjdhjd:
Opinionated editorial:

I don't mind a feature like this existing but it should most definitely not be marketed as a secure solution. At best it is slightly more secure, at worst its security through obscurity and security theater.

Anyone - including any plugin - that has access to the machine Homebridge runs on will be able to cause issues. Hiding login credentials doesn't buy you anything besides "feeling better". I don't mind feeling better, but I mind overpromising what this can and cannot do in providing real security.

Unless you can secure every part of the chain - all the dependencies and all the plugins and Homebridge itself - this is just a veneer of security that will have little real impact to most installations.

And it's going to likely break with a bunch of fragile edge cases that aren't yet fully conceived of or thought through and create more support requests.

TL;DR: nice idea, but this is harder to implement robustly than meets the eye and should be done with extreme caution, proper documentation, proper representation of what this is and is not, and not oversell its capabilities.

Definitely glad it's opt in. 😀

JurgenLB:
@hjdhjd 
 I believe this is not to secure the config-file but it is for easy charing/checking config-file so the credentials are not wide spread.
But I agree with your security opinion.

hjdhjd:
> @hjdhjd 
> 
>  I believe this is not to secure the config-file but it is for easy charing/checking config-file so the credentials are not wide spread.
> 
> But I agree with your security opinion.

I truly do not see what advantage there is that rises to the level of creating a first-party feature in Homebridge itself to do this. This is feature bloat that will be unlikely to be used by the vast majority of people.

If you are adding core features to a platform like Homebridge they should have a decent likelihood of being adopted by a meaningful subset of users. Otherwise you are bloating the codebase for something that likely won't be well tested as Homebridge evolves, likely won't be kept up with, all for a niche feature that hasn't met the threshold of something that deserves to exist in the core codebase.

If a plugin author wanted to provide additional vectors to provide credentials or secrets they can. A centralized feature like this isn't needed.

Put another way: where is the critical mass of humanity clamoring for this to justify the maintenance and support overhead?

cbrunnkvist:
As an outside observer and user, I'd agree this is nice but quite a major change with no clear majority target user, or target use case... 

I'd be glad if the plentiful spellchecker fixes in this PR could go in separately though 😉

PR #3436, opened 2023-09-18
chore: update dependencies

bwp91:

- some dependencies are out of date
- gitignore file does not include `coverage` directory
- package.json engine node versions are at node v10 (we now only support 18, but have added 20 too)

I would like to start looking into updating some dependencies into their major new versions but wanted to get this done as the first stepping stone.

## :bulb: Proposed solution

- updates `dependencies` to latest minor/patch versions
- updates `devDependensies` to some latest major/minor/patch versions
- full refresh of `package-lock.json`
- include `coverage` in gitignore file
- update package.json file engines - node - to `^16.20.2 || ^18.17.1`

## :gear: Release Notes

```
chore: update dependencies
```

### Testing

no tests changes, but all are passing

### Reviewer Nudging

🤷‍♂️

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=10.17.0"
+    "node": "^16.20.2 || ^18.17.1"
   },
   "files": [
     "README.md",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "homebridge": "bin/homebridge"
   },
   "engines": {
-    "node": ">=10.17.0"
+    "node": "^16.20.2 || ^18.17.1"
```
NorthernMan54, line 36: The synology package is using a slightly older Node 18 package nodejs.org 18.15.0-1008

So minimum version needs to take this into account

bwp91:
Hi @NorthernMan54 

The node engine part was an _extra_ I did, not sure if was really necessary or not. Have updated the downgrade of 18.

But I also want to make it clear that these are the versions of node that homebridge supports, not necessarily the versions of node that homebridge will work with

donavanbecker (dismissed):
Need this beta.

bwp91:
@donavanbecker which is latest beta branch?

package.json
bwp91: Hi @NorthernMan54 I set minimum engine node version to 18.15.0.

I would preferably like to push this into a beta branch of homebridge rather than master, but not totally sure how it works on this repo

donavanbecker:
would think beta-1.6.1

Supereg:
> @donavanbecker which is latest beta branch?

Assuming, we are fine with a minor bump, I just created a 1.7.0 beta right now.

donavanbecker:
Thanks @Supereg, @bwp91 looks like we can get rid of all of the other beta branches also

donavanbecker (approved):

donavanbecker:
@bwp91, I think we are good to go on this. Unless there is something else you want to add. new beta was published for this: `v1.7.0-beta.0`

PR #3466, opened 2023-10-29
do not delete branch `beta`

bwp91:
this branch/PR does not need to be merged.

it is just a reminder to keep this branch name called `beta` so the commit link in the homebridge UI works.

Beta work can continue on branch `beta-1.7.0`


bwp91:
I think this is safe to close now

PR #3467, opened 2023-10-29
beta -> latest

bwp91:
⚠️ we will need to make clear in the release notes that any version of node that is not `18` nor `20` is not supported

A lot of dependencies are out of date.

## :bulb: Proposed solution

Update them :)

## :gear: Release Notes

- Minimum supported node version is now 18, officially v18 and v20 are supported
- Updated dependencies

Node 18 really is the least supported version now, as changes in the package json engines

### Testing

No tests were changed, but all are still passing


bauer-andreas (dismissed):
Thanks for all the work 🚀

bwp91:
@Supereg this check

> build_and_test / build (16) Expected — Waiting for status to be reported

I can't see where it comes from ?

bauer-andreas:
> @Supereg this check
> 
> > build_and_test / build (16) Expected — Waiting for status to be reported
> 
> I can't see where it comes from ?

It's still configured as a required check in the branch protection rules. So it shows up as required (but isn't scheduled as it doesn't exist anymore). I'll remove it real quick.

donavanbecker (approved):

PR #3522, opened 2023-12-31
Please add to WIki - info about SSH and FTP

mkz212:


JurgenLB:
Maybe you can also mention
`ssh-keygen
ssh-copy-id  <user>@<IP>     example; pi@192.168.1.100`
so the connection is secure and the login is automatically

maxileith:
I don't think this should be added to this repo. It has pretty much nothing to do with the npm package. 

Adding this to the pi image repo would be a better fit.

PR #3581, opened 2024-03-23
add logging success colour

bwp91:
I have been thinking about doing this for a while - adding a 'green' logging colour available to plugins to highlight _successful_ actions that might otherwise be missed if white and in the middle of another plugin's spam.

What do you guys think?

### DIFF
--- a/src/logger.spec.ts
+++ b/src/logger.spec.ts
@@ -1,0 +1,166 @@
+import {forceColor, Logger, setDebugEnabled, setTimestampEnabled, withPrefix} from "./logger";
+import chalk from "chalk";
+
+describe("Logger", () => {
+  let consoleLogSpy: jest.SpyInstance;
+  let consoleErrorSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleLogSpy = jest.spyOn(console, "log").mockImplementation();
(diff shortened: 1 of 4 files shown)


src/logger.spec.ts
bauer-andreas, line 1: Thank you for adding new unit tests to the project 🥳

src/logger.ts
```diff
@@ -152,6 +161,9 @@ export class Logger {
 
     let loggingFunction = console.log;
     switch (level) {
+      case LogLevel.SUCCESS:
+        message = chalk.green(message);
+        break;
```
bauer-andreas, line 166: Great addition thanks 👍

bauer-andreas (approved):
Sounds like a cool idea, happy to have it merged.

Just one thing, as you are trying to merge this into the 1.7.1 beta branch. I would consider this to be part of a minor release, so should we aim for a 1.8.0 beta here?

bwp91:
I was thinking to merge this into the beta-1.7.1 branch but when we end up releasing the new version it should be bumped to 1.8.0.

bauer-andreas:
> I was thinking to merge this into the beta-1.7.1 branch but when we end up releasing the new version it should be bumped to 1.8.0.

Sounds good 👍

PR #3609, opened 2024-05-04
Allow plugins to properly type platform config during registration

duddu:

Currently if a plugin developer wants to strongly type their own plugin configuration interface (e.g. extending `PlatformConfig` and thus having config fields type checking throughout all plugin code), they would be somehow blocked once they invoke `api.registerPlatform` to register the platform, since `PlatformPluginConstructor.new` has the argument `config` typed as `PlatformConfig`. A brief snippet to exemplify the scenario:

```typescript
import { API, DynamicPlatformPlugin, Logger, PlatformConfig } from 'homebridge';

interface MyConfig extends PlatformConfig {
  myProperty: string;
}

class MyPlatform implements DynamicPlatformPlugin {
  constructor(
    public readonly log: Logger,
    public readonly config: MyConfig,
    public readonly api: API,
  ) {}
  // ...
}

export = (api: API) => {
  api.registerPlatform('MyPlugin', 'MyPlatform', MyPlatform); // <- Here tsc will bomb
};

// ERROR:
// Argument of type 'typeof MyPlatform' is not assignable to parameter of type 'PlatformPluginConstructor'.
//   Types of parameters 'config' and 'config' are incompatible.
//     Property 'myProperty' is missing in type 'PlatformConfig' but required in type 'MyConfig'. ts(2345)
```

## :bulb: Proposed solution

An easy solution is to update the `PlatformPluginConstructor` interface by adding a generic that allows to auto-detect the type of `config` argument of the `new` method:

```typescript
// src/api.ts

export interface PlatformPluginConstructor<Config extends PlatformConfig = PlatformConfig> {
  new(logger: Logging, config: Config, api: API): DynamicPlatformPlugin | StaticPlatformPlugin | IndependentPlatformPlugin;
}

// And updating `API` interface `registerPlatform` method typing to reflect this.
```
With this change the same snippet above will work fine - example from vscode:
Screenshot 2024-05-04 at 08 02 35

I think this is a useful change because it encourages plugin developers to strongly type their config (extending `PlatformConfig`), and allows to use everywhere a single config interface (i.e. when cascading down config from platform instance).


bwp91:
Hi @duddu 

Would you be able to resolve the conflicts on the docs/ files, perhaps remove the changes to these files from your branch, are they necessary to be regenerated as part of your changes?

duddu:
thanks @bwp91, rebased from homebridge/latest and re-generated docs.
i'm afraid diffs in typedoc dir are needed as my change updates an interface (see https://github.com/homebridge/homebridge/pull/3609/files#diff-769911c416ccf8514d8fd941ae0abe8fb5c606ade0c218e22151a5f5f9f3d700R84)

bwp91:
@donavanbecker can we merge this into a beta?

donavanbecker (approved):

PR #3664, opened 2024-07-21
Fix code scanning alert - Use of a broken or weak cryptographic algorithm

donavanbecker:
Tracking issue for:
- [x] https://github.com/homebridge/homebridge/security/code-scanning/2

### DIFF
--- a/src/util/mac.ts
+++ b/src/util/mac.ts
@@ -9,9 +9,9 @@ export function validMacAddress(address: string): boolean {
 }
 
 export function generate(data: string | Buffer | NodeJS.TypedArray | DataView): MacAddress {
-  const sha1sum = crypto.createHash("sha1");
-  sha1sum.update(data);
-  const s = sha1sum.digest("hex");
+  const sha256sum = crypto.createHash("sha256");
+  sha256sum.update(data);
+  const s = sha256sum.digest("hex");
 
   let i = 0;
   return "xx:xx:xx:xx:xx:xx".replace(/x/g, () => s[i++]).toUpperCase();


bwp91:
@donavanbecker this is just used for generating new pins right? it won't have any effect on existing stuff? i mean this isn't a breaking change right?

src/util/mac.ts
```diff
@@ -9,10 +9,10 @@ export function validMacAddress(address: string): boolean {
 }
 
 export function generate(data: string | Buffer | NodeJS.TypedArray | DataView): MacAddress {
-  const sha1sum = crypto.createHash("sha1");
-  sha1sum.update(data);
-  const s = sha1sum.digest("hex");
+  const sha256sum = crypto.createHash("sha256");
```
bauer-andreas, line 12: If I remember correctly, the MAC address (called username in HAP-NodeJS) is used as an identifier for all storage files. See  https://github.com/homebridge/HAP-NodeJS/blob/8804db59230b4fc8e2d4a2f9971ea05a3def28d2/src/lib/Accessory.ts#L1126

This will break all persistent storage of HAP-NodeJS users. The usage of sha1 ist not security relevant. So I would just leave it like that! Or what do you think? 
Otherwise we definitely need some migration in place.

bauer-andreas (changes requested):

src/util/mac.ts
bauer-andreas: This is the critical part where we use it in the homebridge code base https://github.com/homebridge/homebridge/blob/12b5338bede5dad0c6fee038254db627eb4f958d/src/bridgeService.ts#L412

https://github.com/homebridge/homebridge/blob/12b5338bede5dad0c6fee038254db627eb4f958d/src/bridgeService.ts#L438

donavanbecker: I am good with leaving and closing PR.

PR #3674, opened 2024-08-14
[DRAFT] Implementation of runtime checks for deprecated elements with v2

muenzpraeger:

With the upcoming breaking changes in v2 it's crucial to inform users that some of their used plugins may not work after performing the upgrade. The upgrade work relies on the plugin developers to update their plugins, which may - or may not - happen.

## :bulb: Proposed solution

To support the users of homebridge runtime checks should be implemented to direct them at what/where things can break in case they would upgrade

## :gear: Release Notes

To be added if we're aligned on this as a viable solution.

This implementation has the caveat that it purely (has to) rely on runtime invocation. 

Example:
 - An accessory may have some unit format assigned via `Characteristic.Units` during creation - this will cause the warning to log. If it's loaded from cache the warning won't be logged as it's already defined.

### Testing

To be added if we're aligned on this as a viable solution.

### Reviewer Nudging

To be added if we're aligned on this as a viable solution.


hjdhjd:
Love the idea of this...would you be willing to take on a bit further evolution of it potentially. Essentially, many of the "fixes" (though certainly not all) are relatively trivial to make. Perhaps:

* Create diffs to patch the plugin.
* Point users to a "fixed" version of the plugin. In the case of abandoned plugins, a reference to a directory (likely in the homebridge org) referencing the corrected plugin.
* Potentially patching-in-place? See patch-package and it's kin...

TL;DR: would be nice to self-heal/upgrade-in-place nonconforming plugins where possible. There will be instances where this can't be done, but it's a perennial 80/20 problem.

One of the assets Homebridge has is it's vast library of plugins...we'd love to find a way to bring them into the future without being overly encumbered by legacy code where we can be.

Welcome perspectives...

ghost (approved):

bwp91:
Thank you for the work on this one, and apologies it sat for as long as it did.

Closing it now rather than carrying it forward: this was designed to warn users about deprecated elements ahead of the v2 upgrade, and with v2 released quite a while back (we are on the v2.2.x line now) that pre-upgrade window has passed — the runtime checks no longer have a job to do.

The thinking behind it was sound, and the cache-loaded-accessory caveat you called out in the description was a genuinely useful observation for the upgrade work at the time. Thanks again for contributing it.

PR #3687, opened 2024-09-01
support plugins with dots in their @scope

gtalusan:

I'm writing a plugin and my npm @scope has a '.' in it.  Homebridge won't load it.

## :bulb: Proposed solution

Changing PLUGIN_IDENTIFIER_PATTERN to include a period in the regex. 

## :gear: Release Notes

Support plugins with dots in their @scope

N/A

### Testing

Add corresponding tests to pluginManager.spec.ts.

### DIFF
--- a/src/pluginManager.ts
+++ b/src/pluginManager.ts
@@ -64,7 +64,7 @@ export interface PluginManagerOptions {
 export class PluginManager {
 
   // name must be prefixed with 'homebridge-' or '@scope/homebridge-'
-  private static readonly PLUGIN_IDENTIFIER_PATTERN = /^((@[\w-]*)\/)?(homebridge-[\w-]*)$/;
+  private static readonly PLUGIN_IDENTIFIER_PATTERN = /^((@[\w-]+(\.[\w-]+)*)\/)?(homebridge-[\w-]+)$/;
 
   private readonly api: HomebridgeAPI;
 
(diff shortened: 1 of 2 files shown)


src/pluginManager.ts
```diff
@@ -64,7 +64,7 @@ export interface PluginManagerOptions {
 export class PluginManager {
 
   // name must be prefixed with 'homebridge-' or '@scope/homebridge-'
-  private static readonly PLUGIN_IDENTIFIER_PATTERN = /^((@[\w-]*)\/)?(homebridge-[\w-]*)$/;
+  private static readonly PLUGIN_IDENTIFIER_PATTERN = /^((@[\w-\.]*)\/)?(homebridge-[\w-]*)$/;
```
bwp91: since the lint needs fixing anyway, i wondered if we should disallow a `.` as the first or last character of a scope name.

i think something like this should do it

```
/^(@[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)*\/)?homebridge-[\w-]+$/
```

bwp91: didn't test mine against the lint! updated:

```
/^(@[\w-]+(\.[\w-]+)*\/)?homebridge-[\w-]+$/
```

gtalusan: Yep, that works and makes more sense.  Thanks!

gtalusan: Oops, looks like it's missing some brackets.  Will fix up.

bwp91: does the one in the UI repo need fixing?

gtalusan: No, it doesn't -- I was in the middle of double checking when you merged it.  It only uses .test() so it's safe even if it's not consistent with homebridge.

bwp91 (approved):

bwp91:
@gtalusan i have pushed your changes to both the alpha and beta versions of homebridge and the UI. maybe you have these running and could update and test that you can install your plugin correctly!

PR #3899, opened 2026-03-02
Update bridgeService.ts: make publishInfo.addIdentifyingMaterial opt-out

LukasMojzis:

Currently homebridge forcefully adds unique suffix to whatever homebridge name is chosen and resulting name may make things confusing for end-users.

## :bulb: Proposed solution

This makes it optional. Needs documentation though.

## :gear: Release Notes

Make bridge name suffix optional.

If admin decides to name a bridge a certain way, respect it.

### Testing

None

### Reviewer Nudging

Self evident

### DIFF
--- a/src/bridgeService.ts
+++ b/src/bridgeService.ts
@@ -204,7 +204,7 @@ export class BridgeService {
       category: Categories.BRIDGE,
       bind: bridgeConfig.bind,
       mdns: this.config.mdns, // this is deprecated now
-      addIdentifyingMaterial: true,
+      addIdentifyingMaterial: bridgeConfig.addNameSuffix !== false,
       advertiser: bridgeConfig.advertiser,
     };
 


bwp91:
Hi @LukasMojzis, and first of all my apologies — this PR has sat here since March without a single response, which is not the experience anyone should have after taking the time to contribute. That one is on us, and I am sorry.

Thank you for raising it. You spotted a real problem and you were first to it: the hard-coded `addIdentifyingMaterial: true` did need to become configurable, and your reasoning about the resulting Bonjour name was sound.

This has now been implemented in #3965. That version grew a little beyond the original scope — it hangs the option off the existing `bridge.hap` config block, applies to external accessories as well as the main bridge, forwards to child bridges, validates the value, and warns when it is disabled, since that suffix is what keeps mDNS service names unique. It lands as `hap.disableIdentifyingMaterial`.

None of that is a criticism of your patch, which was correct in what it did. The surrounding requirements simply became clearer over the months this sat unattended, and the newer PR was written against them.

Closing as superseded — but genuinely, thank you for the contribution and for your patience. If you fancy sending anything else our way, it will get a far quicker response than this did.

PR #3912, opened 2026-04-19
fix: add powerSource to clusterNames and ClusterStateMap

gtalusan:
PowerSource cluster can be used by accessories to report battery level. Add it to `clusterNames` (eliminates spurious warning on `updateAccessoryState`), `ClusterStateMap` (type safety), and `clusterTypes.ts` (`PowerSourceState` interface).

PR #3914, opened 2026-04-19
fix: enable Rechargeable feature for PowerSource cluster when batChargeState is provided

gtalusan:
## Summary

When a plugin provides `batChargeState` in the `powerSource` cluster config, the `AccessoryManager` now enables the `Rechargeable` feature flag via `PowerSourceServer.with('Battery', 'Rechargeable')`.

## Problem

The `batChargeState` attribute (IsCharging / IsNotCharging) is only available under the Matter `Rechargeable` feature. Previously, `AccessoryManager` only used `with('Battery')`, so any `batChargeState` values provided by plugins were silently ignored and HomeKit always showed "Charging: No".

## Change

```
if (hasBattery && hasRechargeable) {
  powerSourceBehavior = PowerSourceServer.with('Battery', 'Rechargeable')
} else if (hasBattery) {
  powerSourceBehavior = PowerSourceServer.with('Battery')
}
```

Where `hasRechargeable` is true when `batChargeState !== undefined` in the initial cluster config.

## Related

Follows on from #3912 which added `powerSource` to `clusterNames` and `ClusterStateMap`.

### DIFF
--- a/src/matter/server/AccessoryManager.ts
+++ b/src/matter/server/AccessoryManager.ts
@@ -400,8 +400,12 @@ export class AccessoryManager {
       if (accessory.clusters?.powerSource) {
         const hasBattery = accessory.clusters.powerSource.batPercentRemaining !== undefined
           || accessory.clusters.powerSource.batChargeLevel !== undefined
+        const hasRechargeable = accessory.clusters.powerSource.batChargeState !== undefined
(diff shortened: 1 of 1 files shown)


gtalusan:
Screenshot 2026-04-19 at 12 03 33

bwp91 (approved):

PR #3915, opened 2026-04-19
fix: remove API event listeners in Matter manager teardown

gtalusan:
`MatterBridgeManager` and `ChildBridgeMatterManager` each registered 6 inline arrow-function listeners on the API `EventEmitter` in `setupEventListeners()` but never removed them in `teardown()`. Because the listeners were anonymous closures, they could not be deregistered and held a reference to the entire manager object, preventing garbage collection on any restart/reload path that re-creates the managers without a full process exit.

**Fix:** Each listener is now a `private readonly` arrow-property (bound once at construction). `setupEventListeners()` registers these stored references; `teardown()` removes them with `api.removeListener()`.

Affects: `src/matter/MatterBridgeManager.ts`, `src/matter/ChildBridgeMatterManager.ts`

### DIFF
--- a/src/matter/ChildBridgeMatterManager.ts
+++ b/src/matter/ChildBridgeMatterManager.ts
@@ -59,6 +59,44 @@ export class ChildBridgeMatterManager extends BaseMatterManager {
     this.matterConfig = bridgeConfig.matter
   }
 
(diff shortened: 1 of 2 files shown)


bwp91 (approved):
Thanks!

PR #3917, opened 2026-04-20
fix: remove duplicate API event listeners from Server

gtalusan:
The Matter implementation added three `InternalAPIEvent` listeners to `Server` (`REGISTER_PLATFORM_ACCESSORIES`, `UNREGISTER_PLATFORM_ACCESSORIES`, `PUBLISH_EXTERNAL_ACCESSORIES`), but `BridgeService` already registers the same three events. Every plugin accessory event fired both handlers, causing `saveCachedPlatformAccessoriesOnDisk()` to be called twice and logging spurious UUID collision warnings for every registered accessory.

Removes the duplicate listeners and their now-dead handler methods from `Server`.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -21,14 +21,13 @@ import { AccessoryEventTypes, MDNSAdvertiser } from '@homebridge/hap-nodejs'
 import chalk from 'chalk'
 import qrcode from 'qrcode-terminal'
 
-import { HomebridgeAPI, InternalAPIEvent, PluginType } from './api.js'
+import { HomebridgeAPI, PluginType } from './api.js'
 import { BridgeService } from './bridgeService.js'
 import { ChildBridgeService } from './childBridgeService.js'
 import { ExternalPortService } from './externalPortService.js'
 import { IpcIncomingEvent, IpcOutgoingEvent, IpcService, ServerStatusUpdate } from './ipcService.js'
 import { Logger } from './logger.js'
 import { MatterConfigCollector } from './matter/config.js'
-import { PlatformAccessory } from './platformAccessory.js'
 import { PluginManager } from './pluginManager.js'
 import { User } from './user.js'
 import { validMacAddress } from './util/mac.js'
@@ -140,13 +139,6 @@ export class Server {
 
     // Note: MatterBridgeManager creation is deferred to start() to avoid loading
     // heavy Matter.js libraries during construction when Matter may not be configured
-
-    // Handle platform accessory registration
-    this.api.on(InternalAPIEvent.REGISTER_PLATFORM_ACCESSORIES, this.handleRegisterPlatformAccessories.bind(this))
-    this.api.on(InternalAPIEvent.UNREGISTER_PLATFORM_ACCESSORIES, this.handleUnregisterPlatformAccessories.bind(this))
-
-    // Handle external accessories (cameras, etc.)
-    this.api.on(InternalAPIEvent.PUBLISH_EXTERNAL_ACCESSORIES, this.handlePublishExternalAccessories.bind(this))
 
     // Watch bridge events to check when server is online
     this.bridgeService.bridge.on(AccessoryEventTypes.ADVERTISED, () => {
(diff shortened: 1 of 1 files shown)


gtalusan:
Since installing beta-2.0.0 to play around with Matter, I've noticed increased CPU usage and memory growth of my homebridge docker container even while sleeping:

Screenshot 2026-04-20 at 10 49 00

The initial peak at 2026-04-18 21:00 is from my dev work on my plugin.  The 3 saw-tooth waves in the second half of the graph, are with leaving this particular HB alone and just restarting it twice.

Not sure if this is the fix, but I'll keep stabbing at it...

bwp91 (approved):

gtalusan:
@bwp91 thanks for the quick merges!  Unfortunately, I'm still seeing increased memory usage with beta-2.0.0.  It's about 15k objects per hour.  I took some heap snapshots with Chrome to track what's going on.  Here's the analysis of the delta from copilot:

Heap snapshot delta analysis (two snapshots ~1hr apart):

- Net new objects: 14,849 unique objects in snapshot 2 vs snapshot 1
- Dominant growth: 8,400 HAP-TCP strings + ~8,441 DNS-SD strings (5.2MB combined self_size)
- All trace back to one root: MDNSServer.sentPackets — a Map<InterfaceName, string[]> in @homebridge/ciao that stores a base64-encoded copy of every sent mDNS packet per network interface
- Entries are only removed when the exact same packet bytes arrive back via UDP loopback — if that receipt is missed (dropped packet, busy network), the entry stays forever
- After ~1hr: thousands of these strings are piling up, never freed

This felt like deja vu... I have a PR to ciao here: https://github.com/homebridge/ciao/pull/66 - any chance that could get some love?

Thanks!!

bwp91:
Hi @gtalusan i'm grateful for the PRs, so thanks again for your improvements!

Can you give homebridge 2.0.0-beta.99? this includes a beta version of hap-nodejs which includes a beta version of ciao (which itself _hopefully_ includes a fix for the memory leak).

Let me know how you get on 👍

gtalusan:
Hi @bwp91, that did the trick!  Here's a trace for my run last night:

Screenshot 2026-04-21 at 10 29 39

Between 16:00ish-21:00ish was with my ciao patch.  Afterward is with 2.0.0-beta.99.  There is still a bit of see-sawing.  I'm not sure if it's due to my inspector being attached with beta.99 or what, but in either case, this is still way more controlled.

copilot synopsis:

What's "growing" and why it doesn't matter
- V8 JIT warm-up (4.2MB new, 3.6MB freed, net 640KB) — normal, will plateau once code paths are warm                                                                                                                                                                     
- matter.js DNS-SD Scheduler strings — 388 strings/hr vs 16,000/hr pre-fix; structural objects all Δ=0; rate fell 13× from 30-min burst, confirming it was initial device discovery                                                          │
- +10 StandardTimers over 8 hours — negligible                                                                                                                                                                                              - MdnsAdvertiser Map — only 14 total entries (3 devices × interfaces × ports), fully bounded  

What's confirmed stable
- AbortSignal, DnssdNames, IpService, Peer, Scheduler, Cache, ServerSubscription, MDNSServer, CiaoService — all Δ=0
- HAP-TCP strings: 4 new over 8 hours (was ~8,400/hr pre-fix) ✅                                                                                                                                                                             
- sentPackets: 1 node (empty Map) ✅ 

Thanks for the quick turnarounds.  Much appreciated!!

gtalusan:
One more thing of note: CPU usage is still higher than homebridge@latest and the temperature difference is measurable.

Screenshot 2026-04-21 at 10 43 13

7 day trace, 4/18 21:00-ish is when 2.0.0-beta is installed, and you can see when the ciao fixes went in last night around 4/20 21:00-ish.

```
62.3°C today
slope: +1.65°C/day
last 7 days:
  2026-04-15: 52.4°C
  2026-04-16: 52.3°C
  2026-04-17: 53.0°C
  2026-04-18: 51.7°C
  2026-04-19: 63.6°C
  2026-04-20: 65.4°C
  2026-04-21: 62.3°C
```

PR #3920, opened 2026-04-22
fix(matter): bind MdnsService to bridge.bind interface

gtalusan:
## Problem

Matter's `MdnsService` opens mDNS multicast sockets on every host interface. In a Docker `--network=host` container with 8+ interfaces this created **16 mDNS sockets**, causing **1,694 voluntary context switches/sec** and ~25% CPU usage driven by `recvmsg` syscalls processing all LAN multicast traffic.

## Root Cause — Two Bugs

**Bug 1 (timing):** `mdns.networkInterface` must be set in `Environment.default.vars` *before* `ServerNode.create()`, because `MdnsService` reads it at construction time. It was never being set at all, so Matter always bound to every interface.

**Bug 2 (redundancy):** The existing `network.interface` fix (abe5b71) already handled Matter UDP transport correctly, but `mdns.networkInterface` was a completely separate env var that nothing was setting.

## Fix

In `ServerLifecycle.start()`:
1. Before `createServerNodeWithRecovery()`: set `mdns.networkInterface` to the first entry in `config.networkInterfaces` (derived from `bridge.bind`), consistent with how Homebridge's own HAP/ciao mDNS stack uses `bridge.bind`.
2. If `bridge.bind` is not set: log a warning and let Matter fall back to all interfaces — same behaviour as ciao, and avoids custom detection heuristics.
3. Clear any stale `mdns.networkInterface` from the shared `Environment.default` singleton before each `ServerNode` creation (mirrors the existing `network.interface` clear added in abe5b71).

Insight into normal socket counts came from comparing against a `node-red-matter-bridge` deployment that exhibited no CPU anomalies.

## Result

| Metric | Before | After |
|--------|--------|-------|
| mDNS sockets (eufy child bridge) | 16 | 5 |
| Voluntary ctx switches/sec | 1,694 | ~23 |
| Container CPU (median) | ~25% | ~9.5% |

### DIFF
--- a/src/matter/server/ServerLifecycle.ts
+++ b/src/matter/server/ServerLifecycle.ts
@@ -261,38 +261,53 @@ export class ServerLifecycle {
         }
       }
 
-      // Clear any previously set network.interface from the global environment BEFORE
-      // creating the ServerNode. Environment.default is a singleton shared across all
-      // server instances in the process. If a previous server already set network.interface,
-      // the ValueCaster in Behaviors.defaultsFor() would reject the unknown 'interface'
-      // property when initializing NetworkBehavior on this new node.
-      // VariableService.get() returns a direct reference to the internal vars object,
-      // so deleting the key here mutates the stored value without needing a private API.
+      // Determine the mDNS network interface to use.  MdnsService reads
+      // 'mdns.networkInterface' at construction time (inside createServerNodeWithRecovery),
+      // so this value MUST be set before calling createServerNodeWithRecovery().
+      //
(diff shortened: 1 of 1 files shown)


gtalusan:
This led me on quite a bit of a wild goose chase - I was looking at JS CPU profiles, but looks like most of the CPU burning was waiting on kernelspace.

The CPU usage is now no worse than node-red-matter-bridge which also uses roughly 7-9% CPU.  I would hope that it'd be less since the heat generated by matter  is tripping my networking monitoring alarms.

PR #3965, opened 2026-07-20
feat: make identifying material in bonjour configurable

naterator:
_Note: This PR addresses the same goal as #3899. This implementation uses the current `bridge.hap` configuration block, applies the setting to both the main bridge and external accessories, validates the configured value, covers child bridge forwarding, and includes regression tests._

Homebridge currently hard-codes HAP-NodeJS `addIdentifyingMaterial: true`. This appends additional identifying characters to (potentially) the bridge display name and Bonjour service instance name. The resulting service name might be confusing to end users due to its inconsistency with the bridge display name or what the end user might expect to see.

## :bulb: Proposed solution

HAP-NodeJS already supports disabling this behavior (ie, setting it to `false`), but Homebridge does not expose the option in `config.json`. This change makes the behavior configurable without changing the default for existing users.

## :gear: Release Notes

Update `config.json` to disable appending the extra four identifying characters like this:

```json
"bridge": {
  "name": "Homebridge",
  "hap": {
    "addIdentifyingMaterial": false
  }
}
```

## :test_tube: Testing

Six test cases were added. Here is a brief description of each:

### `src/bridgeService.spec.ts`

- **Forwards `hap.addIdentifyingMaterial` to external accessories**  
  Verifies that external HAP accessories receive the configured value in their publish options.

- **Disables identifying material when configured in the HAP block**  
  Verifies that the main bridge passes `addIdentifyingMaterial: false` to HAP-NodeJS.

- **Defaults to `true` when the HAP block or option is omitted**  
  Confirms backward-compatible behavior, including legacy boolean HAP configurations.

- **Honors an explicit `false` value**  
  Verifies that the new configuration helper recognizes the opt-out.

- **Rejects a non-boolean `hap.addIdentifyingMaterial` value**  
  Ensures invalid values, such as the string `"false"`, fail configuration validation.

### `src/childBridgeService.spec.ts`

- **Includes `hap.addIdentifyingMaterial` in the child bridge `LOAD` configuration**  
  Confirms that the setting is forwarded to child bridge processes.

### Validation

- `npm test` - 54 test files and 1,107 tests passed
- `npx vitest run src/bridgeService.spec.ts src/childBridgeService.spec.ts` - 2 test files and 121 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run lint-docs`
- `npm run build`

_EDIT: added validation info_

### DIFF
--- a/src/bridgeService.spec.ts
+++ b/src/bridgeService.spec.ts
@@ -895,9 +895,9 @@ describe('bridgeService', () => {
 
       validateHapConfig(cfg, { bridgeLabel: 'main bridge' })
 
-      expect(warnSpy).toHaveBeenCalledWith(expect.stringMatching(
-        /main bridge.*identifying material is disabled.*bridge names.*unique.*mDNS name collisions.*pairing instability/i,
-      ))
+      expect(warnSpy).toHaveBeenCalledOnce()
+      expect(warnSpy).toHaveBeenCalledWith(expect.stringMatching(/^main bridge:/))
+      expect(warnSpy).toHaveBeenCalledWith(expect.stringMatching(/identifying material is disabled/i))
     })
 
     it('does not warn when hap.disableIdentifyingMaterial is omitted or false', () => {
--- a/src/bridgeService.spec.ts
+++ b/src/bridgeService.spec.ts
@@ -6,7 +6,13 @@ import { Accessory, Categories, CharacteristicWarningType, uuid } from '@homebri
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { HomebridgeAPI, InternalAPIEvent } from './api.js'
-import { BridgeService, isHapConfigEnabled, isHapExternalsOnly, validateHapConfig } from './bridgeService.js'
+import {
+  BridgeService,
+  isHapConfigEnabled,
+  isHapExternalsOnly,
+  shouldAddIdentifyingMaterial,
+  validateHapConfig,
+} from './bridgeService.js'
 import { Logger } from './logger.js'
 import { PlatformAccessory } from './platformAccessory.js'
 
@@ -404,6 +410,25 @@ describe('bridgeService', () => {
       expect(externalPortService.requestPort).toHaveBeenCalledTimes(1)
     })
 
+    it('forwards hap.addIdentifyingMaterial to external accessories', async () => {
+      externalPortService.requestPort.mockResolvedValue(50002)
+      const service = new BridgeService(
+        api,
+        pluginManager,
+        externalPortService,
+        makeBridgeOptions(),
+        makeBridgeConfig({ hap: { addIdentifyingMaterial: false } }),
+      )
+
+      const accessory = makePlatformAccessory('External-Exact-Name')
+      const publishSpy = vi.spyOn(accessory._associatedHAPAccessory, 'publish').mockResolvedValue(undefined)
+
+      await service.handlePublishExternalAccessories([accessory])
+
+      const publishInfo = publishSpy.mock.calls[0][0] as any
+      expect(publishInfo.addIdentifyingMaterial).toBe(false)
+    })
+
     it('throws when an accessory address collides with an existing one', async () => {
       externalPortService.requestPort.mockResolvedValue(50000)
       const service = new BridgeService(api, pluginManager, externalPortService, makeBridgeOptions(), makeBridgeConfig())
@@ -729,6 +754,17 @@ describe('bridgeService', () => {
       expect(publishInfo.addIdentifyingMaterial).toBe(true)
     })
 
+    it('disables identifying material when configured in the HAP block', () => {
+      const config = makeBridgeConfig({ hap: { addIdentifyingMaterial: false } })
+      const service = new BridgeService(api, pluginManager, externalPortService, makeBridgeOptions(), config)
+      const publishSpy = vi.spyOn(service.bridge, 'publish').mockResolvedValue(undefined)
+
+      service.publishBridge()
+
+      const publishInfo = publishSpy.mock.calls[0][0] as any
+      expect(publishInfo.addIdentifyingMaterial).toBe(false)
+    })
+
     it('includes setupID when it is exactly 4 characters', () => {
       const config = makeBridgeConfig({ setupID: 'AB12' })
       const service = new BridgeService(api, pluginManager, externalPortService, makeBridgeOptions(), config)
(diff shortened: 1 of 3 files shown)


bwp91:
Thanks for this @naterator, and for the thorough write-up. This is a much better fit than the earlier attempt — using the existing `bridge.hap` block is the right home for it, and I appreciate that you covered both publish sites (main bridge *and* external accessories), the child bridge forwarding, config validation, and tests for all of it. That's the set of things that usually gets missed.

I'd like one change before we take it, plus a couple of smaller additions.

### Flip the option to a negative name with a `false` default

Rather than `addIdentifyingMaterial` defaulting to `true`, could we make it `disableIdentifyingMaterial` defaulting to `false`? So omitting it (or setting it `false`) keeps today's behaviour, and setting it `true` opts out.

Two reasons:

**1. Precedent.** We recently added `matter.disableIpv4` using exactly this shape — a negatively-named boolean that is `false` by default, where you have to explicitly set `true` to change behaviour. Having two config flags in neighbouring blocks with opposite defaulting conventions would be confusing for users and for us.

**2. It simplifies the code.** The current version needs an explicit "is it exactly `false`" check. With the flag inverted it becomes a plain falsy check, and it also removes a latent edge case — `typeof null === 'object'`, so `shouldAddIdentifyingMaterial(null)` currently throws if it is ever reached before validation:

```ts
export function shouldAddIdentifyingMaterial(hap: BridgeHapConfig | boolean | undefined): boolean {
  return typeof hap !== 'object' || !hap?.disableIdentifyingMaterial
}
```

Note the helper stays positively named, so the call sites still read `addIdentifyingMaterial: shouldAddIdentifyingMaterial(...)` with no double negative — only the config key is negative, which is the part that needs the `false` default. Worth a line in the JSDoc naming the underlying HAP-NodeJS `addIdentifyingMaterial` option, so the link to the library is still obvious to anyone reading it.

Everything else — the validation, the external accessory coverage, the child bridge forwarding, the test structure — can stay as it is; it should just be the rename, the inverted check, and flipping the test expectations.

### A warning when it is turned off

Those four characters exist to keep the advertised mDNS service name unique. With them disabled, two bridges that share a display name will advertise the same service name on the network, and mDNS name collisions are a common cause of pairing instability that is genuinely unpleasant to debug.

The option is still worth having — it is opt-in and the default does not change — but could we log a warning at startup when it is disabled, along the lines of "identifying material disabled, bridge names must be unique on your network"? That puts the caveat in front of the person who set it.

### Documentation

This adds a user-facing config option, so it will need a corresponding entry in the wiki alongside the other `hap` options, otherwise it will only be discoverable by reading the source.

### Note on the UI

Just so you are aware (not something you need to do): this will also need work in `homebridge-config-ui-x` before users can use it comfortably. The UI validates the bridge config against a schema, so a new key needs to be added there with a minimum-homebridge-version feature flag — otherwise the option shows as invalid in the config editor. We will handle that side separately once this lands.

Thanks again — happy to take this once the option is flipped round.

bwp91:
A couple of follow-ups to make that easier:

- **The wiki can wait.** No need to hold this PR up for it — documentation is easier to write once the code is released and the option is live, so let's do that afterwards.
- **Happy to help with the UI side.** If you fancy having a go at the `homebridge-config-ui-x` changes I'm glad to guide you through where they go, but equally I'm happy to just do all of it — no expectation on you at all, it's a separate repo and a different set of conventions.

naterator:
Thanks for the quick and detailed feedback @bwp91! I think I've implemented both of the aforementioned changes specific to this repo:

- flip the option to a negative name with a `false` default
- add a warning when it is turned off

The new `config.json` shape looks like:

```json
"bridge": {
  "name": "Homebridge",
  "hap": {
    "disableIdentifyingMaterial": true
  }
}
```

Regarding the other changes:

- **Wiki update.** Thanks for the heads-up on this. It looks like the wiki is restricted to certain editors. Should I add a request or `TODO` anywhere, or mention anything in Discord to trigger this wiki update after this PR lands?
- **UI addition.** I'm happy to implement this on the UI side. I'll keep an eye out for the proper `minimum-homebridge-version` to use whenever this PR lands and will submit a PR in `homebridge-config-ui-x` shortly thereafter.

Thanks for the help.

bwp91:
Thanks for turning that around so quickly @naterator — the flip looks exactly right, and nice touch adding the `null` case to the helper tests unprompted.

Two small things, then answers to your questions.

### The JSDoc has drifted slightly

The new comment says the material is "derived from the bridge name", but it is actually derived from the **username**. From HAP-NodeJS:

```js
this.displayName = this.displayName + " " + crypto.createHash("sha512")
    .update(info.username, "utf8")
    .digest("hex").slice(0, 4).toUpperCase();
```

Your original wording had this right. Worth fixing because it is the whole reason the warning exists: the suffix comes from the username, which is unique per bridge, so it is exactly what keeps two bridges *with the same display name* distinguishable on the network. If it were derived from the name it could not disambiguate anything.

### The warning test is a little over-specified

```
/main bridge.*identifying material is disabled.*bridge names.*unique.*mDNS name collisions.*pairing instability/i
```

That pins nearly the whole sentence, in order, so any future rewording of the log message — even an improvement to it — would break the test. Could we loosen it to something like `/identifying material is disabled/i`, and keep the strict part on the bridge label? Purely a maintenance nicety rather than a correctness issue.

### Wiki

Please don't worry about this one at all. You're right that the wiki is restricted, but it is no trouble for me to write it up after release, and I'm happy to just do it. If you would particularly like a hand in how it reads, feel free to drop the wording you would want as a comment here and I will use that as the basis — but there is genuinely no obligation, and no need to track it.

### UI minimum version

v2.3.0 is the plan for these changes. One wrinkle worth knowing: I will almost certainly put them out as a beta first, so the cleanest thing is to use that beta as the minimum version in the UI feature flag — a range like `>=2.3.0-beta.0` will match both the betas and the final release. I will confirm the exact version here once I have cut it, so you don't have to guess.

Thanks again, this is in good shape.

naterator:
Thanks for the additional feedback @bwp91. I believe I've implemented fixes for both of the items you mentioned now:

* **The JSDoc has drifted slightly.** I confused myself and my own understanding of the source of the bridge name when I was going through the code and made that adjustment (which is goofy, because I've used Homebridge for a _long_ time). Thanks for the explanation. It has been corrected.
* **The warning test is a little over-specified.** I've loosened this up. Let me know if there are any issues with the less-specific test implementation.

Thanks again for the assistance!

bwp91 (approved):
Thanks for making those changes!

bwp91:
@naterator — a correction on the version I gave you earlier, before you start on the UI side.

The betas will be versioned **`2.2.2-beta.x`**, and the stable release will then be **`v2.3.0`**. So the minimum version I mentioned before (`>=2.3.0-beta.0`) would not have worked — it would have missed the betas entirely.

The range you want in the feature flag registry is:

```
>=2.2.2-beta.0
```

That matches the `2.2.2-beta.x` betas and, because `2.3.0` is greater than `2.2.2`, the stable release and everything after it too. It also matches the style already used in that registry (`>=2.0.3-beta.18` and similar).

Sorry for the muddle — this is exactly the kind of thing I'd rather correct before you write the code than after.

Issue #3967, opened 2026-07-21
Bug: Matter PowerSource (battery) cluster only composed for RoboticVacuumCleaner -- silently dropped for all other device types


zbuc:
### Analysis

`AccessoryManager.buildCustomBehaviors()` composes the `PowerSource` cluster (from an accessory's `clusters.powerSource`) **only when the device type is `RoboticVacuumCleanerDevice`**. The `if (accessory.clusters?.powerSource) { ... }` block is nested inside `if (deviceType.deviceType === devices.RoboticVacuumCleanerDevice.deviceType) { ... }`, alongside the genuinely vacuum-specific `rvcCleanMode` and `serviceArea` handling.

As a result, a plugin that declares `clusters.powerSource` (battery) on **any other device type** -- DoorLock, ContactSensor, Thermostat, OccupancySensor, etc. -- silently gets **no PowerSource cluster** on its Matter endpoint, and Matter controllers (e.g. Home Assistant) never expose a battery entity. The cluster is a recognized key (it passes `validateAccessory` unchanged, no "Unknown cluster name" warning), so nothing signals that it was dropped.

This looks like accidental mis-scoping: `rvcCleanMode` / `serviceArea` are RVC-specific and belong in that block, but `powerSource` is a general utility cluster. The parallel case -- `ElectricalPowerMeasurement` / `ElectricalEnergyMeasurement` -- is handled at the **top level** (`applyElectricalMeasurementClusters` runs for any device type, main endpoint and parts). `powerSource` is the outlier.

**Fix:** move the `if (accessory.clusters?.powerSource) { ... }` block out of the `RoboticVacuumCleanerDevice` conditional so it applies to all device types, like the electrical-measurement clusters. The block's existing feature selection (`PowerSourceServer.with('Battery')` / `.with('Battery','Rechargeable')` based on the declared attributes) is unchanged -- only its scope moves.

I have a verified patch here: `zbuc/homebridge#fix/matter-powersource-all-devicetypes`.

### Expected Behavior

When a plugin declares `clusters.powerSource` on an accessory of any device type (not just RoboticVacuumCleaner), Homebridge should compose the Matter `PowerSource` cluster (`0x2F`) on that endpoint -- the same way it composes `ElectricalPowerMeasurement` / `ElectricalEnergyMeasurement` for any device type -- so the battery is exposed over Matter and controllers surface a battery entity.

### Steps To Reproduce

[...]

PR #3968, opened 2026-07-21
fix(matter): compose PowerSource for all device types

zbuc:

  `AccessoryManager.buildCustomBehaviors()` composes the Matter `PowerSource`
  cluster **only inside the `RoboticVacuumCleanerDevice` branch**, alongside the
  genuinely vacuum-specific `rvcCleanMode` / `serviceArea` handling:

  ```ts
  if (deviceType.deviceType === devices.RoboticVacuumCleanerDevice.deviceType) {
    // ...rvcCleanMode / serviceArea...
    if (accessory.clusters?.powerSource) {
      // compose PowerSource
    }
  }
```

  So a plugin that declares clusters.powerSource (battery) on any other device
  type — DoorLock, ContactSensor, LeakSensor, OccupancySensor, Thermostat, … —
  silently gets no PowerSource cluster on its Matter endpoint, and controllers
  (e.g. Home Assistant) never expose a battery entity. powerSource is a
  recognized key that passes validateAccessory unchanged, so nothing warns that
  it was dropped — buildCustomBehaviors just never reaches it.

  There is also a second, subtler gap: even at the top level, a battery-only,
  read-only accessory (a sensor) declares no handlers, so it would be caught by
  the if (!accessory.handlers) return customBehaviors early return before any
  battery is composed.

  Reported and diagnosed in #3967, including before/after cluster reads
  from a real bridge.

## :bulb: Proposed solution

  Treat powerSource as the general utility cluster it is — the same way
  ElectricalPowerMeasurement / ElectricalEnergyMeasurement are applied at the
  top level for any device type.

  - Move the if (accessory.clusters?.powerSource) { … } block out of the
  RoboticVacuumCleanerDevice branch.
  - Place it at the top of buildCustomBehaviors, above the !accessory.handlers
  early return, so it also composes for read-only, handler-less accessories
  (battery sensors).

  This is a pure move — the feature selection
  (PowerSourceServer.with('Battery') / .with('Battery', 'Rechargeable') based
  on the declared attributes) is unchanged; only the scope and position of the
  block change. No new API surface.

## :gear: Release Notes

  Fixed: plugins declaring a powerSource (battery) cluster via api.matter now
  expose it over Matter for all device types, not just RoboticVacuumCleaner —
  including read-only sensors that declare no handlers. Battery-powered locks and
  contact/leak/occupancy sensors now surface a battery entity on Matter controllers
  (e.g. Home Assistant), matching what they already expose over HAP.

  No breaking changes: this only adds a cluster that was previously dropped;
  accessories that don't declare powerSource are unaffected, and RVC accessories
  behave exactly as before.
accessories that don't declare powerSource are unaffected, and RVC accessories
behave exactly as before.

Verified end-to-end against a real bridge (3× August locks via a fork of
homebridge-august declaring clusters.powerSource on the DoorLock
accessories): buildCustomBehaviors went from Applied 1 custom behavior(s) to
Applied 2, cluster 0x2F now appears on each lock endpoint with the Battery
feature (…/47/65532 = 2, …/47/12 = batPercentRemaining), and Home Assistant
created a working battery sensor.

## Testing

Added three cases to src/matter/server/AccessoryManager.spec.ts, all asserting
on the mocked PowerSourceServer.with:

- non-RVC device with handlers (a lock) → composes .with('Battery').
- non-RVC device without handlers (a read-only sensor) → still composes
.with('Battery'). This is the explicit regression guard for the
!accessory.handlers early return; without the fix it composes nothing.
- rechargeable variant (a batChargeState is declared) → composes
.with('Battery', 'Rechargeable').

Existing tests are unchanged and still pass (20/20). npm run typecheck and
eslint are clean. Not covered by unit tests: the live commissioning /
controller-side battery mapping, which was instead verified manually as described
above.

## Reviewer Nudging

Start in src/matter/server/AccessoryManager.ts, buildCustomBehaviors() — the
whole change is the relocated powerSource block (now at the top of the method,
above the !accessory.handlers early return) and its removal from the
RoboticVacuumCleanerDevice branch. The diff reads cleanly as a move. Then see
the new describe('powerSource (battery) composition', …) block in the spec,
where the no-handlers case is the one that pins the early-return behavior.

### DIFF
--- a/src/matter/server/AccessoryManager.spec.ts
+++ b/src/matter/server/AccessoryManager.spec.ts
@@ -1,6 +1,7 @@
 import type { InternalMatterAccessory, MatterAccessory } from '../types.js'
 import type { AccessoryManagerDeps } from './AccessoryManager.js'
 
+import { PowerSourceServer } from '@matter/node/behaviors'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import {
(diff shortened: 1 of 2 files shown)


bwp91 (approved):
thank you for finding and fixing this :)

bwp91:
This is now included in `v2.2.2-beta.3` and will be released in the near future in `v2.3.0` 😊

PR #3969, opened 2026-07-21
fix(matter): restore cached accessories into the bridge before going online

keremerkan:
## Description

**The problem.** On every restart of a commissioned Matter bridge, the server
node comes online with an empty aggregator parts list — cached accessories are
only delivered to plugins via `configureMatterAccessory`, never rebuilt into
the bridge. A paired controller re-establishes its subscription within the
first second (log excerpt below, Apple TV hub), reads a parts list containing
0-2 of 11 devices, treats the missing endpoints as removed, and discards their
metadata: **room assignments and tile settings are lost on nearly every
restart**, with the devices re-added as "new" seconds later when plugin
registrations land. Endpoint identity is not the issue — endpoint numbers and
`BridgedDeviceBasicInformation.uniqueId` are byte-stable across restarts
(verified from matter.js storage); the transient parts-list shrink alone
triggers the controller-side deletion.

```
11:05:00 PM [Matter/Node] 0E8093EB8676 is online
11:05:01 PM [Matter/InteractionServer] Subscription successfully reestablished ...
11:05:01 PM [Matter/Server] Registered Matter accessory: (1 of 11)   <- too late
...
11:05:11 PM [Matter/Server] Registered Matter accessory: (11 of 11)
```

There is no race plugins can win: even batching all registrations into one
call loses to a same-second subscription. HAP solved the equivalent problem
years ago by restoring cached accessories into the bridge before publishing;
Matter needs the same.

**The fix.**

1. `ServerLifecycle.start()` invokes a new optional dep,
   `restoreAccessoriesFromCache()`, after the aggregator is created and
   **before** `startServerNode()`.
2. `MatterServer` implements it: load the accessory cache, rebuild each cached
   accessory (device types resolved back through the public `deviceTypes`
   registry from their cached names) and register it through the normal
   `AccessoryManager.registerAccessory` path, marked with an internal
   `_restoredFromCache` flag. Handlers cannot be cached, but custom behaviors
   (e.g. the OnOff server on switch device types) are only attached for
   clusters that have handlers — empty stubs are synthesized per cached
   cluster so the restored endpoint is structurally identical to the original.
3. `AccessoryManager.registerAccessory`: when a plugin re-registers a UUID
   whose accessory is `_restoredFromCache`, attach the plugin's handlers and
   metadata to the existing endpoint in place (no parts-list churn) instead of
   throwing the duplicate-UUID error. If the structure changed (device type or
   part ids), unregister the restored endpoint and register fresh.

**Also fixed as a side effect:** `restoreCachedAccessories()` currently logs
`Restoring 0 cached Matter accessories` even with a populated
`accessories.json`, because the cache is only loaded at the end of
`startServerNode()` — after the synchronous restore call sites in `server.ts`
/ `childBridgeFork.ts` have already read it. The pre-online hook loads the
cache first, so `configureMatterAccessory` now fires with the cached
accessories as designed.

**Testing.** Live bridge (child bridge, `hap.enabled: false`), 11 accessories
covering single-endpoint devices, a composed BridgedNode with typed parts, and
electrical measurement clusters, paired with an Apple Home hub (tvOS 27):

- restore completes before `going online`; the controller's first read sees all 11
- plugin re-registration attaches in place: no duplicate-UUID errors, no
  "Cluster 'onOff' not found", commands work end to end
- endpoint numbers and uniqueIds byte-identical across restarts
- room assignments survive repeated bridge restarts (previously lost)
- fresh-pair (empty cache) path unchanged; structural-change path re-registers

**Notes for reviewers.**

- Stub handlers are a workaround for handlers being non-serializable. A
  cleaner follow-up would be caching the handler *shape* (cluster/command
  names) in `SerializedMatterAccessory`, but the stubs keep the cache format
  unchanged and behave identically for behavior selection.
- Restored accessories could be marked `reachable: false` until the plugin
  attaches; left out to keep the diff minimal.
- Related but not fixed here (can file separately): registrations arriving
  while the Matter server is starting are logged
  (`MatterDeviceError: Matter server not started`) but the API call resolves,
  so plugins cannot detect the drop; storage-lock contention on a quick
  restart permanently disables Matter for the process lifetime with no retry;
  and `notifyPartsListChanged`'s `increaseConfigurationVersion` acquires locks
  synchronously and drops notifications under concurrent transactions.

### DIFF
--- a/src/matter/server.ts
+++ b/src/matter/server.ts
@@ -5,7 +5,7 @@
  * All public method signatures are preserved for external callers.
  */
 
-import type { ServerNode } from '@matter/main'
+import type { EndpointType, ServerNode } from '@matter/main'
 
 import type { SerializedMatterAccessory } from './accessoryCache.js'
 import type { MatterServerConfig } from './sharedTypes.js'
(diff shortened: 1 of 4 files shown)


bwp91:
Thanks @keremerkan — this is an excellent piece of work. The diagnosis is correct and the fix is the right shape: HAP has always restored cached accessories into the bridge before publishing (`addBridgedAccessory` in `publishBridge`), and Matter's restore only ever notified the plugin via `configureMatterAccessory` without rebuilding the endpoints, so the aggregator came online almost empty and controllers pruned the "missing" devices. Rebuilding the endpoints before `startServerNode()` and attaching the plugin's later registration in place is exactly the parallel to HAP.

I've been through the change and traced a few things I wanted to be sure of, all of which hold up:

- **No double-registration of parts.** `registerAccessoryHandlers` only wires the main endpoint; the attach path deliberately wires part handlers to the *existing* part endpoints rather than recreating them. Correct.
- **`restoreCachedState` runs once** — the attach path returns before reaching it.
- **The stub-handlers approach is sound.** Empty `{}` per cached cluster is enough for `buildCustomBehaviors` to attach the right server (e.g. OnOff), so the restored endpoint is structurally identical, and the real handler routing is wired into `behaviorRegistry` on re-registration — which is what the already-built matter.js behaviors call into.

Typecheck, lint and the full matter suite (820 tests) pass on the branch.

Two things I checked in the code so they don't need chasing:

- **Orphan handling is fine.** The pre-online restore rebuilds every cached accessory unconditionally, but `restoreCachedAccessories(keepOrphaned)` still runs immediately after in `server.ts`, and its `unregisterAccessory` now reaches the freshly-built endpoints, so orphans from an uninstalled plugin are still removed and `keepOrphanedCachedAccessories` is still honoured. The only cost is that an orphan exists as an endpoint for the brief window between the two calls during startup — acceptable, and no worse than today.
- **The `reachable: false` note is a non-issue.** Homebridge never manages Matter reachability dynamically — every endpoint (and the bridge node) is statically `reachable: true` and nothing ever flips it. So leaving restored endpoints reachable is consistent with the rest of the codebase; adding false-until-attached would have been entirely new machinery. Right call to leave it out.

I'll add unit test coverage for the new branches (attach-in-place same-shape, structural-change re-register, and the still-throwing plain duplicate) after this lands, and I'll open separate tickets for the three follow-ups you flagged in your notes — the registrations dropped during server-start, the storage-lock contention disabling Matter, and the `notifyPartsListChanged` lock contention. All three are worth pursuing. Thanks for calling them out.

bwp91 (approved):
🎉

PR #3972, opened 2026-07-23
fix(matter): add FixedLabel and PowerSource to composed parents

keremerkan:
## Summary

Any Matter accessory registered with `parts` (a BridgedNode parent with child
endpoints) is uncontrollable from Apple Home beyond the first ~30 seconds
after pairing. Reads keep working indefinitely; every command fails silently
("No Response") with no error on either side. Flat accessories are unaffected.

Root cause (established by log-streaming the Apple TV home hub's `homed`
daemon and byte-diffing Matter descriptor dumps against a working reference
bridge): Apple's controller cannot build its internal model for a composed
bridged device unless the parent endpoint also exposes a **FixedLabel**
cluster and a **PowerSource** cluster. Without them, homed's per-accessory
session setup silently stalls:

- the child endpoints never bind to the composed accessory,
- homed's Matter-report→characteristic translation produces zero updates for
  the whole node (its `MTRDevice` layer receives our reports fine),
- the accessory-info model never populates, the accessory is never
  "configured", and homed's *Add accessory pairing operation* times out after
  ~30 s with `HMErrorDomain Code=4 (accessoryNotReachable)`.

The ~30 s of working control after pairing is the commissioning iPhone
controlling the accessory directly while that operation is still pending —
which makes the failure look like a mysterious timeout rather than a
modeling error. Nothing is logged at any level a bridge operator can see.

The fix is two clusters on the parent endpoint of parts-bearing accessories:

- `FixedLabel` with `labelList: [{ label: "composed", value: <kind> }]`
- `PowerSource` (Wired feature) with static wired-AC state

This matches what the matterbridge project has always exposed on composed
parents (`addFixedLabel('composed', …)` and
`createDefaultPowerSourceWiredClusterServer()`), which is why bridges built
on it never exhibited the failure.

## Evidence

- Descriptor diff of a WORKING composed parent (matterbridge, same Apple
  hub/LAN) vs a Homebridge composed parent:
  ServerList `[Descriptor, BridgedDeviceBasicInformation, PowerSource,
  FixedLabel]` vs `[BridgedDeviceBasicInformation, Descriptor]`. Child
  endpoints are equivalent.
- Bisect on a 40-line test plugin (included as the reproducer below): flat
  OnOffLight works; the same accessory as BridgedNode+part fails; adding
  FixedLabel alone → accessory no longer surfaces in Home; adding
  FixedLabel + PowerSource → works indefinitely (verified >2 min of
  continuous toggling, including hub-routed commands with the phone on
  cellular).
- Hub-side marker: `homed`'s `HMDMatterAttributeChangedNotification` count
  for the node goes from 0 (broken, all captures) to hundreds (fixed),
  with `Configuring with HAPAccessory` appearing only once the clusters are
  present.
- With the fix, an 11-device production bridge (mixed flat/composed, mixed
  light/outlet/switch types) survives everything that previously failed:
  control past the timeout, Homebridge restarts with room assignments
  intact, and home-hub elections (which previously re-adopted composed
  accessories as new, resetting rooms, every time).

Environment: Homebridge v2.2.2-beta.5, @matter 0.17.5, Apple Home on
iOS/tvOS 27 beta (multiple hubs). The same failure signature was present on
earlier tvOS-26.x observations of composed accessories.

## Implementation

All changes are in `AccessoryManager` (validated as a dist patch running in
production; this PR ports it to the TypeScript sources):

- **FixedLabel + PowerSource on parts-bearing parents.** After
  `createEndpointOptions()`, a parts-bearing accessory whose device type does
  not already carry `fixedLabel` gets
  `FixedLabelServer` (with `labelList: [{ label: 'composed', value: <kind> }]`)
  and `PowerSourceServer.with('Wired')` (static wired-AC state) composed onto
  the parent device type. The `<kind>` value is derived from the first part's
  device type name mapped to a human word (Light/Outlet/Switch, else Device) —
  the validated dist patch hardcoded one value and Apple did not appear to care
  about the string, but the derived label matches what matterbridge exposes.
- **BridgedDeviceBasicInformation completeness**, bundled here because it is
  the same "Apple builds its accessory model from this metadata" class of fix,
  discovered in the same investigation:
  - `softwareVersion`/`softwareVersionString` parsed from the accessory's
    `firmwareRevision` (`x.y.z` triplet, bytes packed into the numeric
    version; `1.0.0` fallback). Apple homed refuses to register accessories
    whose firmware metadata is missing or unparseable ("invalid matter AFU
    settings") — which silently breaks its whole control path.
  - `vendorId` (test vendor ID, matching the bridge's own), `hardwareVersion`,
    `configurationVersion`, `productUrl`.
  - deterministic `uniqueId` derived by hashing the accessory UUID (truncated sha256). matter.js otherwise fills
    it with a random string persisted only in its own storage; controllers key
    bridged accessory identity on it, so deriving it from the accessory UUID
    makes the same registration always yield the same identity.
  - `leave` and `reachableChanged` events enabled on the
    BridgedDeviceBasicInformationServer.
- **Descriptor TagList on part endpoints**: each child gets a semantic Number
  tag (`namespaceId: 7`, `tag` = part index, `label` = part display name) so
  controllers can stably re-map otherwise-identical children of a composed
  device; without it they are distinguishable only by transient endpoint
  number, which Apple Home mishandles across hub changes.

Both parent clusters are cheap static state; no runtime maintenance needed.
PowerSource here describes the bridge device's mains supply; if the API later
grows battery-state support, the plugin should be able to override.

## Reproducer

Minimal plugin (~40 lines, one virtual OnOffLight, no hardware needed) that
registers the accessory as BridgedNode+part (fails without this fix; register
the part's device type flat instead and it works). Pair the child bridge with
Apple Home, toggle for 60 s. Failure mode: toggles work ~30 s, then permanent
"No Response" while the tile still shows live state.

`index.js`:

```js
module.exports = (api) => {
  api.registerPlatform("homebridge-matter-minimal", "MatterMinimal", class {
    constructor(log, config, api) {
      this.log = log;
      this.api = api;
      this.state = false;
      api.on("didFinishLaunching", () => { this.register().catch((e) => log.error(String(e))); });
    }
    configureMatterAccessory() {}
    async register() {
      const uuid = this.api.hap.uuid.generate("matter-minimal-light-1");
      const accessory = {
        UUID: uuid,
        displayName: "Minimal Light",
        serialNumber: "MIN-0001",
        manufacturer: "Minimal",
        model: "MinimalLight",
        firmwareRevision: "1.0.0",
        context: {},
        deviceType: this.api.matter.deviceTypes.BridgedNode,
        parts: [{
          id: "switch-0-light",
          displayName: "Minimal Light",
          deviceType: this.api.matter.deviceTypes.OnOffLight,
          clusters: { onOff: { onOff: this.state } },
          handlers: {
            onOff: {
              on: () => { this.state = true; this.log.info("Minimal Light ON"); },
              off: () => { this.state = false; this.log.info("Minimal Light OFF"); },
            },
          },
        }],
      };
      for (let i = 0; i < 24; i++) {
        try {
          await this.api.matter.registerPlatformAccessories("homebridge-matter-minimal", "MatterMinimal", [accessory]);
          const ok = await this.api.matter.getAccessoryState(uuid, "onOff", "switch-0-light");
          if (ok !== undefined) { this.log.info("Minimal Light registered."); return; }
        } catch (e) { this.log.debug("retrying registration: " + e.message); }
        await new Promise((r) => setTimeout(r, 5000));
      }
      this.log.error("Minimal Light never registered");
    }
  });
};
```

`package.json`:

```json
{
  "name": "homebridge-matter-minimal",
  "version": "0.0.1",
  "description": "Minimal Matter test plugin - one static OnOffLight",
  "main": "index.js",
  "engines": { "homebridge": ">=2.2.0" },
  "keywords": ["homebridge-plugin"]
}
```

## Notes for reviewers

- Same investigation lineage as #3969/#3970; we can run any scenario on our
  deployment (11 Shelly devices, Apple Home, multiple tvOS 27 hubs).
- This is Apple-behavior-driven, not spec-mandated — the spec does not
  require these clusters on composed parents. We plan to file the silent
  30-second-timeout failure mode with Apple separately; but exposing the
  clusters is the pragmatic fix and matches the only bridge stack Apple
  demonstrably works with.
- Related cosmetic fix worth bundling separately: the "No custom behavior
  class available for cluster 'electricalPowerMeasurement'" warning fires for
  attribute-only clusters that have no commands; it should be suppressed
  for clusters without accepted commands.

### DIFF
--- a/src/matter/server/AccessoryManager.spec.ts
+++ b/src/matter/server/AccessoryManager.spec.ts
@@ -26,12 +26,19 @@ vi.mock('@matter/main', () => {
       this.id = options?.id
     }
   }
-  return { Endpoint: MockEndpoint }
+  return { Endpoint: MockEndpoint, VendorId: vi.fn((id: number) => id) }
 })
 vi.mock('@matter/main/behaviors', () => ({
   BasicInformationServer: { name: 'BasicInformationServer' },
-  BridgedDeviceBasicInformationServer: { name: 'BridgedDeviceBasicInformationServer' },
-  DescriptorServer: { name: 'DescriptorServer' },
+  BridgedDeviceBasicInformationServer: {
+    name: 'BridgedDeviceBasicInformationServer',
+    enable: vi.fn(() => ({ name: 'BridgedDeviceBasicInformationServer.enable' })),
+  },
+  DescriptorServer: {
+    name: 'DescriptorServer',
+    with: vi.fn((...args: any[]) => ({ name: `DescriptorServer.with(${args.join(',')})` })),
+  },
+  FixedLabelServer: { name: 'FixedLabelServer' },
 }))
 vi.mock('@matter/node/behaviors', () => ({
   PowerSourceServer: { name: 'PowerSourceServer', with: vi.fn((...args: any[]) => ({ name: `PowerSourceServer.with(${args.join(',')})` })) },
@@ -171,11 +178,18 @@ function createMockDeps(overrides: Partial<AccessoryManagerDeps> = {}): Accessor
   }
 }
 
+// Composed (parts-bearing) accessories chain several `.with()` calls on the
+// parent device type (BridgedDeviceBasicInformation, FixedLabel, PowerSource),
+// so the mock must stay chainable at any depth.
+function createChainableDeviceType(props: Record<string, unknown>): any {
+  return { ...props, with: vi.fn(() => createChainableDeviceType(props)) }
+}
+
 function createMockAccessory(overrides: Partial<MatterAccessory> = {}): MatterAccessory {
   return {
     UUID: 'test-uuid-001',
     displayName: 'Test Light',
-    deviceType: { deviceType: 0x0100, name: 'OnOffLight', with: vi.fn(() => ({ deviceType: 0x0100, with: vi.fn() })) } as any,
+    deviceType: createChainableDeviceType({ deviceType: 0x0100, name: 'OnOffLight' }),
     serialNumber: 'SN-001',
     manufacturer: 'Test Mfg',
     model: 'Test Model',
--- a/src/matter/server/AccessoryManager.spec.ts
+++ b/src/matter/server/AccessoryManager.spec.ts
@@ -1,6 +1,7 @@
 import type { InternalMatterAccessory, MatterAccessory } from '../types.js'
 import type { AccessoryManagerDeps } from './AccessoryManager.js'
 
+import { DescriptorServer, FixedLabelServer } from '@matter/main/behaviors'
 import { PowerSourceServer } from '@matter/node/behaviors'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
@@ -183,6 +184,26 @@ function createMockDeps(overrides: Partial<AccessoryManagerDeps> = {}): Accessor
 // so the mock must stay chainable at any depth.
 function createChainableDeviceType(props: Record<string, unknown>): any {
   return { ...props, with: vi.fn(() => createChainableDeviceType(props)) }
+}
+
+// A parent device type whose `.with()` is a single self-returning spy, so a
+// test can assert exactly which behaviors (FixedLabel, wired PowerSource) were
+// composed onto it across the several chained `.with()` calls registerAccessory
+// makes on the parent.
+function createSpyDeviceType(props: Record<string, unknown>): any {
+  const deviceType: any = { ...props }
+  deviceType.with = vi.fn(() => deviceType)
+  return deviceType
+}
+
+// A composed-device part: a plain device type plus a cluster, no handlers.
+function createComposedPart(id = 'part-1'): any {
+  return {
+    id,
+    displayName: `Part ${id}`,
+    deviceType: createChainableDeviceType({ deviceType: 0x0100, name: 'OnOffLight' }),
+    clusters: { onOff: { onOff: false } },
+  }
 }
 
 function createMockAccessory(overrides: Partial<MatterAccessory> = {}): MatterAccessory {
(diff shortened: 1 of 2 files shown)


bwp91:
Thank you for this — the investigation behind it is outstanding. Byte-diffing the descriptors against a working matterbridge parent, and the bisect showing that `FixedLabel` on its own is not enough, is far better evidence than we usually get for an Apple-side behaviour. I am confident the core finding is right.

I checked the details against the matter.js version we ship, and these all came back correct:

- Every attribute you set really does exist on `BridgedDeviceBasicInformation` (`configurationVersion`, `productUrl`, `uniqueId`, `hardwareVersionString`).
- `namespaceId: 7` really is `CommonNumber` (`@matter/types`, `Namespace.d.ts:57`).
- On the part endpoints you set `descriptor` **before** spreading `...part.clusters`, so a plugin can still override it. That is the right way round.
- The `FixedLabel` strings are inside the 16 character limit.

Below are the things that need fixing first. I have put them in order of importance.

---

## 1. Battery reporting is wiped on any composed accessory (must fix)

This is the one that blocks the PR.

We released `c6050fd` (#3968) a few days ago to stop the PowerSource cluster being dropped. This PR undoes that for any accessory that has `parts`.

Here is what happens today when an accessory declares `clusters.powerSource`:

- `AccessoryManager.ts:491` adds `PowerSourceServer.with('Battery')`.
- `AccessoryManager.ts:604` spreads `...accessory.clusters`, so `endpointOptions.powerSource` holds the battery state.

Your new code then runs **after** both of those, and for any accessory with parts it:

- adds `PowerSourceServer.with('Wired')`, and
- overwrites `endpointOptions.powerSource` completely with the wired values.

Both the battery feature and the battery state are lost. matter.js merges behaviours by simple key assignment — in `SupportedBehaviors.js` it does `target[type.id] = type`. Both versions use the id `powerSource`, so the last one added simply replaces the first.

The result: a battery powered accessory that uses `parts` stops reporting its battery at all.

**What to change:** only add the wired PowerSource when the accessory has not declared its own `powerSource`. Something like:

```ts
if (!accessory.clusters?.powerSource) {
  deviceType = (deviceType as any).with((PowerSourceServer as any).with('Wired'))
  endpointOptions.powerSource = { /* wired defaults */ }
}
```

A test for "composed accessory with a battery keeps its battery" would be good to have alongside it.

---

## 2. The current type says AC but the value means DC (must fix)

```ts
wiredCurrentType: 1, // AC
```

In `@matter/types`, `power-source.d.ts:920`:

- `Ac = 0`
- `Dc = 1`

So this sends DC to controllers while the comment says AC.

**What to change:** use `0` for AC.

---

## 3. Changing `uniqueId` will re-key every accessory people already have paired (needs discussion)

Your reasoning is that controllers use `uniqueId` to decide the identity of a bridged accessory. I agree with that. But the same reasoning means that **changing** it changes the identity of every accessory that is already paired.

For everyone upgrading an existing bridge, Apple Home may treat all their accessories as brand new ones. That would lose room assignments, automations and scenes — which is the exact problem this work is trying to solve.

Your testing notes say room assignments survive restarts, but that is with the change already in place. The risky moment is the single upgrade from an old version to a new one.

**What we need to know:** did you test upgrading a bridge that was already commissioned *before* the change, rather than pairing fresh afterwards? If the accessories came back as new devices, we need a migration plan before this can ship.

---

## 4. There are no tests for the new behaviour

The changes to `AccessoryManager.spec.ts` are all mock plumbing to keep the existing tests passing — the chainable device type, `DescriptorServer.with`, `FixedLabelServer` and `VendorId`. There are no new assertions.

This change alters what every composed accessory advertises, so please add tests covering:

- a composed accessory gets `FixedLabel` and a wired `PowerSource`
- a composed accessory that declares its own battery **keeps** it (point 1 above)
- a flat accessory is unchanged
- part endpoints get their tag list

---

## Smaller points

- **`composedKindLabel()`** — you say Apple does not care about the string, so this function adds risk for no benefit. It also has a small bug: `includes('Light')` is checked before `includes('Switch')`, so a light switch device type returns "Light". A fixed string would be simpler and just as correct.
- **`vendorId: VendorId(0xFFF1)`** — this now claims the Matter *test* vendor ID for every third party accessory we bridge, where before it was unset. Probably harmless, but please add a line explaining why it is needed.
- **`hardwareVersion: 1` and `hardwareVersionString: '1.0.0'`** are invented values. Fine if Apple needs something there, but worth a comment saying so.
- **`.enable({ events: ... })`** is only applied on the path where we add the server ourselves. A `BridgedNodeEndpoint` that already carries the cluster does not get the events, so the behaviour is inconsistent.
- **Branch** — this targets `latest`, but our current Matter work is on `beta-2.2.2`. Expect a conflict in `AccessoryManager.ts` when the two meet.

---

## A suggestion: please split this into two PRs

The two halves have very different risk levels, and I would like to get the good half in quickly.

- **PR A — the composed parent fix.** `FixedLabel` plus wired `PowerSource`, and the part tag list. This is the well evidenced part. Once points 1, 2 and 4 are done it can go in fast.
- **PR B — the `BridgedDeviceBasicInformation` metadata.** Especially `uniqueId`. This one needs the upgrade path in point 3 sorted out first, and it should not hold up PR A.

Thanks again — this is really valuable work, and I would like to land the first half soon.

keremerkan:
Thanks for the thorough review — all addressed, and I took your split suggestion further.

- **Battery regression:** the wired `PowerSource` is now only synthesized when the accessory hasn't declared its own `clusters.powerSource`, so a plugin battery is never overwritten (#3968 stays intact).
- **AC enum:** `wiredCurrentType` is now `0` (AC).
- **`composedKindLabel`:** dropped for a fixed string.
- **Tests:** added for composed→FixedLabel+wired PowerSource, composed-with-battery keeps its battery, flat unchanged, and part tag lists.

On the `uniqueId`/BDBI-metadata half: rather than sort out a migration, I dropped it entirely. It was never isolated as the fix in my bisect (FixedLabel + PowerSource is what actually mattered), its benefit is marginal, and dropping it sidesteps the upgrade re-key risk you flagged — so there's no PR B. This PR is now just FixedLabel + wired PowerSource + part tag list. Verified live on an 11-device bridge: controls survive the ~30s cliff, restarts, and hub switches with the slimmed version.

Rebased onto `beta-2.2.2` and retargeted.

bwp91:
Thanks — this all checks out, and dropping the `uniqueId`/metadata half entirely was the right call. It keeps the PR to the part your bisect actually proved, and removes the upgrade risk completely.

I checked out the branch and ran it locally:

- `npm run lint` — clean
- full Matter suite — **852/852 passing**

I also wanted to be sure the new battery test really protects us, so I deliberately broke the guard (made the wired `PowerSource` apply unconditionally again) and re-ran. The test failed exactly as it should:

```
× a composed accessory that declares its own PowerSource (battery) keeps it and gets no wired PowerSource
```

So the #3968 regression genuinely cannot come back unnoticed. That is what I was hoping for.

## One tiny thing left

There is a leftover comment that no longer matches the code. In `AccessoryManager.ts`:

```ts
// Enable the Leave and ReachableChanged events - controllers use them
// to track bridged device lifecycle, and known-good bridges expose both.
deviceType = (deviceType as any).with(BridgedDeviceBasicInformationServer)
```

The `.enable({ events: ... })` call it describes was part of the metadata half you dropped, so the code below it no longer enables any events. Please just delete those two comment lines.

That is the only thing outstanding. Once it is gone I am happy to merge this.

Thanks again for the very careful investigation — the descriptor diff against a working bridge is what made this reviewable at all.

keremerkan:
Done — the stale comment is removed (674766a). Lint clean, suite green.

bwp91:
@keremerkan did you mean to include a `.tgz` file as part of your last commit?

keremerkan:
@bwp91 sorry, it got swept into the commit with git -A while removing the stale comment. Amended the commit and cleaned it up now.

bwp91 (approved):
Happy with this one — merging it into the beta branch, so it will go out in the next published release alongside #3973.

Issue #3970, opened 2026-07-22
Matter: startup/robustness gaps — silent registration drop, no retry after start failure, dropped parts-list notifications


bwp91:
Three related Matter startup/robustness gaps, split out from the review of #3969 (originally observed by @keremerkan). I've confirmed all three against the code on `beta-2.2.2`; line numbers below are from that branch.

---

### 1. Bridged Matter registrations can be silently dropped

`api.matter.registerPlatformAccessories()` guards with `assertMatterReady()`, which only checks that the Matter manager exists and `hasActiveMatter()` is true (`MatterAPIImpl.ts`). But `hasActiveMatter()` returns `this.matterServer !== undefined` (`MatterBridgeManager.ts:133`), and `matterServer` is assigned (`:251`) *before* `await matterServer.start()` completes (`:266`). So there is a window where the guard passes but the server node is not up.

For bridged accessories the registration is emitted fire-and-forget (`MatterAPIImpl.ts:322`); the async handler reaches `AccessoryManager.registerAccessory()`, which throws `MatterDeviceError('Matter server not started')` (`:108`). Because the emit is not awaited, that rejection never propagates back to the plugin's `registerPlatformAccessories()` promise — **it resolves successfully**. The plugin believes the accessory registered; it did not. External accessories are unaffected (they await a resolver). The existing comment at `MatterAPIImpl.ts:239` already acknowledges this class of "bridged registrations are silently dropped".

**Fix direction:** either have `hasActiveMatter()` reflect that the server node is actually started, or make the bridged path surface the handler's failure back to the caller (await it, or reject the returned promise).

---

### 2. A Matter server start failure disables Matter for the whole process, with no retry

`MatterBridgeManager.initialize()`'s catch (`:281`) logs a friendly banner for known cases (corrupted storage, `EADDRINUSE`) and then simply returns. It never sets `_setMatterEnabled(true)`, and there is no scheduled retry — Matter stays off until the process is restarted.

The storage-lock case is the concerning one: on a quick restart, matter.js can fail to acquire the storage directory lock if the previous process is still releasing it. That error is **not** matched by the corrupted-storage auto-recovery (`ServerLifecycle.ts:93` only matches `Invalid public key encoding` / `FabricManager unavailable` / `key-input`), so it falls through to the catch and permanently disables Matter for the process lifetime — a transient, self-resolving condition turned into a hard failure requiring manual intervention.

[...]

PR #3973, opened 2026-07-23
fix(matter): defer bridge online until registrations settle

keremerkan:
## Summary

When a commissioned matter.js node comes online,
`SubscriptionsServer.reestablishFormerSubscriptions()` proactively reconnects
every controller that had a persisted subscription (2-second connection
timeout per peer, no retry). When this succeeds, a bridge restart is
invisible: the hub is re-subscribed within seconds and bridged accessories
never drop out. When it fails, Apple hubs keep transmitting on their dead
CASE sessions ("Ignoring message for unknown session") for minutes — visible
to users as every bridged device going "No Response" after each Homebridge
restart, historically with room assignments being lost.

Today Homebridge undermines this mechanism in two ways:

1. **The node goes online ~200 ms after creation**, while plugin
   registrations stream in immediately afterwards. Their transactions
   (`aggregator.add`, handler attachment, `increaseConfigurationVersion()`)
   land inside the 2-second window and abort re-establishment
   ("Reestablished 0 of 2 former subscriptions", "Operation aborted").
   Standalone bridges (matterbridge) never hit this because they build the
   complete aggregator before starting the node.
2. **Persisted subscriptions for deleted fabrics are never purged**, so
   re-establishment wastes its window on peers that no longer exist
   ("Failed to connect to @N Fabric index #N does not exist").

This PR makes the child-bridge Matter server start in a deferred-online mode:
the node is fully built (accessory cache restored, initial plugin
registrations applied to the offline node) and only then brought online — so
its first advertisement carries the final structure and re-establishment runs
against a quiescent node. It also purges persisted subscriptions when their
fabric is removed.

## Changes (validated as dist patches in production)

1. **`ServerConfig` / `ServerLifecycle`: `deferOnline` mode.**
   `start()` builds the node, creates the aggregator, restores the accessory
   cache — and, with `deferOnline`, skips `startServerNode()`. The existing
   `runServer()` path ("deferred server with device(s) already attached",
   previously restricted to external-accessory mode) is opened up to
   `deferOnline` and brings the node online later.
2. **`MatterServer`: registration activity signal.**
   `registerPlatformAccessories()` stamps `lastRegistrationAt`, exposed via
   `getLastRegistrationAt()`.
3. **`ChildBridgeMatterManager`: settle loop.** `deferOnline` is enabled for
   the child-bridge server; after `start()`, `runServerWhenSettled()` polls
   (500 ms) until registrations have been idle for 2 s (capped at 45 s so a
   stuck or registration-less plugin cannot keep the bridge offline — the cap
   logs a warning and goes online anyway), then `runServer()`. The loop stops
   if the server is torn down mid-defer.
4. **`CommissioningManager`: purge persisted subscriptions on fabric
   removal.** The existing `fabricsChanged` handler now, on a `deleted`
   action, filters the removed fabric's entries out of the root endpoint's
   `SubscriptionsServer` state (each persisted entry carries
   `peerAddress.fabricIndex`), so re-establishment only spends its 2-second
   window on live peers. matter.js removes the fabric's sessions on removal
   but leaves the persisted subscription entries behind.
5. **Root node BasicInformation completeness** (same AFU-metadata class of
   fix as the per-accessory changes in the sibling PR #3972):
   `softwareVersion`/`softwareVersionString` parsed from the configured
   firmware revision (`x.y.z` triplet; Apple homed rejects nodes whose
   firmware metadata is missing or unparseable — "invalid matter AFU
   settings"), plus `configurationVersion`, `productUrl`, and a deterministic
   `uniqueId` derived from the bridge's `uniqueId` instead of the random
   matter.js-persisted string controllers would otherwise key node identity
   on.

## Evidence

- Before: every restart of a commissioned bridge logged
  `Reestablished 0 of 2 former subscriptions` / `Operation aborted`,
  followed by minutes of hub dead-session traffic and "No Response" tiles;
  with a stale fabric present, also `Fabric index #N does not exist`.
- After: restarts log `Deferred online mode - Matter node built, waiting for
  initial registrations` → online ~7-10 s later →
  `Reestablished 1 of 1 former subscriptions successfully` within the same
  second — verified repeatedly on an 11-device production bridge via both
  `systemctl restart` and stop/start. Room assignments survive; the hub's
  only trace is a few seconds of stale tiles while secondary peers (phone,
  standby hubs) re-establish their own sessions.
- Subscription persistence itself already works (written at graceful close);
  the failures were entirely the two issues above.

## Notes for reviewers

- Same lineage as #3969/#3970 and the composed-parent FixedLabel/PowerSource
  PR (#3972); validated together on the same deployment, but independent —
  this one is about restart behavior, that one about Apple's composed-device
  modeling.
- Interaction with the beta branch: the registration reject-guard on
  `beta-2.2.2` ("reject bridged registrations that arrive before the server
  is running", part of the #3970 work, not yet on `latest`) must be made
  defer-aware when these meet, or deferred mode deadlocks — the server waits
  for registrations, the guard rejects them for the server not running.
  Offline registration is exactly what deferred mode wants; the cache-restore
  path (#3969) already proves registration works on the offline node. Our
  production validation ran on the beta dist with that guard relaxed to
  treat "deferred, pre-online" as registrable; happy to include that here or
  in a follow-up on the beta branch, whichever the maintainers prefer.
- The 2 s / 45 s settle constants are pragmatic; an explicit "initial
  registrations complete" signal from plugins would be cleaner, but no such
  API contract exists today and the idle heuristic has been reliable.
- Plugins that register accessories long after launch (e.g. from network
  discovery) are unaffected: late registrations apply to the live node
  exactly as today; deferred mode only front-loads whatever arrives during
  the settle window.
- Main-bridge Matter and external accessories are unchanged; the mode is
  enabled for child bridges, where all plugin registrations are in-process.

### DIFF
--- a/src/matter/BaseMatterManager.spec.ts
+++ b/src/matter/BaseMatterManager.spec.ts
@@ -89,6 +89,30 @@ describe('baseMatterManager', () => {
     })
   })
 
+  describe('isBridgeServerStarting', () => {
+    it('is false when no bridge MatterServer exists', () => {
+      expect(manager.isBridgeServerStarting()).toBe(false)
+    })
+
+    it('is false once the server is running', () => {
+      manager.setMatterServer({ isServerRunning: () => true, isDeferredPreOnline: () => false } as any)
+      expect(manager.isBridgeServerStarting()).toBe(false)
+    })
+
+    it('is true while the server is still starting (not running, not deferred)', () => {
+      manager.setMatterServer({ isServerRunning: () => false, isDeferredPreOnline: () => false } as any)
+      expect(manager.isBridgeServerStarting()).toBe(true)
+    })
+
+    it('is false while deliberately offline in deferOnline mode, so registrations are accepted', () => {
+      // The deadlock this guards against: if registrations were rejected while
+      // the deferred node is offline, the settle signal would never fire and
+      // the node would come online with no accessories.
+      manager.setMatterServer({ isServerRunning: () => false, isDeferredPreOnline: () => true } as any)
+      expect(manager.isBridgeServerStarting()).toBe(false)
+    })
+  })
+
   describe('handleTriggerCommand', () => {
     it('should route commands to external server if accessory is external', async () => {
       const uuid = 'test-uuid'
--- a/src/matter/BaseMatterManager.spec.ts
+++ b/src/matter/BaseMatterManager.spec.ts
@@ -111,6 +111,28 @@ describe('baseMatterManager', () => {
       manager.setMatterServer({ isServerRunning: () => false, isDeferredPreOnline: () => true } as any)
       expect(manager.isBridgeServerStarting()).toBe(false)
     })
+
+    it('treats a constructed but not yet built deferred server as starting (real isDeferredPreOnline)', () => {
+      // Exercises the real method, not a stub: between `new MatterServer()`
+      // and start() building the node, deferOnline is already set but no
+      // aggregator exists yet. Registrations in that gap must be rejected as
+      // "starting", or they would hit the silent-drop path from #3970.
+      const server = new MatterServer({ uniqueId: 'test-bridge', deferOnline: true })
+      manager.setMatterServer(server as any)
+      expect(server.isDeferredPreOnline()).toBe(false)
+      expect(manager.isBridgeServerStarting()).toBe(true)
+
+      // Once start() has built the aggregator, the defer window opens...
+      ;(server as any).aggregator = {}
+      expect(server.isDeferredPreOnline()).toBe(true)
+      expect(manager.isBridgeServerStarting()).toBe(false)
+
+      // ...and once the node is online, normal behaviour resumes
+      // (running, so no longer "starting" either).
+      ;(server as any).isRunning = true
+      expect(server.isDeferredPreOnline()).toBe(false)
+      expect(manager.isBridgeServerStarting()).toBe(false)
+    })
   })
 
   describe('handleTriggerCommand', () => {
--- a/src/matter/BaseMatterManager.ts
+++ b/src/matter/BaseMatterManager.ts
@@ -65,6 +65,13 @@ export abstract class BaseMatterManager {
    * are intentionally dropped by the drop stubs — that must not throw.
    */
   isBridgeServerStarting(): boolean {
+    // In deferOnline mode the server is deliberately built but kept offline
+    // until the initial registration burst settles, so registrations MUST be
+    // accepted while it is offline — otherwise they are rejected, the settle
+    // signal never fires, and the node comes online with no accessories.
+    if (this.matterServer?.isDeferredPreOnline()) {
+      return false
+    }
     return this.matterServer !== undefined && !this.matterServer.isServerRunning()
   }
 
--- a/src/matter/ChildBridgeMatterManager.spec.ts
+++ b/src/matter/ChildBridgeMatterManager.spec.ts
@@ -3,7 +3,7 @@ import type { BridgeConfiguration, BridgeOptions } from '../bridgeService.js'
 import type { ChildBridgeExternalPortService } from '../externalPortService.js'
 import type { MatterConfig } from './types.js'
 
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { InternalAPIEvent } from '../api.js'
 import { Logger } from '../logger.js'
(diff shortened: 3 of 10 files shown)


bwp91:
Thank you for this one too. The diagnosis is convincing: bringing the node online before the registration burst has finished, and then letting re-establishment waste its one short window on dead fabrics, explains the "everything goes No Response after a restart" reports very well. The fabric purge in particular is a clean, obviously correct fix.

I checked `SubscriptionsServer` is exported from `@matter/node` in the version we ship — it is, so that import is fine.

Below are the problems, in order of importance.

---

## 1. On the beta branch this does not just conflict — it deadlocks completely (must fix)

You flagged this yourself, and you were right to, but it is worse than "needs to be made defer aware". Here is the exact loop:

- `isBridgeServerStarting()` (in `BaseMatterManager.ts`, from `d3518b3`) returns true whenever the server exists but `isServerRunning()` is false.
- In `deferOnline` mode `isServerRunning()` stays false until `runServer()` is called.
- `runServer()` is only called once registrations have settled.
- But every registration is rejected while the server is not running, so `lastRegistrationAt` is never set.
- `settled` is therefore never true, the loop runs the full 45 seconds, logs the warning, and brings the node online **with no accessories at all**.

On top of that, every plugin's `registerPlatformAccessories()` call throws during that window, so users would see a wall of errors.

**What to change:** the guard needs to allow registrations when the server is deliberately offline in deferred mode. Roughly, `isBridgeServerStarting()` should return false when `deferOnline` is set and the node is built but not yet online. Please include that change in this PR rather than leaving it as a follow up — the two cannot be merged separately without a broken state in between.

---

## 2. A bridge with no Matter registrations waits the full 45 seconds every start (must fix)

In `runServerWhenSettled`:

```ts
const settled = lastRegistrationAt > 0 && Date.now() - idleSince >= DEFER_ONLINE_SETTLE_MS
```

Because of the `lastRegistrationAt > 0` part, a bridge that receives no registrations can never be "settled". It always runs to the 45 second cap, logs a warning, and only then goes online.

This is not a rare case:

- A plugin that finds its devices over the network may register well after startup.
- Restoring from the accessory cache does **not** count as a registration. `restoreAccessoriesFromCache` (`server.ts:371`) has its own path and never touches `lastRegistrationAt`.

So a restart of a bridge whose accessories all come from the cache waits 45 seconds and prints a warning, every single time.

**What to change:** treat "no registrations yet" as settled after the normal 2 second window, rather than waiting for the cap. Counting a cache restore as registration activity would also help.

---

## 3. The registration timestamp is taken too early (must fix)

In `server.ts`:

```ts
async registerPlatformAccessories(...) {
  this.lastRegistrationAt = Date.now()
  for (const accessory of accessories) {
    await this.accessoryManager.registerAccessory(...)
  }
```

The stamp is set when the call *starts*, not when it finishes. If a plugin registers a batch that takes longer than 2 seconds to work through, the settle loop decides things are quiet while registrations are still in progress — and brings the node online in the middle of them. That is the exact situation this PR is meant to avoid.

**What to change:** stamp the time after the loop finishes as well as before it, or keep a count of registrations currently in flight and treat "in flight" as not settled.

---

## 4. `uniqueId` on the root node has the same upgrade risk as #3972

Same point as the other PR, so I will keep it short: your own reasoning is that controllers use `uniqueId` to decide node identity, which means changing it changes the identity of a bridge people have already paired.

**What we need to know:** did you test upgrading an already commissioned bridge, rather than pairing fresh after the change?

---

## 5. The version string loses the pre-release part

Before:

```ts
softwareVersionString: deps.config.firmwareRevision || getVersion()
```

After:

```ts
softwareVersionString: version.join('.')
```

The new code rebuilds the string from the three numbers it parsed, so `2.2.2-beta.5` is shown to controllers as `2.2.2`. That makes beta testers harder to support, because we can no longer see which beta someone is on.

**What to change:** keep the numeric `softwareVersion` fix, which is good, but leave `softwareVersionString` as the original full string.

---

## 6. There are no tests at all

Six files changed and no test file touched. The settle loop especially needs them, since points 1, 2 and 3 above are all logic errors inside it that tests would have caught. Please add coverage for:

- registrations settle, then the node goes online
- no registrations arrive, and the node still goes online promptly
- the server is torn down mid wait, and the loop stops
- a deleted fabric has its persisted subscriptions purged

---

## Smaller points

- **First time pairing during the wait.** While the node is offline it is not advertising, so a user who opens the UI and scans the QR code in that window cannot pair until it comes online. With the 45 second cap in point 2 that could be a long and confusing wait. Worth a note in the log at least.
- **The teardown guard is good.** Checking `this.matterServer !== server` inside the loop is the right way to handle a restart mid wait — no change needed, just noting it.
- **Branch** — this targets `latest`, but the guard from point 1 lives on `beta-2.2.2`, which is where this needs to be tested.

---

Points 1, 2 and 3 are all in the same settle loop, so they should be quick to fix together. The fabric purge part of this PR is ready as it stands. Thanks again for the very careful write up — the evidence made this much easier to review.

keremerkan:
Thanks — all three settle-loop bugs fixed, in one PR as you asked:

- **Deadlock:** `isBridgeServerStarting()` now returns false while the node is deliberately offline in `deferOnline`, so registrations are accepted and the settle signal fires.
- **45s wait:** dropped the `lastRegistrationAt > 0` gate; a bridge with no registrations (or a cache-only restore) settles the window after start and comes online promptly.
- **Early timestamp:** added an in-flight counter and stamp on both entry and exit, so a batch that outlasts the settle window is never treated as idle.
- **Version string:** kept the numeric `softwareVersion` fix, restored the full `softwareVersionString` (pre-release suffix stays visible).
- Dropped the root `uniqueId` for the same reason as #3972.

Tests added for the settle loop, the defer-aware guard, and the fabric purge. Rebased onto `beta-2.2.2`.

keremerkan:
Small housekeeping note while you're here: the merge box on this PR (and every PR since April) shows `build_and_test / build (20)` as "Expected — waiting for status to be reported". That context can never report — `homebridge/.github` dropped Node 20 from the shared build matrix in April (matching the `^22 || ^24` engines), but it's still listed as a required status check in this repo's branch protection. Removing the stale `build (20)` entry from the required checks should clear it repo-wide.

bwp91:
Thanks — all three settle-loop fixes look right, and keeping the full `softwareVersionString` is appreciated.

I checked out the branch and ran it locally:

- `npm run lint` — clean
- full Matter suite — **860/860 passing**

I also tested that the new guard test is meaningful, by removing the defer-aware branch from `isBridgeServerStarting()` and re-running. It failed as it should:

```
× is false while deliberately offline in deferOnline mode, so registrations are accepted
```

Good. But while checking that interaction I found one more problem, and I think it needs fixing before this can go in.

---

## The #3970 guard is now switched off for the whole of startup, not just the defer window

The intent is "accept registrations while the node is deliberately offline". The implementation is slightly wider than that, and the extra bit re-opens the silent-drop bug from #3970.

Here is the chain:

- `deferOnline: true` is set for **every** child bridge (`ChildBridgeMatterManager.ts`, in the `new MatterServer({...})` call).
- `isDeferredPreOnline()` is `this.config.deferOnline === true && !this.isRunning`.
- Both of those are true **as soon as the MatterServer object is constructed** — before anything has been built.
- In `startMatterServer()`, `this.matterServer = new MatterServer({...})` happens on one line, and `await this.matterServer.start()` on a later one.
- In the gap between those two, `aggregator` is still `null`.

So during that gap:

- `isDeferredPreOnline()` returns true,
- `isBridgeServerStarting()` therefore returns false,
- the registration is allowed through,
- it reaches `registerAccessory`, which throws `'Matter server not started'` because there is no aggregator,
- and that error is only **logged** — the plugin's `registerPlatformAccessories()` still resolves as if it worked.

That is exactly the silent drop `d3518b3` was written to prevent.

It is a narrower window than before, and a plugin registering from `didFinishLaunching` will not hit it — but a plugin registering earlier will, which is precisely the case the guard's own error message tells people about. The same applies if `start()` throws: `isRunning` stays false and `aggregator` stays null forever, so every later registration is silently dropped instead of reporting an error.

Your new test does not catch this because it stubs `isDeferredPreOnline()` directly, so the real condition is never exercised.

### The fix is one line

`isDeferredPreOnline()` should mean "deferred **and already built**", not just "deferred". In `server.ts`:

```ts
isDeferredPreOnline(): boolean {
  return this.config.deferOnline === true && !this.isRunning && this.aggregator !== null
}
```

I checked this does not bring the deadlock back. In `ServerLifecycle.ts` the aggregator is set (`deps.setAggregator(aggregator)`) well before the `if (deps.config.deferOnline)` branch that keeps the node offline. So:

- **before `start()` builds anything** — aggregator is null, guard fires, the plugin gets a clear error instead of a silent drop
- **during the real defer window** — aggregator exists, guard stands aside, registrations apply to the offline node (what this PR wants)
- **after `runServer()`** — `isRunning` is true, normal behaviour
- **if `start()` fails** — aggregator stays null, so registrations report an error rather than vanishing

I applied that one line locally and re-ran the full Matter suite: still **860/860 passing**.

Please also add a test that exercises the real `isDeferredPreOnline()` rather than a stub — something like "a constructed but not yet built deferred server is still treated as starting".

---

That is the only thing left from my side. Everything else — the settle loop, the in-flight counter, the fabric purge, the version string — I am happy with. Thanks for turning this around so quickly.

On your CI note: you are right, and thank you for chasing it down. `build (20)` is a stale required status check in this repo's branch protection left over from when the shared workflow dropped Node 20. I will clear it — that needs admin access on the repo, so it is on me rather than something you can fix in a PR.

keremerkan:
Good catch, and you're right on all points — the guard as written opened the defer window at construction, which is exactly the #3970 gap. Applied your one-liner: `isDeferredPreOnline()` now also requires `this.aggregator !== null` (b887e92), with a comment explaining why the aggregator check is load-bearing.

Also added the test you asked for — it exercises the **real** method (no stub) across the lifecycle: constructed-but-not-built → treated as starting; aggregator built but offline → defer window open, registrations accepted; online → normal behaviour. I confirmed it fails if the aggregator condition is removed.

Full suite green (1153), lint clean, and the change is running on my production bridge: registrations still settle and come online normally with the tightened guard, subscriptions re-establish on restart.

Thanks for clearing the `build (20)` check!

bwp91:
Thanks — that is exactly it, and the test is better than what I asked for: walking a real `MatterServer` through constructed → built-but-offline → online pins all three states rather than just the one I flagged. I confirmed here that removing the `aggregator !== null` condition makes it fail, so the #3970 gap cannot quietly reopen.

Verified locally on `b887e92`: lint clean, full suite green at 1153.

I am happy with this one. Merging it into the beta branch now, so it will go out in the next published release.

Thanks as well for the detail you put into the write-ups on this and the sibling PR — the log evidence and the reproducers made both of these genuinely straightforward to review, which is not always the case with this kind of timing bug.

bwp91 (approved):

PR #3976, opened 2026-07-26
feat(matter): populate bridged accessory firmware version from firmwareRevision

keremerkan:
## Summary

HAP-bridged accessories surface their firmware through the AccessoryInformation service's `FirmwareRevision` characteristic. Matter-bridged accessories currently expose nothing: the BridgedDeviceBasicInformation cluster's `SoftwareVersionString`/`SoftwareVersion` attributes are never set, even though every `PlatformAccessory` already carries `firmwareRevision`.

This PR maps it across when registering a Matter accessory:

- `softwareVersionString` ← `accessory.firmwareRevision` whenever it is a non-empty string (the attribute is free-form per spec).
- `softwareVersion` (numeric) ← additionally encoded as `major<<16 | minor<<8 | patch` when the string starts with a `x.y.z` semver core, so version comparisons work for controllers that use the numeric attribute. Non-semver strings set only the string attribute.

## Validation

- Verified on the wire (controller read from a second fabric): each bridged endpoint reports its own device's version — e.g. `swString=1.7.5 sw=67333` on most devices and a distinct `2.0.0-g87fbfa4` (string only) on another — running as a dist patch on a production bridge with 12 bridged accessories.
- One honest caveat: Apple Home currently renders per-accessory firmware for HAP-bridged accessories but not (yet) for Matter-bridged ones, even with these attributes present — we tested both raw vendor strings and clean numeric strings across hub reboots. The mapping is spec-correct and consumable by other controllers/ecosystems (and by Apple whenever they wire it up); we're filing the rendering gap with Apple separately.
- 3 new unit tests (semver → string+numeric, non-semver → string only, absent → neither); full suite passes.

### DIFF
--- a/src/matter/server/AccessoryManager.spec.ts
+++ b/src/matter/server/AccessoryManager.spec.ts
@@ -713,6 +713,53 @@ describe('accessoryManager', () => {
       expect(bdbi.softwareVersion).toBeUndefined()
     })
 
+    it('encodes a large major that still fits 16 bits as an unsigned value', async () => {
+      const deps = createMockDeps()
+      const accessory = createMockAccessory({ firmwareRevision: '40000.0.0' } as any)
+
+      await manager.registerAccessory('homebridge-test', 'TestPlatform', accessory, deps)
+
+      const bdbi = (deps.accessories.get('test-uuid-001') as any).endpoint.options.bridgedDeviceBasicInformation
+      expect(bdbi.softwareVersionString).toBe('40000.0.0')
+      // 40000 sets the sign bit under a plain 32-bit shift; the value must stay unsigned
+      expect(bdbi.softwareVersion).toBe(40000 * 0x10000)
+    })
+
+    it('omits the numeric version when the major does not fit 16 bits', async () => {
+      const deps = createMockDeps()
+      const accessory = createMockAccessory({ firmwareRevision: '70000.0.0' } as any)
+
+      await manager.registerAccessory('homebridge-test', 'TestPlatform', accessory, deps)
+
+      const bdbi = (deps.accessories.get('test-uuid-001') as any).endpoint.options.bridgedDeviceBasicInformation
+      expect(bdbi.softwareVersionString).toBe('70000.0.0')
+      expect(bdbi.softwareVersion).toBeUndefined()
+    })
+
+    it('omits the numeric version when minor or patch exceeds 255', async () => {
+      const deps = createMockDeps()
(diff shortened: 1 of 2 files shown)


bwp91:
Thanks @keremerkan — nice feature, and the validation write-up is appreciated (including the honest note about Apple Home not rendering it yet — the attributes are still worth having for other controllers, and for Apple whenever they wire it up).

Checked on our side:

- Both attributes are independently optional in the BridgedDeviceBasicInformation cluster (conformance "O" each in the spec model), so the string-only path for non-semver versions is fine.
- The tests cover the three paths well.

One edge case we'd like guarded before merge:

- The numeric derivation uses `<<`, which operates on signed 32-bit values. A version string with a large first number (e.g. `40000.0.0`) makes `major << 16` go negative, and a negative value fails matter.js's uint32 validation — which could stop that accessory registering at all. That's a heavy failure mode for a cosmetic attribute.
- A third number over 255 (e.g. `1.7.300`) also spills into the wrong byte — harmless, but the ordering semantics break.

Suggested fix: only set the numeric version when the parts fit their fields (`major <= 65535`, `minor <= 255`, `patch <= 255`), and otherwise fall back to the string-only path you already have. Something like:

```ts
const [major, minor, patch] = [Number(semver[1]), Number(semver[2]), Number(semver[3])]
if (major <= 0xFFFF && minor <= 0xFF && patch <= 0xFF) {
  endpointOptions.bridgedDeviceBasicInformation.softwareVersion = (major << 16) | (minor << 8) | patch
}
```

Optional nice-to-have: the spec caps `SoftwareVersionString` at 64 characters, so truncating the string to 64 would guard against an unusually long `firmwareRevision` failing validation too.

With the guard in (ideally plus a test for the out-of-range case), this is good to merge.

keremerkan:
Guard and truncation added in 5d998b0, with tests for the out-of-range and truncation cases — thanks for catching this.

One adjustment to the suggested snippet: `major <= 0xFFFF` alone still produces a negative value for majors ≥ 32768 (`40000 << 16` sets bit 31 under JS's signed 32-bit shift, even though 40000 fits 16 bits). Since those majors are legitimately encodable in a uint32, I kept your field-fit bounds and added `>>> 0` to keep the encoded value unsigned — covered by a dedicated test (`40000.0.0` → `2621440000`, `70000.0.0` → string only).

bwp91 (approved):
All looks good now yes, good catch on the `>>> 0`!

This will be included in the next beta which will publish just after this merges into the beta branch.

Thanks!

PR #3977, opened 2026-07-26
fix(matter): declare ac/dc feature on power measurement

keremerkan:
## Summary

The ElectricalPowerMeasurement cluster's feature conformance requires at least one of the current-type features (DirectCurrent/AlternatingCurrent). The auto-detected power measurement server in `applyElectricalMeasurementClusters()` composes the cluster with no features at all — a controller reading the featureMap sees `0`, which is non-conformant. Certified mains-measuring devices declare AlternatingCurrent (verified with a controller read against an Eve Energy).

The fix follows the accessory's `powerMode`, which `applyElectricalMeasurementDefaults()` has already resolved by the time the composition runs (plugin-declared value, AC default): DC (`powerMode: 1`) declares DirectCurrent, everything else declares AlternatingCurrent — mirroring how PowerTopology already gets TreeTopology in the same function.

## Validation

- The AlternatingCurrent path is running as a dist patch on a production bridge (12 bridged accessories, 7 metering): clean startup, no re-registration churn, Apple Home controls/tile wattage/energy totals unchanged.
- Tests: featureMap assertion added to the existing AC-default case, plus a new DC case asserting DirectCurrent/not-AlternatingCurrent; full suite passes.

### DIFF
--- a/src/matter/serverHelpers.spec.ts
+++ b/src/matter/serverHelpers.spec.ts
@@ -856,10 +856,24 @@ describe('serverHelpers', () => {
       expect(result.behaviors.powerTopology).toBeDefined()
       expect(result.behaviors.powerTopology.features.treeTopology).toBe(true)
       expect(result.behaviors.electricalPowerMeasurement).toBeDefined()
+      expect(result.behaviors.electricalPowerMeasurement.features.alternatingCurrent).toBe(true)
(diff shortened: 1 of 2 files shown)


bwp91 (approved):
Thanks @keremerkan — this is spot on.

Happy to merge this as-is. It will go out in the next 2.2.2 beta.

PR #3983, opened 2026-08-03
offer one shot windows desktop deployment

NannaOlympicBroadcast:

_Describe the current situation. Explain current problems, if there are any. Be as descriptive as possible (e.g., including examples or code snippets)._

## :bulb: Proposed solution

_Describe the proposed solution and changes. How does it affect the project? How does it affect the internal structure (e.g., refactorings)?_

## :gear: Release Notes

_Provide a summary of the changes or features from a user's point of view. If there are breaking changes, provide migration guides using code examples of the affected features._

_If applicable, provide additional context in this section._

### Testing

_Which tests were added? Which existing tests were adapted/changed? Which situations are covered, and what edge cases are missing?_

### Reviewer Nudging

_Where should the reviewer start? what is a good entry point?_


NorthernMan54:
A couple of comments

 1 - What's the benefit of this approach versus the existing VM Image approach for Windows ? https://github.com/homebridge/homebridge/wiki/Install-Homebridge-on-Virtual-Machine

2 - If we were going to consider this, it would need to be a wrapper package and not part of the homebridge itself.  Similar to the other run time environments available.

NannaOlympicBroadcast:
this service may be hard to configure running in a vm or containerlized environment struggling with local service discovery and if it runs locally in windows it will be easier for that and that is why i make this contribution

NorthernMan54:
Have you tried the VM installation in Hyper-V, the container setup is already done for you.  Just download and install

NannaOlympicBroadcast:
I think vm installation is complicated and one shot installer is fit for everyone

Issue #3980, opened 2026-07-27
Provide a supported way for plugins to reconcile restored cached accessories


bwp91:
Splitting this out of #3919, where @n0rt0nthec4t raised it and it risks getting lost inside a narrower question.

## The gap

When Homebridge restores a cached accessory, the plugin gets back whatever shape was last published. If the plugin's expectations have since changed — a service is no longer wanted, a characteristic was dropped, a context field was added, or HAP itself changed between versions — the plugin is on its own in noticing and repairing that.

The tools to fix it exist (`removeService`, `updatePlatformAccessories`), so this is not about missing capability. It is that every plugin author ends up writing their own reconciliation pass, and each one discovers the same traps independently:

- removing every service breaks, because `AccessoryInformation` is required
- doing the work too early, during cache load, crashes deserialisation
- the fallback that always works — delete the accessory and let it be recreated — costs the user their room assignment, name, and automations

In practice a lot of plugins skip it, and the symptom lands on users as a stale service that never goes away.

## What might help

Roughly in order of how much they'd cost us:

1. **Document the pattern.** The safe reconciliation recipe, in the plugin developer docs: remove the specific services you no longer want in your accessory handler, never `AccessoryInformation`, then call `updatePlatformAccessories`. This alone would cover most cases.
2. **A helper on `PlatformAccessory`** — something like "keep exactly these services and drop the rest", so the common case is one call instead of a hand-rolled loop with a footgun in it.
3. **A validation hook** before registration, so a plugin can inspect a restored accessory and say whether it is still usable, with a defined recovery path when it is not.
4. **Cache sanity checking in core**, for the cases where the stored data is incompatible with the current HAP version rather than with the plugin.

Opening this for discussion rather than proposing we build all of it. My own view is that 1 is worth doing regardless and 2 is a small, contained addition; 3 and 4 are a much bigger commitment and want a clearer picture of how often incompatible caches actually occur.

ebaauw:
[...]

Issue #3919, opened 2026-04-20
Cached accessories not updated when accessory is modified


justjam2013:
### Analysis

The cached accessories are not getting updated when the accessory is modified.

Consider the case where two switches are created. `Switch Without Sensor` is a simple Switch. `Switch With Sensor` creates a Switch along with a companion Contact Sensor. The following configuration snippet shows the two accessories:
```
                {
                    "accessoryID": "4260746",
                    "accessoryName": "Switch Without Sensor",
                    "accessoryType": "switch",
                    "accessoryIsStateful": false,
                    "switch": {
                        "defaultState": "off",
                        "hasResetTimer": false,
                        "hasCompanionSensor": false,
                        "muteLogging": false
                    }
                },
                {
                    "accessoryID": "7929993",
                    "accessoryName": "Switch With Sensor",
                    "accessoryType": "switch",
                    "accessoryIsStateful": false,
                    "switch": {
                        "defaultState": "off",
                        "hasResetTimer": false,
                        "hasCompanionSensor": true,
                        "muteLogging": false
                    },
                    "companionSensor": {
                        "name": "Switch With Sensor Companion Sensor",
                        "type": "contact"
                    }
                }

```

After creating the two accessories above and restarting Homebridge, they are shown in the accessories screen. Per the configuration the two Switches are named `Switch Without Sensor` and `Switch With Sensor`, and the sensor is named `Switch With Sensor Companion Sensor`:

<img width="2702" height="748" alt="Image" src="https://github.com/user-attachments/assets/5a506e03-c4f5-4650-b91f-4a64cae36173" /><p>

Now the two accessories are modified by adding a companion Contact Sensor to the Switch `Switch Without Sensor` and removing the companion Contact Sensor from the Switch `Switch With Sensor`. The following updated configuration snippet shows the two accessories:
```
                {
                    "accessoryID": "4260746",
                    "accessoryName": "Switch Without Sensor",
                    "accessoryType": "switch",
                    "accessoryIsStateful": false,
                    "switch": {
                        "defaultState": "off",
                        "hasResetTimer": false,
                        "hasCompanionSensor": true,
                        "muteLogging": false
                    },
                    "companionSensor": {
                        "name": "Switch Without Sensor Companion Sensor",
                        "type": "contact"
                    }
                },
                {
                    "accessoryID": "7929993",
                    "accessoryName": "Switch With Sensor",
                    "accessoryType": "switch",
                    "accessoryIsStateful": false,
[...]

PR #3986, opened 2026-08-09
feat: add a service reconciliation helper

danieloleary:
The original report exposed a small but nasty gap: cached accessories remember the services a plugin used to publish, but Homebridge has no simple way for the plugin to say, “these are the services that still belong here.” The necessary APIs exist, but every plugin ends up rebuilding the same cleanup loop and learning the same lifecycle traps.

This adds `PlatformAccessory.reconcileServices()` for that common case. A plugin finishes discovery, attaches the services it wants to keep, configures any controllers, and then reconciles the restored accessory before persisting it:

```ts
accessory.reconcileServices([switchService, contactSensorService])
api.updatePlatformAccessories([accessory])
```

The helper keeps HomeKit's information services, rejects service objects that do not belong to the accessory, and treats services with the same UUID but different subtypes independently. It deliberately stops at services; characteristics, plugin context, and whole-accessory cleanup have different ownership and safety rules.

I tested the boring-but-important path end to end: restore from cache, reconcile through the real `updatePlatformAccessories` event, save, restart, and confirm the same accessory identity comes back without the stale service. The focused regressions cover controllers, subtypes, repeated runs, and validation before anything is removed.

Thanks to @justjam2013 for the concrete reproduction, and to @ebaauw and @n0rt0nthec4t for sharing the patterns they already use in their plugin libraries. That made the right lifecycle much easier to pin down.

Addresses #3980.


n0rt0nthec4t:
Thanks for splitting this out.

I think PlatformAccessory.reconcileServices() would make the existing plugin-side workaround easier, but I am not sure it addresses the underlying issue I was trying to raise in #3919.

The problem for me is that this still requires every plugin developer to know that reconciliation is required, determine the correct point in the lifecycle to perform it, construct the expected service set, call reconcileServices(), and then persist the accessory again.

So although the helper removes some boilerplate, the responsibility and most of the lifecycle complexity remains in every plugin.

My preference would be for Homebridge to handle this internally as part of restoring/updating cached accessories.

Conceptually, the plugin should describe the accessory as it exists now, and Homebridge should reconcile the cached representation against that current definition. A service that was previously cached but is no longer published by the plugin should not remain indefinitely simply because the plugin did not explicitly remove it.

That also gives us one implementation of the reconciliation rules rather than potentially hundreds of plugins each implementing, or forgetting to implement, some version of the same lifecycle.

I appreciate there are ownership questions around characteristics, context and whole-accessory replacement, and those may need to remain plugin controlled. But services seem different: Homebridge already owns persistence of the PlatformAccessory, so I think stale persisted services are something core should ideally be able to resolve without requiring an explicit cleanup API call from every plugin.

Otherwise I worry reconcileServices() becomes another API plugin authors need to know they should call, while existing plugins , which are where users are most likely to encounter stale cached state  continue to exhibit exactly the same problem.

justjam2013:
@bwp91 you closed this PR with no comment on your end as to why.

bwp91:
sorry i didn't mean to. it happened automatically when i deleted the existing beta branch

bwp91:
Hi @danieloleary i'm sorry this keeps getting closed. Are you please able to create a new PR that points towards the `latest` branch?

danieloleary:
Thanks Ben, no problem. I recreated it cleanly against latest here: #3990. It is the same three-file change, and I reran the tests on the current base. Thanks for pointing me in the right direction.

PR #3990, opened 2026-08-11
feat: add a service reconciliation helper

danieloleary:
This replaces #3986 on `latest`, as requested after the old beta branch was removed.

Cached accessories can retain services that a plugin no longer publishes. Homebridge already exposes the pieces needed to clean them up, but each plugin has to rebuild the same lifecycle-sensitive loop.

This adds `PlatformAccessory.reconcileServices()` for that explicit cleanup step. It preserves HomeKit information services, handles service subtypes independently, validates every desired service before removal, and leaves persistence and whole-accessory replacement under plugin control.

If an omitted service belongs to either a configured controller or controller data restored from cache, reconciliation throws before changing the accessory. That turns a potentially persistent controller-cache failure into an immediate, recoverable plugin error.

```ts
const desiredServices = accessory.services.filter(service => !staleServices.includes(service))
accessory.reconcileServices(desiredServices)
api.updatePlatformAccessories([accessory])
```

The regressions cover ordinary cleanup, validation, subtypes, repeated calls, configured and restored controllers, atomic failure, persistence, and restart. All 1,306 tests pass on Node 22, 24, and 26, along with lint, docs lint, build, and an isolated Homebridge startup.

Thanks to @justjam2013 for the concrete reproduction; @ebaauw and @n0rt0nthec4t for sharing the patterns they already use; and @bwp91 for finding the controller-cache failure before this shipped. Addresses #3980.


bwp91 (changes requested):
Thanks for moving this over to `latest`, and for the test coverage — it is more thorough than most PRs I get.

I ran it locally on Node 24: lint, build and all 1,250 tests pass, so those claims hold up. I also checked a few things the tests do not cover, and one of them is a blocker.

## Blocking: controller services are not protected

`configureController()` adds services that the plugin never created and usually holds no reference to. If a plugin passes only its own services, those controller services get removed, but the controller's internal service map still points at them. That map is written to the cache, and on the next start it comes back empty, so the controller throws.

Minimal repro, which fails on this branch:

```ts
// run 1
const accessory = new PlatformAccessory('Doorbell', uuid.generate('demo'))
accessory.configureController(new RemoteController())
const motion = accessory.addService(Service.MotionSensor)

accessory.reconcileServices([motion])   // the plugin's own services

const cache = PlatformAccessory.serialize(accessory)

// run 2 — restored from cache, plugin re-configures its controller
const restored = PlatformAccessory.deserialize(cache)
restored.configureController(new RemoteController())
// Error: Unexpected state: Services not configured!
```

Why this is worse than a normal footgun:

- It is silent in the session where it happens. Nothing warns.
- The damage is written to `cachedAccessories`, so it survives a restart.
- Every start after that throws in `configureController`, so the plugin cannot start at all.
- The only recovery is deleting the accessory, which costs the user their room, name and automations. That is the exact outcome #3980 set out to avoid.

Cameras are the common case. A camera plugin calls `configureController(new CameraController(...))` and then reconciles against its own motion and doorbell services, which is the natural reading of the docs as written here.

The doc comment does say to pass controller services in, but this is an API where leaving one argument out breaks the accessory for good, and controller services are the ones an author is least likely to have to hand. It needs a guard rather than a note.

Two ways I would consider:

1. **Keep controller-owned services automatically.** Cleanest for plugin authors. The service maps live on `Accessory.controllers` and `Accessory.serializedControllers`, both private, so it needs a small addition in HAP-NodeJS first. I am happy to add that.
2. **Throw if a stale service belongs to a controller.** Keeps everything inside this PR, keeps the API explicit, and turns a permanent break into an obvious error the author sees straight away.

Either way it has to cover both maps: `controllers` for a controller that has been configured, and `serializedControllers` for an accessory restored from cache whose controller has not been re-configured yet.

## Smaller things

- `RangeError` is not quite right for "this service is not attached to this accessory". `TypeError` is the usual choice for a wrong-kind-of-value argument. Worth fixing while it is still unreleased, since the type is documented.
- Please add a CHANGELOG entry. One bullet, matching the commit subject: `feat: add a service reconciliation helper (#3990) (@danieloleary)`.
- Very minor: the validation loop calls `this.services.includes(service)` once per desired service. Building a `Set` of the attached services first makes it linear. It only shows up on something like a Television with a lot of `InputSource` services, so take it or leave it.

## Things I checked that are fine

- Validation runs all the way through before anything is removed, so a bad call leaves the accessory untouched.
- `filter` snapshots the list before removal, so there is no mutation-while-iterating skip.
- HAP unlinks removed services from the ones you keep, so linked services are handled for you.
- `PlatformAccessory.services` stays the same array object that HAP splices, so the mirror does not drift.
- Guarding `ProtocolInformation` is not dead code — external accessories do get one, because it is added in `publish()`.

## Still open

@n0rt0nthec4t's point above is the one I have not answered yet. This is option 2 from #3980, and it does leave the work with the plugin author. I still think that is the right first step, because core cannot know which services a plugin has stopped wanting, but I want to be clear it is a first step and not the whole fix.

danieloleary:
Thanks @bwp91 for the careful review and the controller repro, and @n0rt0nthec4t for raising the broader concern. You were right about the controller-service risk.

I have pushed a guard that checks both configured controllers and controller data restored from cache. If reconciliation would remove a controller-owned service, it now throws before changing anything. I also switched the detached-service error to `TypeError`, used a `Set` for attached-service validation, and added the requested changelog entry.

The new regressions cover configured controllers, restored-but-not-configured controllers, atomic failure, restart safety, and the no-op path. The current `latest` branch passes all 1,306 tests on Node 22, 24, and 26, along with lint, docs lint, and build.

Thanks again for digging into this failure mode.

src/platformAccessory.ts
```diff
@@ -97,6 +96,64 @@ export class PlatformAccessory<T extends UnknownContext = UnknownContext> extend
...
+    )
+    if (staleServices.length === 0) {
+      return []
+    }
+
+    const serializedAccessory = Accessory.serialize(this._associatedHAPAccessory)
+    const controllerServiceIds = new Set(serializedAccessory.controllers
+      ?.flatMap(controller => Object.values(controller.services)) ?? [])
+    if (staleServices.some(service => controllerServiceIds.has(service.getServiceId()))) {
+      throw new TypeError('Cannot reconcile a service managed by an accessory controller')
```
bwp91: Could this name the service it is refusing to remove — its UUID and subtype? As it stands a plugin author gets no identifier to search for, which on an accessory with a camera controller and a dozen services means hunting for which one tripped it.

Same thought applies to the not-attached throw above.

src/platformAccessory.ts
```diff
@@ -97,6 +96,64 @@ export class PlatformAccessory<T extends UnknownContext = UnknownContext> extend
     this._associatedHAPAccessory.removeService(service)
   }
 
+  /**
+   * Removes services omitted from a platform's current discovery result.
+   *
+   * Call this after discovery and controller configuration, never during deserialization. Pass every
```
bwp91: One thing worth adding to the example. This tells the author to pass every controller service that should remain, but the controller service maps are private, so with a camera or remote attached they cannot actually enumerate them.

The way that does work is to express it as a subtraction rather than an inclusion:

```ts
accessory.reconcileServices(accessory.services.filter(service => !staleServices.includes(service)))
```

I tried that with a controller configured and it behaves as you would want. Since it is the only practical route for anyone using a controller, it might be worth being the example rather than a footnote.

bwp91 (commented):
This looks good now — thanks for taking the controller point seriously.

Going through `Accessory.serialize()` to read the controller service maps is a nicer answer than the one I suggested: it picks up both the configured controllers and the ones restored from cache but not yet re-configured, without needing anything added to HAP-NodeJS. I checked it against the sequence from my last review and it now throws before touching the accessory, and the cache round-trip that used to fail on the next restart survives.

I also wondered whether serializing the whole accessory just to read those ids would be costly, so I measured it rather than guess: at HAP's cap of 100 services (99 services, 492 characteristics) it is about 0.7ms, and the early return means it does not run at all when nothing is stale. Not a concern.

1,306 tests, lint and build all pass for me locally too.

Two small things below, neither of them blocking.

n0rt0nthec4t:
I still think this is a bandaid over the underlying issue in #3980.

I’m being vocal about this because I think we should take a more holistic view of the problem and fix it at the core, rather than adding another API that plugin authors need to understand, call at the right time, and account for alongside controllers and other service ownership.

The fact that reconcileServices() itself now needs additional safeguards around controller owned services is a good example of the complexity we’re pushing out to deal with a problem that fundamentally originates in cached accessory state and its lifecycle.

Homebridge has the knowledge of the cached state, service ownership and lifecycle. Ideally, plugin authors shouldn’t each need to invoke another reconciliation mechanism to keep that state valid or recover from historical state that Homebridge has restored.

So while this PR makes the proposed bandaid safer, I don’t think it changes the fundamental concern. I’d much rather we step back, take the holistic view, and fix the underlying lifecycle/cache issue in core rather than introduce another API and another responsibility for plugin authors to worry about.

justjam2013:
I agree with @n0rt0nthec4t. By adding `reconcileServices()` all this PR does is expose the internal workings of Homebridge, like your friendly neighborhood park flasher. That's a bad idea.

**Consider this scenario:** Let's assume that the caching mechanism is removed some day, maybe due to underlying changes to HAP or whatever. The `reconcileServices()` API call will need to remain in place because of all the plugins using it. You could update the documentation, but if any authors don't update their plugins, you would effectively kneecap a bunch of plugins that users rely on.
However, because the cache no longer exists, `reconcileServices()` will effectively become a `noop`. Newer plugin developers, looking at existing plugins as examples, will parrot the dogma and add a useless call to `reconcileServices()` to their own new plugins, further calcifying the API call in place.

So Homebridge will end up stuck with a zombie API call that can never be removed and that is perpetuating its own use.

**Edit:** Plus if the plugin is the golden master, the source of truth, then the cache needs to update itself based on the device published by the plugin. The cache should not behave like a prima donna, like it is the source of truth.

n0rt0nthec4t:
further  update:

In my own plugin I’ve had to work around this with cameras. For example, when changing from an HKSV configuration to one without HKSV, I’ve had to explicitly find and remove a stale CameraOperatingMode service because changing the controller configuration doesn’t necessarily clean up all of the services that the previous controller configuration created.

That’s controller owned state, yet the plugin ends up needing to understand enough about the controller’s internals to identify and clean it up itself.

We also already have updatePlatformAccessories(), but despite the name its role is largely to tell Homebridge that the platform accessory has been changed so that state can be persisted. It doesn’t provide the validation/reconciliation lifecycle that I think is actually missing here. The plugin still has to determine what the correct state should be, manually repair it, and then ask Homebridge to persist the result.

Adding reconcileServices() gives us another API alongside that:

```
accessory.reconcileServices(...);
api.updatePlatformAccessories(...);
```

but we’re still leaving the difficult part with every plugin author: understanding stale cached state, service ownership, controllers, and what is or isn’t safe to remove.

I think the cleaner ownership model is that plugins manage the services they own, controllers manage and reconcile the services they own as their configuration changes, and Homebridge/HAP validates and migrates the persisted accessory state it owns.

src/platformAccessory.ts
danieloleary: Good call. Both TypeError paths now include the service UUID and subtype, and the tests pin those diagnostics.

src/platformAccessory.ts
danieloleary: Agreed. I changed the example here and in the PR description to filter known stale services out of accessory.services, so private controller services remain included.

bwp91:
I have asked Fable to try to find a path forward, where everybody is happy. Here goes...!

---

## PR #3990 (this PR) as phase 1 of a bigger plan

### The problem in one paragraph

When Homebridge restarts, it rebuilds each accessory from a saved copy (the cache). If the plugin has changed since — say it no longer offers a fan service — the old fan service comes back anyway and never goes away. Today, every plugin author has to write their own clean-up code, and they all trip over the same traps.

### Who wants what

- **@danieloleary (PR author)**: a safe, ready-made way for a plugin to remove its own leftover services. Wants it now.
- **@n0rt0nthec4t**: a clear split of responsibility — plugins look after their services, controllers (cameras, remotes, TVs) look after theirs, and Homebridge looks after the saved state itself. Worried this PR pushes the hard work onto plugin authors instead.
- **@justjam2013**: worried we add a public method we can never remove, and that the cache acts like the boss when the plugin should be.

The phases below give the author their tool immediately, and end at exactly the split @n0rt0nthec4t describes. @justjam2013's worry is answered by naming the tool narrowly and keeping it useful even in the end state.

---

### Phase 1 — a safe tool for plugin-owned leftovers (this PR)

A plugin can say "these services of mine are stale, remove them", and Homebridge does it with guard rails: it never removes the required information services, and it never touches services owned by a controller.

**Example:** a plugin stops supporting a humidity sensor. After a restart the old sensor service is still there. The plugin calls the new method with that service, and it is gone — safely, with the accessory keeping its room, name and automations.

#### Changes needed to this PR first

1. **Take the stale list, not the keep list.** Rename to something like `removeStaleServices(staleServices)`. Callers already know what is stale; asking them for "everything except the stale ones" invites mistakes. Bonus: passing an empty list now does nothing, instead of wiping the whole accessory.
2. **Don't refuse everything over one controller service.** Today, if even one passed service belongs to a controller, nothing at all is removed. Instead: remove what is allowed, and hand back a list of what was refused and why.
3. **Two kinds of "no".** Throw an error when the caller made a mistake (service not on this accessory, or one of the protected information services). Quietly report — not throw — when the service belongs to a controller, because that is a gap Homebridge itself will fix in phase 3.
4. **Honest docs.** Say plainly: this removes services *the plugin owns*. Controller clean-up is a separate, tracked piece of work (#3980).

#### Does n0rt0nthec4t still get what they want with these changes?

Yes — these changes help, they don't hurt. The refusal to touch controller services *is* their ownership split, written into the code from day one. And the "refused" list from change 2 becomes a running record of real-world cases where controllers leave mess behind — exactly the evidence needed to justify phase 3.

---

### Phase 2 — write down the safe pattern

Document the recipe in the plugin developer docs: when to clean up, what never to remove, and how to save the result. Cheap, uncontroversial, and already agreed as worth doing in #3980. Can happen alongside phase 1.

---

### Phase 3 — controllers clean up after themselves

The real gap n0rt0nthec4t hit: turn a camera feature (like HKSV) off, and the controller leaves one of its old services behind. The plugin has to poke around inside controller internals to fix it — which is backwards.

In this phase, HAP's controllers remove their own leftover services whenever their configuration changes. Plugins stop needing to know anything about controller internals. The phase-1 method's "refused" list should, over time, come back empty.

---

### Phase 4 — Homebridge checks its own saved state

The last layer: when restoring from cache, Homebridge itself notices saved state that no longer makes sense (for example, after a HAP upgrade) and repairs or migrates it. Plugins are only ever asked about things only they can know — like whether a device still exists.

---

### The end state

- Plugins manage their services — with the phase-1 tool.
- Controllers manage theirs — automatically, from phase 3.
- Homebridge manages the saved state — from phase 4.

The phase-1 method does **not** become a useless leftover (justjam2013's worry): even in the end state, only the plugin can know that a device or feature is gone, so a plugin-facing call stays necessary. Naming it narrowly now (`removeStaleServices`, not `reconcileServices`) keeps the bigger name free for the bigger job, and keeps every promise the method makes true forever.

The main risk is not technical: it is that phase 1 relieves the pain and phases 3–4 quietly never happen. The fix is to keep #3980 open with the later phases written down, and say so on the PR when it merges.

bwp91:
## PR #3990 — the exact changes for phase 1

Companion to `pr-3990-phases.md`. That doc explains the plan; this one lists the precise edits to the PR. All line numbers refer to the PR's current head (`600b5ad`).

The PR touches four files: `src/platformAccessory.ts` (the method), `src/platformAccessory.spec.ts` (its tests), `src/bridgeService.spec.ts`, and `CHANGELOG.md`. Every change below lands in those same four files — nothing new is added.

---

### Change 1 — take the stale list, not the keep list

**Where:** `src/platformAccessory.ts`, line 126.

**Now:**

```ts
public reconcileServices(desiredServices: readonly Service[]): Service[] {
```

**Becomes:**

```ts
public removeStaleServices(staleServices: readonly Service[]): { removed: Service[], refused: Service[] } {
```

**Why, with an example:** today a plugin that knows "the fan service is stale" must first build the opposite list — every service *except* the fan — just to satisfy the method. After the change it passes the fan service directly. And if a bug hands the method an empty list, nothing happens, instead of every service being removed.

The whole first block (lines 127–147: the `desired` set, the attached-check loop, and the `staleServices` filter) simplifies, because the caller now hands us the stale list directly:

```ts
const attached = new Set(this.services)
for (const service of staleServices) {
  if (!attached.has(service)) {
    throw new TypeError(
      `Cannot remove service ${service.UUID} (subtype: ${service.subtype ?? 'none'}): service is not attached to this accessory`,
    )
  }
}
```

The line-145 early return (`if (staleServices.length === 0) return []`) keeps its job: an empty list skips the serialize step and returns `{ removed: [], refused: [] }`.

---

### Change 2 — protect the information services with an error, not silence

**Where:** the filter at lines 140–144 currently drops `AccessoryInformation` and `ProtocolInformation` silently.

With a stale list, a plugin that passes one of these has made a mistake, and silence would hide it. Add to the loop from change 1:

```ts
if (service.UUID === Service.AccessoryInformation.UUID || service.UUID === Service.ProtocolInformation.UUID) {
  throw new TypeError(
    `Cannot remove service ${service.UUID}: this service is required and managed by Homebridge`,
  )
}
```

---

### Change 3 — remove what is allowed, report what is not

**Where:** lines 148–156. This is the heart of it.

**Now** (one controller-owned service blocks everything):

```ts
const controllerService = staleServices.find(service => controllerServiceIds.has(service.getServiceId()))
if (controllerService) {
  throw new TypeError(
    `Cannot reconcile service ${controllerService.UUID} (subtype: ${controllerService.subtype ?? 'none'}): service is managed by an accessory controller`,
  )
}

for (const service of staleServices) {
  this.removeService(service)
}
return staleServices
```

**Becomes** (split into refused and removed):

```ts
const refused = staleServices.filter(service => controllerServiceIds.has(service.getServiceId()))
const removed = staleServices.filter(service => !controllerServiceIds.has(service.getServiceId()))

for (const service of removed) {
  this.removeService(service)
}
return { removed, refused }
```

The controller detection itself (lines 148–150, using `Accessory.serialize`) is kept exactly as the author wrote it — it works, including for controllers restored from cache but not yet configured.

**Why, with an example:** a camera plugin has two leftovers — its own old motion sensor, and a controller-owned recording service. Today the method throws and removes neither. After the change the motion sensor is removed and the recording service comes back in `refused`, so the plugin can log it or file an issue. The `refused` list is the running evidence for phase 3.

---

### Change 4 — rewrite the doc comment

**Where:** lines 99–125.

Drop the sentences about building a desired list and about controller services being impossible to enumerate (lines 102–104) — that problem no longer exists. The example (lines 115–119) loses its filter line:

```ts
const { refused } = accessory.removeStaleServices(staleServices)
api.updatePlatformAccessories([accessory])
```

Add two plain statements: this removes services *the plugin owns*; controller-owned services are returned in `refused` rather than removed, and that gap is tracked in homebridge#3980.

---

### Change 5 — the tests

**Where:** `src/platformAccessory.spec.ts`, the `describe` at line 91 (13 tests) and the three reconcile tests in `src/bridgeService.spec.ts`.

Most adapt mechanically — pass the stale list instead of its complement, expect `{ removed, refused }` instead of an array. Three change meaning:

- Line 181 `rejects desired services that are not attached` — same test, now the stale service is the unattached one.
- Line 215 `rejects removing services owned by a configured controller` — now expects the service in `refused` with nothing thrown.
- Line 235 `does not remove ordinary stale services when a controller service blocks reconciliation` — inverted: the ordinary service **is** removed, and only the controller one is refused. This test is the phase-1 behaviour in one line.

Two new tests: an empty list removes nothing and touches nothing (extends line 114's no-serialize test), and passing `AccessoryInformation` throws.

---

### Change 6 — changelog and PR description

The changelog line becomes its commit subject as usual, e.g. `feat: add removeStaleServices so plugins can drop their stale services safely (#3990) (@danieloleary)`. The PR description swaps its example for the new two-liner and states the scope: plugin-owned services now, controller-owned tracked in #3980.

---

### What is deliberately not changed

- The controller detection via `Accessory.serialize` — the author's approach, kept as is.
- The instance-comparison rule (same UUID, different subtype stay independent) — carries over untouched.
- `removeService` (line 95) and everything else in the file.

justjam2013:
To me Phase 1 is completely useless.

> Take the stale list, not the keep list.

That means that the plugin must hold a hold a history of the previous configuration, so that it can diff it against the current configuration and then determine which services, if any have been removed and are now stale in the cache.

This step is not needed if the full list is removed. The plugin will simply add the services it needs during its initialization, just like it does the first time the accessory is created.

`api.updatePlatformAccessories(...);` updates the cache, just like `api.registerPlatformAccessories(...);` creates them.

n0rt0nthec4t:
@justjam2013 I agree.  ```api.updatePlatformAccessories(...)``` should be the method for this. The plugin should update the accessory to reflect its current state, including removing stale services, and then have Homebridge persist that state correctly. Adding another reconciliation API feels unnecessary if the existing update lifecycle is made to work as expected.

bwp91:
Again, just ideas:

---

@justjam2013 — one correction first, because it matters for what follows. The plugin does not need to keep any history to know what is stale. The restored accessory *is* the history: `accessory.services` at runtime is what was published last time, so "which of these do I no longer want" is a simple filter against what the plugin wants now. And I would steer you away from the remove-everything-and-re-add version specifically: the instance IDs that HomeKit scenes and automations point at are only kept stable while a service stays attached. There is a cleanup pass that expires the IDs of anything not attached at that moment, so remove-and-re-add works when it all happens synchronously and silently breaks automations when any async work lands in between — and it works on a fresh install but not on one with a restored cache, which is the worst way for something to fail. Targeted removal of the services you no longer want has none of these problems.

@n0rt0nthec4t — your framing is the one I want to build on: the plugin removes its stale services, then updates, and Homebridge makes that lifecycle work properly. That is in fact today's documented pattern. What this PR was really adding was never the removal — `removeService` already exists — it was the guard rails: don't let anyone remove `AccessoryInformation`, and know when a service belongs to a controller. And you are both right that those guard rails do not need a new method to live in.

**So here is the reshape I would like to make:**

1. **`removeService` refuses the HAP-managed services** (`AccessoryInformation`, `ProtocolInformation`) with a clear error. Removing those is never valid and currently breaks the accessory in confusing ways.
2. **`removeService` warns — but does not refuse — when the service is controller-owned**, using the controller detection written for this PR. A warning rather than a refusal is deliberate: today plugins sometimes *have* to reach across that boundary (the stale `CameraOperatingMode` after switching HKSV off is the standing example), because controllers do not yet clean up after themselves. The warning tells the author they are working around a core gap, and tells us how often it happens.
3. **`updatePlatformAccessories` persistence gets fixed wherever it is found lacking** — one such fix (a restored accessory losing its plugin attribution on alternate restarts) already went into 2.4.1.
4. **The pattern gets documented properly** in the plugin developer docs: remove the specific services you no longer want, never the information services, then update.
5. **No new API.** `reconcileServices` is dropped.

The longer-term piece stays tracked on #3980: controllers cleaning up their own services when their configuration changes (at which point the warning in 2 can graduate to a refusal), and core validating the cached state it owns. The warnings from 2 are exactly the evidence that work needs.

@danieloleary — I want to be straight with you about what this means for the PR, because you have engaged with every round of review and the quality of the work is not in question. The controller-ownership detection, the error messages, and most of your test suite carry over directly into the `removeService` hardening — the substance of what you built ships, in the API every plugin already calls, protecting people who would never have found a new method. If you are willing to rework the PR along those lines I would be glad to review it; if you would rather I take it from here, I will make sure the commits credit where the detection came from.

Thanks all three — this landed somewhere better than where any of us started.

justjam2013:
@AI — you are really getting on my nerves.

> The plugin does not need to keep any history to know what is stale

It needs to reconcile the restored accessory against what the plugin wants now. Same difference. The plugin is the golden master, so if Homebridge wants to cache, then it is the one that reconciles against the plugin. The plugin should not be aware of Homebridge's internals.

> And I would steer you away from the remove-everything-and-re-add version specifically

Please do not steer me towards the current internal implementation as an excuse for why something won't work. The plugin should not be aware of Homebridge's internals.

> Targeted removal of the services you no longer want has none of these problems.

If you understood the problem don't be a d**k about it.

n0rt0nthec4t:
Agreed.. AI slop.. no offence

justjam2013:
> `removeService` refuses the HAP-managed services (AccessoryInformation, ProtocolInformation) with a clear error. Removing those is never valid and currently breaks the accessory in confusing ways.

This is a bunch of horse manure also. If the plugin developer wants to remove a service, then it's their choice to do it and break things if they want to. If a developer can call `addService` to add a service, then they should be able to call `removeService` to remove it. `removeService` shouldn't be a magical service that works all the time except for full moon nights because ... you know, werewolves. Once again, for the slow AI in the back seat, the plugin is the source, the golden master, not the Homebridge API.

**Edit:** It would be fine for `updatePlatformAccessories` to throw an exception if the provided accessories were not properly configured. However, this would match exceptions thrown by `registerPlatformAccessories`.

danieloleary:
Thanks @bwp91 for laying out the revised direction and for being clear that the quality of the work is not the issue. Given the broader ownership discussion, I think the cleanest path is for you to take the reshape from here.

Please reuse the controller detection, diagnostics, and tests where they help, with credit as appropriate. I’m happy to answer concrete questions, but I don’t want to add more noise to the design discussion.

danieloleary:
Hi @bwp91, checking back on the August 29 handoff. Are you still planning to take on the `removeService` changes, or should we leave this parked for now?

You’re welcome to reuse the controller detection, diagnostics, and tests. If refreshing the tests would help, let me know which direction you want to take and I can do that.

Thanks for working through this with me.

PR #3999, opened 2026-08-23
fix: guard child bridge shutdown before bridgeService is assigned

### LINKED ISSUE #3998: Child bridge shutdown throws when SIGTERM arrives before bridgeService is assigned
### Describe the bug

`ChildBridgeFork.shutdown()` calls `this.bridgeService.teardown()` unguarded. `bridgeService` is only assigned once the child bridge has finished initialising, so a `SIGTERM` that arrives in the window between the fork starting and that assignment throws inside the signal handler:

```
[8/23/2026, 6:27:50 PM] Registering platform 'homebridge-roborock-matter.RoborockVacuumPlatform'
[8/23/2026, 6:27:50 PM] [Roborock Vacuum] Child bridge started successfully (plugin v3.17.1).
[8/23/2026, 6:27:50 PM] Got SIGTERM, shutting down child bridge process...
[8/23/2026, 6:27:50 PM] Error during child bridge shutdown: TypeError: Cannot read properties of undefined (reading 'teardown')
    at ChildBridgeFork.shutdown (file:///var/lib/homebridge/node_modules/homebridge/src/childBridgeFork.ts:463:24)
    at signalHandler (file:///var/lib/homebridge/node_modules/homebridge/src/childBridgeFork.ts:552:21)
    at process.emit (node:events:509:28)
    at process.emit (…/source-map-support/source-map-support.js:516:21)
```

The current source has the same shape:

```ts
shutdown(): void {
  this.bridgeService.teardown()
  ...
}
```

The Matter teardown immediately below it is already guarded with `if (this.matterManager)`; `bridgeService` is not.

### To Reproduce

Restart Homebridge from the UI while one child bridge is still coming up. On this instance it happened during a plugin-update restart: eleven child bridges took `SIGTERM` at `6:27:49`, and the twelfth signal landed on a bridge that had logged "Child bridge started successfully" one second earlier and had not yet finished initialising.

It is a race, so it is timing-dependent rather than reliably reproducible — which is also why it only showed up once in a log covering thirteen restarts.

### Expected behaviour

A shutdown signal that arrives before the bridge service exists should end the process quietly. There is nothing to tear down yet, and the process is being killed regardless.

```ts
shutdown(): void {
  this.bridgeService?.teardown()
  ...
}
```

### Impact

Cosmetic as far as I can measure it — the process was exiting anyway, all eleven child bridges came back healthy on the next start, and nothing was lost. The cost is that it prints a `TypeError` with a stack trace into the user's log during a routine restart, which reads like a plugin fault and sends people looking in the wrong place. It reached me as a suspected bug in my own plugin.

### Environment

- Homebridge **2.4.1-beta.6** (beta channel), HAP v2.2.3
- Homebridge UI 5.28.0
- Node v24.19.0, Linux, `hb-service`
- 11 child bridges, two of them Matter (`@matter/main` 0.18.0-alpha.0-20260823-b851ce019)

Not observed on 2.4.1-beta.3 or beta.5 in the same log, but the earlier restarts on those versions did not log a child-bridge `SIGTERM` shutdown at all, so I would not read that as evidence of a regression — only that this is the first shutdown of this shape I have a record of.

SAY-5:

`ChildBridgeFork.shutdown()` calls `this.bridgeService.teardown()` unguarded, but `bridgeService` is only assigned once the child bridge has finished initialising. A SIGTERM in the window between the fork starting and that assignment throws `TypeError: Cannot read properties of undefined (reading 'teardown')` inside the signal handler (#3998).

## :bulb: Proposed solution

Use optional chaining, matching how `sendPairedStatusEvent` already accesses `bridgeService`. The Matter teardown directly below is already guarded the same way.

## :gear: Release Notes

Fixed a TypeError during child bridge shutdown when a termination signal arrived before the bridge finished starting.

Fixes #3998

### Testing

Added a spec that calls `shutdown()` on a fresh `ChildBridgeFork` and expects no throw; it fails without the fix. `npm run lint` and the existing childBridgeFork specs pass.

### Reviewer Nudging

`src/childBridgeFork.ts`, `shutdown()`.

### DIFF
--- a/src/childBridgeFork.spec.ts
+++ b/src/childBridgeFork.spec.ts
@@ -79,6 +79,13 @@ describe('childBridgeFork - Matter Accessory Guard', () => {
   })
 })
 
+describe('childBridgeFork - shutdown', () => {
+  it('does not throw when a signal arrives before bridgeService is assigned', () => {
+    const fork = new ChildBridgeFork()
+    expect(() => fork.shutdown()).not.toThrow()
+  })
+})
+
 describe('childBridgeFork - Matter Handlers', () => {
   let childBridgeFork: ChildBridgeFork
   let mockMatterManager: any
--- a/src/childBridgeFork.ts
+++ b/src/childBridgeFork.ts
@@ -460,7 +460,9 @@ export class ChildBridgeFork {
   }
 
   shutdown(): void {
-    this.bridgeService.teardown()
+    // bridgeService is only assigned once the bridge has started, and a signal
+    // can arrive before that.
+    this.bridgeService?.teardown()
 
     // Teardown Matter servers (main bridge and external accessories)
     if (this.matterManager) {


bwp91 (approved):
Thanks for taking the time to fix this. I just ran the CI and it passed.

PR #4000, opened 2026-08-25
perf: reduce plugin discovery filesystem calls

n0rt0nthec4t:
```markdown

Homebridge plugin discovery currently calls `statSync` for every entry found in each plugin search path, including entries inside npm scope directories.

This requires a separate filesystem operation for every file and directory in the scanned `node_modules` locations. The additional calls can increase startup work, particularly on systems with slow storage or large global module directories.

Plugin discovery must continue supporting symbolic links because plugins installed using `npm link` appear as symlinked directories.

## :bulb: Proposed solution

Use `readdirSync(path, { withFileTypes: true })` so the returned `Dirent` objects can identify ordinary files and directories without an additional `statSync` call.

A focused `isDirectoryEntry` helper handles directory detection:

- Ordinary directories are accepted directly from their `Dirent`.
- Ordinary files are rejected directly.
- Symbolic links and entries with an unknown filesystem type fall back to `statSync`.

The fallback preserves existing support for linked plugins and less common filesystems that do not report an entry type.

The change applies to both top-level plugin search paths and scoped npm package directories. It does not change plugin qualification, ordering, loading, or public APIs.

## :gear: Release Notes

Reduced filesystem operations during Homebridge plugin discovery, improving startup efficiency on installations with large plugin directories or slower storage.

There are no breaking changes or migration requirements.

This is an internal startup-path optimization. Linked plugins installed using `npm link` remain supported through the targeted `statSync` fallback.

### Testing

Added a plugin-discovery regression test covering:

- Direct Homebridge plugin directories
- Scoped Homebridge plugin directories
- Symlinked plugin directories
- Non-plugin directories
- Files whose names resemble Homebridge plugins

Existing verification results:

- Lint passes
- Build passes
- All 1,296 tests pass

The fallback error path for inaccessible or broken symbolic links is not explicitly covered by the new test, but retains the existing behavior of logging and ignoring unreadable entries.

### Reviewer Nudging

Start with `PluginManager.isDirectoryEntry`, then review the two `readdirSync` call sites in `loadInstalledPlugins`.

The main compatibility consideration is the `statSync` fallback for symbolic links and unknown entry types, which preserves `npm link` support while avoiding per-entry stats for ordinary files and directories.
```

### DIFF
--- a/src/pluginManager.spec.ts
+++ b/src/pluginManager.spec.ts
@@ -1,3 +1,7 @@
+import { mkdirSync, mkdtempSync, rmSync, symlinkSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { HomebridgeAPI } from './api.js'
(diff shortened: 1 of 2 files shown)


bwp91 (approved):
Nice improvement, thanks!

PR #4001, opened 2026-08-26
feat: expose RVC direct mode change capability

jbwhy:
## Summary

- add an explicit per-accessory RVC Clean Mode feature declaration
- compose the standard Matter `DirectModeChange` feature only when a plugin explicitly opts in
- persist and restore feature metadata for root accessories and composed parts
- rebuild cached endpoints immediately when an enabled feature changes
- cover opt-in, default-disabled, runtime validation, cache persistence, and restart transitions

## Why

Matter RVC Clean Mode defines `DirectModeChange` for vacuums that can change clean mode while the RVC is not idle. Homebridge currently has no plugin API for advertising that capability, so controllers may disable the mode picker during an active run even when the physical robot and its plugin support live changes.

Handlers alone do not prove that a device can change modes while running, so this remains an explicit capability rather than being inferred.

## Plugin API

A compatible plugin opts in on the accessory:

```ts
features: {
  rvcCleanMode: {
    directModeChange: true,
  },
},
```

Only the literal boolean `true` enables the feature. Accessories that omit the declaration or set it to `false` retain the existing behavior.

## Persistence and restart behavior

Feature metadata is cached and restored for both root accessories and composed parts so the endpoint is rebuilt with the same Matter feature set before its plugin re-registers after a restart.

Enabled feature declarations are also part of Homebridge's restored-endpoint shape comparison. If a plugin adds or removes a feature, Homebridge rebuilds the endpoint during that same startup instead of retaining a stale endpoint until a later restart. Omitted and explicitly disabled features are normalized as the same shape to avoid unnecessary endpoint churn.

## Validation

- branch is current with `homebridge:latest`
- positive coverage verifies that opting in composes `DirectModeChange`
- negative coverage verifies that the feature remains disabled by default and for non-boolean runtime values
- cache coverage verifies feature persistence for root accessories and composed parts
- restart coverage verifies that changed features rebuild endpoints and equivalent disabled declarations do not
- dependent plugin draft: mathiashornbek/homebridge-roborock-matter#18
- the dependent plugin's existing CI matrix is green

The full Homebridge workflow matrix is awaiting maintainer approval because this PR originates from a fork.

### DIFF
--- a/src/matter/accessoryCache.spec.ts
+++ b/src/matter/accessoryCache.spec.ts
@@ -314,6 +314,7 @@ describe('matterAccessoryCache', () => {
         UUID: 'test-uuid',
         displayName: 'Test Device',
         deviceType: { name: 'Aggregator', code: 14 } as any,
+        features: { rvcCleanMode: { directModeChange: true } },
         serialNumber: 'SN-001',
         manufacturer: 'Test',
         model: 'Test',
@@ -323,6 +324,7 @@ describe('matterAccessoryCache', () => {
             id: 'part-1',
             displayName: 'Part 1',
             deviceType: { name: 'OnOffLight', code: 256 } as any,
+            features: { rvcCleanMode: { directModeChange: true } },
             clusters: { onOff: { onOff: true } },
           },
         ],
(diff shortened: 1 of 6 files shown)


bwp91:
Thanks for this — really nicely done, especially the cache/restore handling and the shape comparison, which is exactly where this kind of change usually goes wrong.

One small thing I'd like before merging: a warning for the case where a plugin declares `directModeChange: true` but doesn't provide a `changeToMode` handler. That combination is contradictory — the accessory advertises that modes can be changed mid-run, but nothing implements the change — and today it would be invisible until a controller tries it and nothing happens. Something like this in the RvcCleanMode branch:

```ts
if (accessory.features?.rvcCleanMode?.directModeChange === true) {
  if (!accessory.handlers?.rvcCleanMode?.changeToMode) {
    log.warn(`[${accessory.displayName}] rvcCleanMode.directModeChange is declared but there is no changeToMode handler - controllers will offer mid-run mode changes that nothing implements`)
  }
  behaviorClass = (behaviorClass as any).with('DirectModeChange')
  log.info('RvcCleanMode DirectModeChange feature enabled')
}
```

A warning (rather than rejecting the declaration) matches how we handle similar mismatches elsewhere. If you could add a test or two around it — the warning fires without the handler, and stays quiet with one — even better.

Happy to merge once that's in!

bwp91 (approved):
Thanks for taking the time to make this PR, looks good!

